PR #2, opened 2025-02-11
Added nix repo

jakemas:
Hanno and I will be doing some work on CMBC and formal verification for ML-DSA. Adding the nix repository will help with our work.

Callouts: have I included all the necessary files here?


nix/.DS_Store
mkannwischer, line 1: delete this file.

mkannwischer (changes requested):
You are missing the `flake.lock` and `flake.nix` in the root - without those the rest won't work. 
Maybe we can include the CI testing the nix setup at the same time. 
Simply including the `.github/workflows/nix.yml` should be sufficient. 

Also, to pass the DCO, you'll need to `git commit --signoff` for every commit.

PR #4, opened 2025-02-11
Added nix

jakemas:
Adding nix repo so Hanno and I can start some work on CBMC this week.

Testing my workflow and also `--signed` the commit :)

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,82 @@
+{
+  "nodes": {
+    "flake-parts": {
+      "inputs": {
+        "nixpkgs-lib": [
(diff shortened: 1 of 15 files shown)


mkannwischer (approved):
Works on my machine.
I took the liberty to rename `mlkem-native` to `mldsa-native`. 

Note that, there is currently a clean-up PR for the nix setup in mlkem-native: https://github.com/pq-code-package/mlkem-native/pull/765. 
Once that is merged, the same changes should be ported here. I'm not sure how we can best keep the two in sync.

Issue #8, opened 2025-02-21
Setup CBMC proof infrastructure


hanno-becker:
- [x] Setup `proofs/cbmc` directory as in mlkem-native
- [x] Add at least one proof to demonstrate that things work
- [x] Add CBMC to nix and CI

hanno-becker:
Done by https://github.com/pq-code-package/mldsa-native/issues/8

PR #13, opened 2025-02-21
Add CBMC infrastructure and prove `poly_add`

mkannwischer:
See the individual commit messages.

Note that CBMC is built from source and then cached in the Github cache. 
This allows us to quickly update CBMC in case there are updates, but it takes a while to build the first time (~15 minutes on the ubuntu-latest runners). 
In case this is not needed at some point, we may want to switch to the pre-built packages available in nix.

For now the CI runs on the default github Ubuntu runners. They are not super powerful, so once we have a large number of proofs we may want to switch to beefier machines (as in mlkem-native). 

* Resolves https://github.com/pq-code-package/mldsa-native/issues/8

### DIFF
--- a/.github/actions/cbmc/action.yml
+++ b/.github/actions/cbmc/action.yml
@@ -1,0 +1,53 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: CBMC
+description: Run CBMC proofs for mldsa-native
+
+inputs:
+  nix-shell:
+    description: Run in the specified Nix environment if exists
+    default: "ci-cbmc"
+  nix-cache:
+    description: Determine whether to enable nix cache
+    default: 'true'
+  nix-verbose:
+    description: Determine wether to suppress nix log or not
+    default: 'false'
+  custom_shell:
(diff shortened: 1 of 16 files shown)


.github/actions/cbmc/action.yml
```diff
@@ -0,0 +1,53 @@
...
+            cat >> $GITHUB_STEP_SUMMARY << EOF
+              ## Setup
+              Architecture: $(uname -m)
+              - $(nix --version)
+              - $(cbmc --version)
+              - litani Version $(litani --version)
+              - Cadical Version $(cadical --version)
+              - $(bash --version | grep -m1 "")
+            EOF
+      - name: Run CBMC proofs (MLKEM_K=${{ inputs.mlkem_k }})
```
hanno-becker: Copy pasta

hanno-becker (approved):
Minor typo, otherwise looks good -- thanks Matthias!

.github/actions/cbmc/action.yml
mkannwischer: thanks. fixed.

Issue #39, opened 2025-02-22
ACVP: Consider adding preHash test cases


mkannwischer:
FIPS-204 has a preHashing mode that passes a hash of the message instead of the full message. 

ACVP has test cases for the preHash mode with the external hash being one of 
"SHA2-224", "SHA2-256", "SHA2-384", "SHA2-512", "SHA2-512/224", "SHA2-512/256", "SHA3-224", "SHA3-256", "SHA3-384", "SHA3-512", "SHAKE-128", "SHAKE-256". 

Currently these test cases are being skipped as the current code does not support pre-hashing.
Do we want to support pre-hashing?

It is also allowed to compute the entirety of mu <-  H(BytesToBits(𝑡𝑟)||𝑀′, 64) externally.

jakemas:
Yes! I think we should add pre-hash modes. I plan to implement external mu.

PR #47, opened 2025-03-04
Implementation of external-mu mode for ML-DSA signing and verification

jakemas:
#### Description:
This PR implements the external-mu variant of ML-DSA signing and verification as specified in FIPS 204. The changes allow direct input of the mu value rather than computing it from the message during signing and verification.

#### Key changes:
- Added `crypto_sign_signature_extmu()` and `crypto_sign_verify_extmu()` functions that accept pre-computed mu values
- Modified internal signing/verification functions to support an external mu mode flag
- Updated function documentation to reference relevant FIPS 204 algorithms
- Maintained backward compatibility with existing signing/verification APIs

#### Call outs:
- Uses a boolean flag `externalmu` to control mu handling in internal functions
- Skips CRH computation when mu is provided externally
- Copies external mu directly when in external mode

The external-mu mode enables more flexible usage patterns while maintaining the security properties of ML-DSA. This is particularly useful for applications that need to separate mu computation from the core signing/verification operations.

#### Out of Scope:
Testing:
- Unit tests for external-mu mode signing and verification
- Verification of compatibility with existing usage
- Test vectors matching FIPS 204 specification

For reviewers: 

- naming conventions don't have to be final here. I'll propose a switch over from the `crypto_sign_*` convention for all APIs that can include any names we don't like here.
- I'm open to suggestions on how we implement external mu. Currently implemented as a boolean flag to the internal functions, to reduce duplicated code. The `verify_extmu` function is essentially just a wrapper, as no pre-processing needs to be completed (as context strings and pre-message processing is performed when constructing `mu`).

### DIFF
--- a/mldsa/sign.c
+++ b/mldsa/sign.c
@@ -108,6 +108,7 @@ int crypto_sign_keypair(uint8_t *pk, uint8_t *sk)
  *              - size_t prelen:  length of prefix string
  *              - uint8_t *rnd:   pointer to random seed
  *              - uint8_t *sk:    pointer to bit-packed secret key
+ *              - int externalmu: indicates input message m is processed as mu
  *
  * Returns 0 (success)
  **************************************************/
@@ -115,7 +116,7 @@ int crypto_sign_signature_internal(uint8_t *sig, size_t *siglen,
                                    const uint8_t *m, size_t mlen,
                                    const uint8_t *pre, size_t prelen,
                                    const uint8_t rnd[RNDBYTES],
-                                   const uint8_t *sk)
+                                   const uint8_t *sk, int externalmu)
 {
   unsigned int n;
   uint8_t seedbuf[2 * SEEDBYTES + TRBYTES + 2 * CRHBYTES];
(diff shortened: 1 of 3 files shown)


hanno-becker (changes requested):
@jakemas Can you make sure that the essential points of your change are reflected in the commit message? I see that you wrote a detailed PR description -- you can probably just use part/all of the PR message. (Most of my PR descriptions are just a copy or slight extensions of the commit message.)

mldsa/sign.c
```diff
@@ -254,8 +264,40 @@ int crypto_sign_signature(uint8_t *sig, size_t *siglen, const uint8_t *m,
...
+ *CRYPTO_BYTES)
+ *              - size_t *siglen: pointer to output length of signature
+ *              - uint8_t *mu:    pointer to input mu to be signed
+ *              - size_t mulen:   length of mu
+ *              - uint8_t *sk:    pointer to bit-packed secret key
+ *
+ * Returns 0 (success) or -1 (context string too long)
+ **************************************************/
+int crypto_sign_signature_extmu(uint8_t *sig, size_t *siglen, const uint8_t *mu,
+                                size_t mulen, const uint8_t *sk)
```
mkannwischer: I don't understand why a mulen is needed? Isn't this always 64 bytes? 
The internal functiosn ignore it anyway. 
Can we remove `mulen` and replace `const uint8_t *mu` by `const uint8_t mu[CRHBYTES]`?

mkannwischer (changes requested):
I'm fine with adding `*_extmu` functions, but I have one question/suggestion.

mldsa/sign.c
jakemas: Agreed - not used. [f8410a3](https://github.com/pq-code-package/mldsa-native/pull/47/commits/f8410a33d86a80b76089b473a9b242cd29bb9988)

mkannwischer: Thanks!

mkannwischer (approved):
Thanks @jakemas. 
With that it should be possible to cover the ACVP external mu tests (#39) - but we can do that in a follow-up.

PR #38, opened 2025-02-22
Add partial ACVP tests

mkannwischer:
This PR adds partial ACVP tests.
For keyGen, all tests are covered.

For sigGen and sigVer, the tests are currently limited to
preHash=pure (i.e., not testing the prehashing modes)
signatureInterface=external (i.e., only testing the external interface)
externalMu=false (i.e., not testing the external mu mode).

The other test types will be added later (probably with some further API tweaks).

The ACVP tests are executed (in CI) as a part of the quickcheck.

PR #49, opened 2025-03-05
Add benchmarks

### LINKED ISSUE #48: Add benchmarking code

mkannwischer:
Resolves #48 

Depends on #38

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,55 @@ jobs:
       - name: make quickcheck
         run: |
           make quickcheck
+  quickcheck_bench:
+    strategy:
+      fail-fast: false
+      matrix:
+        external:
+         - ${{ github.repository_owner != 'pq-code-package' }}
+        target:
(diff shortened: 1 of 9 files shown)


hanno-becker (changes requested):
Thanks @mkannwischer!

Seeing that you add `tests`, can you add a CI job which exercises `tests all`, or create an issue for it?

mkannwischer:
> Thanks @mkannwischer!
> 
> Seeing that you add `tests`, can you add a CI job which exercises `tests all`, or create an issue for it?

I created #51 for that. There is quite a bit of aligning with mlkem-native CI to do. I'll do it in a follow-up PR.

hanno-becker (changes requested):
I may be doing something wrong, but `tests bench -c MAC -r` does show only `0`s for me. Is that expected?

The benchmarking time is also quite long -- could you consider reducing the number of iterations?

hanno-becker (approved):

PR #50, opened 2025-03-05
Adds benchmarking on Raspberry Pi 4 (Cortex-A72)

mkannwischer:
Depends on #49.

Once merged to main, this should create a website in the [gh-pages branch ](https://github.com/pq-code-package/mldsa-native/tree/gh-pages) similar to https://pq-code-package.github.io/mlkem-native/dev/bench/ that we can use to keep track of performance changes over time. 
Once we have a baseline the bechmarks will trigger comments in pull requests showing the performance change.

If that all works I can easily add Raspberry Pi5, Cortex-A55, and Banana Pi F3 (rv64) benchmarks. We can also add the AWS EC2-based benchmarks then, but that's a bit more work.

### DIFF
--- a/.github/actions/bench/action.yml
+++ b/.github/actions/bench/action.yml
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
-name: Bench mlkem-native
+name: Bench mldsa-native
 description: Run benchmarking script
 
 inputs:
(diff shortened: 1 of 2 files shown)


.github/actions/bench/action.yml
```diff
@@ -0,0 +1,103 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: Bench mlkem-native
```
hanno-becker:
```suggestion
name: Bench mldsa-native
```

hanno-becker (approved):
LGTM, thanks @mkannwischer

Issue #10, opened 2025-02-21
Rename Dilithium -> MLDSA everywhere


hanno-becker:
Where Dilithium refers to the cryptographic scheme, replace by `MLDSA`. Where `Dilithium_` is used as a prefix for something implementation specific, use `MLD_...`, analogous to `MLK_...` in mlkem-native.

rod-chapman:
Suggest also remove the use of "PQCrystals" throughout. e.g. in proofs/cbmc/Makefile_params.common

Why does that file set "MLS_NAMESPACE". Should it not be "MLD_NAMESPACE"?

mkannwischer:
I don't have a strong opinion regarding MLS vs MLD.
Module-lattice signature (MLS) seemed consistent with module-lattice KEM (MLK) to me. 
But we can switch to MLD if you find  more appropriate.

hanno-becker:
I don't mind either way.

rod-chapman:
Agree - as long as we're consistent. Thinking about it, I think I prefer "MLS" short for "Module Lattice Signature" over
"MLD" short for "Module Lattice Digital". The former is more descriptive I guess.

hanno-becker:
@jakemas Can you take care of this? Let's to this ASAP so we don't accrue more code that later needs changing.

mkannwischer:
Resolved by #69

PR #59, opened 2025-03-12
Rename Dilithium to MLDSA

jakemas:
https://github.com/pq-code-package/mldsa-native/issues/10


.github/actions/cbmc/action.yml
```diff
@@ -43,11 +43,11 @@ runs:
               - Cadical Version $(cadical --version)
               - $(bash --version | grep -m1 "")
             EOF
-      - name: Run CBMC proofs (DILITHIUM_MODE=${{ inputs.mldsa_mode }})
+      - name: Run CBMC proofs (MLD_MODE=${{ inputs.mldsa_mode }})
```
mkannwischer: mode is scheme specific, so it should be `MLDSA_MODE`.

mldsa/params.h
```diff
@@ -27,7 +27,7 @@
 #define OMEGA 80
 #define CTILDEBYTES 32
 
-#elif DILITHIUM_MODE == 3
+#elif MLD_MODE == 3
 #define K 6
```
mkannwischer, line 31: Can you please also prefix those parameters with `MLDSA_` - this is only going to lead to clashes later on.

test/gen_NISTKAT.c
```diff
@@ -13,12 +13,12 @@
...
-#define CRYPTO_ALGNAME "Dilithium3"
-#elif DILITHIUM_MODE == 5
-#define CRYPTO_ALGNAME "Dilithium5"
+#if MLD_MODE == 2
+#define CRYPTO_ALGNAME "ML-DSA-44"
+#elif MLD_MODE == 3
+#define CRYPTO_ALGNAME "ML-DSA-65"
+#elif MLD_MODE == 5
+#define CRYPTO_ALGNAME "ML-DSA-87"
 #endif
```
mkannwischer, line 22: Oh... this shouldn't work. I opened https://github.com/pq-code-package/mldsa-native/issues/61 for this.

mkannwischer (changes requested):
Thanks @jakemas! Some small suggestions. 

The question remains if it should be `MLD_` or `MLS_` - I think in https://github.com/pq-code-package/mldsa-native/issues/10 there was an overall preference for `MLS_`. With this PR we are mixing `MLD_` and `MLS_` - that's not an option.

jakemas:
Opened https://github.com/pq-code-package/mldsa-native/pull/69

.github/actions/cbmc/action.yml
jakemas: Addressed in https://github.com/pq-code-package/mldsa-native/pull/69/files#diff-22bad1c6e1944cd640c5485c0eeaec078c78898153c8272c56dfaef762a6797a

mldsa/params.h
jakemas: Addressed in https://github.com/pq-code-package/mldsa-native/pull/69/files#diff-9cef4f9d244bae47e51787bb07fd9437a0db940688d882fd8f3a1ce9eefc6077

test/gen_NISTKAT.c
jakemas: Addressed in https://github.com/pq-code-package/mldsa-native/pull/69/files#diff-b467b00ff5e52a98e2d2cb2d46247325542d8e0e46a66d6928aa8d04651fca18R16-R21. Left the names the same.

PR #65, opened 2025-03-13
CBMC: switch --native-compiler to $(CC) instead of gcc

mkannwischer:
Currently CBMC defaults to use gcc. This is a problem when on MacOS, where
we default to clang. On these systems our CBMC proofs only work if the user
happens to have gcc in their $PATH.
This commit changes the flags for goto-cc to use the compiler set in $CC which
is clang on MacOS and gcc on other platforms.


mkannwischer:
@jakemas, can you check if this resolves your problems, please?

jakemas (approved):
Tested locally and no gcc errors

PR #66, opened 2025-03-14
Update CBMC to 6.5.0

mkannwischer:
This pulls in the most recent nix configuration from mlkem-native. Most notably this includes the recently released CBMC 6.5.0.

It also updates HOL-Light which is not required yet for mldsa-native, but let's keep it in sync.


jakemas (approved):
looks good to me!

PR #69, opened 2025-03-20
Rename Dilithium to ML-DSA

jakemas:
Addresses: https://github.com/pq-code-package/mldsa-native/issues/10

I went `MLD` over `MLS` because:
- The naming conventions from NIST are ML-DSA, FN-DSA, SLH-DSA, so this is more natural.
- MLS in crypto makes me think of Messaging Layer Security


mkannwischer (approved):
Thanks @jakemas. LGTM.
Okay - let's stick to MLD_

jakemas:
Thank you!

PR #82, opened 2025-03-29
proof of pack_sig

### LINKED ISSUE #72: CBMC: Proof `pack_sig`

rod-chapman:
Fixes #72 
1. Simplifies pack_sig() through use of memcpy() and memset() where possible.
2. Introduces new function polyvecl_pack_z() in polyvec.[hc] and its proof.
3. Adds formal parameter "number_of_hints" to pack_sig() and pre-condition that can be used to ensure type-safety.
4. Updates call to pack_sig() in sign.c appropriately.
5. Proof of pack_sig() updated.

Please review explanatory comments in body of pack_sig()

### DIFF
--- a/mldsa/packing.c
+++ b/mldsa/packing.c
@@ -130,35 +130,82 @@ void unpack_sk(uint8_t rho[MLDSA_SEEDBYTES], uint8_t tr[MLDSA_TRBYTES],
  * Description: Bit-pack signature sig = (c, z, h).
  *
  * Arguments:   - uint8_t sig[]: output byte array
- *              - const uint8_t *c: pointer to challenge hash length
- *MLDSA_SEEDBYTES
+ *              - const uint8_t *c:  pointer to challenge hash length
+ *                                   MLDSA_SEEDBYTES
  *              - const polyvecl *z: pointer to vector z
  *              - const polyveck *h: pointer to hint vector h
+ *              - const unsigned int number_of_hints: total
+ *                                   hints in *h
+ *
+ * Note that the number_of_hints argument is not present
+ * in the reference implementation. It is added here to ease
+ * proof of type safety.
  **************************************************/
 void pack_sig(uint8_t sig[CRYPTO_BYTES], const uint8_t c[MLDSA_CTILDEBYTES],
-              const polyvecl *z, const polyveck *h)
+              const polyvecl *z, const polyveck *h,
+              const unsigned int number_of_hints)
 {
   unsigned int i, j, k;
 
-  for (i = 0; i < MLDSA_CTILDEBYTES; ++i)
-    sig[i] = c[i];
+  memcpy(sig, c, MLDSA_CTILDEBYTES);
   sig += MLDSA_CTILDEBYTES;
 
-  for (i = 0; i < MLDSA_L; ++i)
-    polyz_pack(sig + i * MLDSA_POLYZ_PACKEDBYTES, &z->vec[i]);
+  polyvecl_pack_z(sig, z);
   sig += MLDSA_L * MLDSA_POLYZ_PACKEDBYTES;
 
-  /* Encode h */
-  for (i = 0; i < MLDSA_OMEGA + MLDSA_K; ++i)
-    sig[i] = 0;
+  /* Encode hints h */
+
+  /* The final section of sig[] is MLDSA_POLYVECH_PACKEDBYTES long, where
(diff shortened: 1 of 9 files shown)


rod-chapman:
CBMC proofs are running fine locally, but hanging up in CI. Odd.

rod-chapman:
OK... CI looks good now.

mldsa/packing.c
```diff
@@ -130,35 +130,74 @@ void unpack_sk(uint8_t rho[MLDSA_SEEDBYTES], uint8_t tr[MLDSA_TRBYTES],
...
+   *
+   * The first OMEGA bytes record the index numbers of the coefficients
+   * that are not equal to 0
+   *
+   * The final K bytes record a running tally of the number of hints
+   * coming from each of the K polynomials in h.
+   *
+   * The pre-condition tells us that number_of_hints <= OMEGA, so not
+   * all bytes will be written, so we initialize all of them to zero
+   * to start.
```
mkannwischer: "not all bytes will be written" -> "some bytes may not be written"? 
(it's <=, so we may actually write all bytes).

mldsa/packing.c
```diff
@@ -130,35 +130,74 @@ void unpack_sk(uint8_t rho[MLDSA_SEEDBYTES], uint8_t tr[MLDSA_TRBYTES],
...
-      if (h->vec[i].coeffs[j] != 0)
+    __loop__(
+      assigns(j, k, memory_slice(sig, MLDSA_POLYVECH_PACKEDBYTES))
+      invariant(i <= MLDSA_K)
+      invariant(j <= MLDSA_N)
+      invariant(k <= MLDSA_OMEGA)
+      invariant(number_of_hints <= MLDSA_OMEGA)
+    )
+    {
+      if (h->vec[i].coeffs[j] != 0 && k < number_of_hints)
```
mkannwischer, line 198: Please document the difference to the reference implementation here clearly.
It is also worth pointing out that the first condition implies the second for valid inputs.

mldsa/packing.c
```diff
@@ -130,35 +130,74 @@ void unpack_sk(uint8_t rho[MLDSA_SEEDBYTES], uint8_t tr[MLDSA_TRBYTES],
...
  *
  * Arguments:   - uint8_t sig[]: output byte array
- *              - const uint8_t *c: pointer to challenge hash length
- *MLDSA_SEEDBYTES
+ *              - const uint8_t *c:  pointer to challenge hash length
+ *                                   MLDSA_SEEDBYTES
  *              - const polyvecl *z: pointer to vector z
  *              - const polyveck *h: pointer to hint vector h
+ *              - const unsigned int number_of_hints: total
+ *                                   hints in *h
```
mkannwischer, line 138: Point out that `number_of_hints` is not present in the reference implementation and only required to proof type-safety of the function.

mldsa/packing.c
```diff
@@ -130,35 +130,74 @@ void unpack_sk(uint8_t rho[MLDSA_SEEDBYTES], uint8_t tr[MLDSA_TRBYTES],
...
+   */
+  memset(sig, 0, MLDSA_POLYVECH_PACKEDBYTES);
 
   k = 0;
+  /* For each polynomial in h... */
   for (i = 0; i < MLDSA_K; ++i)
+  __loop__(
+    assigns(i, j, k, memory_slice(sig, MLDSA_POLYVECH_PACKEDBYTES))
+    invariant(i <= MLDSA_K)
+    invariant(k <= MLDSA_OMEGA)
```
mkannwischer: If I change the two k invariants to `invariant(k <= number_of_hints)` and switch to `--smt2` instead of `--bitwuzla`, I get _much_ faster proof run-times (20 seconds instead of 3 minutes for `tests cbmc --single-step -p pack_sig`). 
Can you confirm?

mldsa/packing.c
rod-chapman: Done

mldsa/packing.c
rod-chapman: Done

mldsa/packing.c
rod-chapman: Confirmed. Good spot!

mldsa/packing.c
rod-chapman: Done. That comment should be in packing.h, but we will move them all later in a separate PR.

mkannwischer (approved):
Thanks for the changes @rod-chapman! LGTM. 
Can you squash the two commits, please?

Issue #22, opened 2025-02-21
CBMC: Prove `poly_sub`


mkannwischer:

hanno-becker:
@manastasova plans to do this one.

mkannwischer:
Wait - wasn't @praveksharma doing that?

hanno-becker:
My apologies, we'll find a different one for @manastasova then.

hanno-becker:
> Wait - wasn't @praveksharma doing that?

@praveksharma were you?

praveksharma:
> > Wait - wasn't [@praveksharma](https://github.com/praveksharma) doing that?
> 
> [@praveksharma](https://github.com/praveksharma) were you?

Hi, sorry for not getting back sooner. I've started work on https://github.com/pq-code-package/mldsa-native/issues/64 but am happy to look at this as well.

hanno-becker:
@praveksharma If you are happy doing #64 as a learning exercise, and then pick up another more complicated function as a follow-up, we could leave this one as the starter for @manastasova. Does that work?

praveksharma:
> [@praveksharma](https://github.com/praveksharma) If you are happy doing [#64](https://github.com/pq-code-package/mldsa-native/issues/64) as a learning exercise, and then pick up another more complicated function as a follow-up, we could leave this one as the starter for [@manastasova](https://github.com/manastasova). Does that work?

Certainly!

mkannwischer:
Resolved by #84

PR #84, opened 2025-03-31
Add spec and proof for poly_sub

manastasova:
This PR adds the spec and proof for `poly_sub` function. 
This PR does _not_ remove `poly_sub` benchmarking `DBENCH_START();` and `DBENCH_STOP(*tsub);` calls.

Solves https://github.com/pq-code-package/mldsa-native/issues/22.

### DIFF
--- a/mldsa/poly.c
+++ b/mldsa/poly.c
@@ -13,7 +13,7 @@
 #ifdef DBENCH
 #include "test/cpucycles.h"
 extern const uint64_t timing_overhead;
-extern uint64_t *tred, *tadd, *tmul, *tround, *tsample, *tpack;
+extern uint64_t *tred, *tadd, *tsub, *tmul, *tround, *tsample, *tpack;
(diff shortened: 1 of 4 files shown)


mkannwischer (approved):
Thanks @manastasova! Looks good to me.
We are not using the profiling with `DBENCH_START()`/`DBENCH_STOP()`, so these can be removed. But we can remove it globally in a separate PR>

mkannwischer (changes requested):
Can you please run `format` in the `nix develop` shell to eliminate the linting warnings?

mkannwischer (changes requested):
Thanks! Looks good now. 

Can you please clean up the commit history before we merge. Simply squashing all your commits should be sufficient.
Please don't do any merge commits from `main` to your branch, but rebase your commits on top of main to keep the history clean.

mkannwischer (approved):

Issue #26, opened 2025-02-21
CBMC: Prove `rej_eta`


mkannwischer:

manastasova:
Addressed in https://github.com/pq-code-package/mldsa-native/pull/86.

PR #86, opened 2025-04-01
Add spec and proof for rej_eta

manastasova:
This PR adds the spec and proof for rej_eta function.

Notes::
- `rej_eta` is static, thus, `CHECK_FUNCTION_CONTRACTS` is defined as `rej_eta` instead of $`(MLD_NAMESPACE)rej_eta` in the `Makefile`.
- `rej_eta` is static, thus, additional assumptions, such as `requires(len <= MLDSA_N && buflen <= (POLY_UNIFORM_ETA_NBLOCKS * STREAM256_BLOCKBYTES))` are made. This accelerates the performance of CBMC proofs.

Another Note::
- [change types](https://github.com/manastasova/mldsa-native/blob/783550404c42fa5659f4116d8b7c177c32a4d2b6/mldsa/poly.c#L442) `uint32_t t0, t1;` to `int32_t t0, t1;` due to [potential overflow in](https://github.com/manastasova/mldsa-native/blob/783550404c42fa5659f4116d8b7c177c32a4d2b6/mldsa/poly.c#L467) `if (t0 < 9){a[ctr++] = 4 - t0;}` causing cbmc fail.

Solves https://github.com/pq-code-package/mldsa-native/issues/26.

### DIFF
--- a/mldsa/poly.c
+++ b/mldsa/poly.c
@@ -435,15 +435,37 @@ void poly_uniform(poly *a, const uint8_t seed[MLDSA_SEEDBYTES], uint16_t nonce)
  * Returns number of sampled coefficients. Can be smaller than len if not enough
  * random bytes were given.
  **************************************************/
+#if MLDSA_ETA == 2
+#define POLY_UNIFORM_ETA_NBLOCKS \
+  ((136 + STREAM256_BLOCKBYTES - 1) / STREAM256_BLOCKBYTES)
+#elif MLDSA_ETA == 4
+#define POLY_UNIFORM_ETA_NBLOCKS \
+  ((227 + STREAM256_BLOCKBYTES - 1) / STREAM256_BLOCKBYTES)
+#else
+#error "Invalid value of MLDSA_ETA"
+#endif
 static unsigned int rej_eta(int32_t *a, unsigned int len, const uint8_t *buf,
                             unsigned int buflen)
+__contract__(
+  requires(len <= buflen && len <= MLDSA_N && \
+                  buflen <= (POLY_UNIFORM_ETA_NBLOCKS * STREAM256_BLOCKBYTES))
+  requires(memory_no_alias(a, sizeof(int32_t) * len))
+  requires(memory_no_alias(buf, buflen))
+  assigns(memory_slice(a, sizeof(int32_t) * len))
+  ensures(return_value <= len)
+  ensures(array_abs_bound(a, 0, return_value, MLDSA_ETA + 1))
+)
 {
   unsigned int ctr, pos;
-  uint32_t t0, t1;
+  int32_t t0, t1;
   DBENCH_START();
 
   ctr = pos = 0;
   while (ctr < len && pos < buflen)
+  __loop__(
+    invariant(0 <= ctr && ctr <= len && pos <= buflen)
+    invariant(ctr > 0 ==> array_abs_bound(a, 0, ctr, MLDSA_ETA + 1))
+  )
   {
     t0 = buf[pos] & 0x0F;
     t1 = buf[pos++] >> 4;
@@ -489,15 +511,6 @@ static unsigned int rej_eta(int32_t *a, unsigned int len, const uint8_t *buf,
  *MLDSA_CRHBYTES
  *              - uint16_t nonce: 2-byte nonce
  **************************************************/
-#if MLDSA_ETA == 2
-#define POLY_UNIFORM_ETA_NBLOCKS \
-  ((136 + STREAM256_BLOCKBYTES - 1) / STREAM256_BLOCKBYTES)
-#elif MLDSA_ETA == 4
-#define POLY_UNIFORM_ETA_NBLOCKS \
-  ((227 + STREAM256_BLOCKBYTES - 1) / STREAM256_BLOCKBYTES)
-#else
-#error "Invalid value of MLDSA_ETA"
-#endif
 void poly_uniform_eta(poly *a, const uint8_t seed[MLDSA_CRHBYTES],
                       uint16_t nonce)
 {
--- a/mldsa/poly.c
+++ b/mldsa/poly.c
@@ -457,14 +457,14 @@ __contract__(
 )
 {
   unsigned int ctr, pos;
-  int32_t t0, t1;
+  uint32_t t0, t1;
   DBENCH_START();
 
   ctr = pos = 0;
   while (ctr < len && pos < buflen)
   __loop__(
     invariant(0 <= ctr && ctr <= len && pos <= buflen)
-    invariant(ctr > 0 ==> array_abs_bound(a, 0, ctr, MLDSA_ETA + 1))
+    invariant(array_abs_bound(a, 0, ctr, MLDSA_ETA + 1))
   )
   {
     t0 = buf[pos] & 0x0F;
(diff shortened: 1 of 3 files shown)


mldsa/poly.c
```diff
@@ -417,15 +417,37 @@ void poly_uniform(poly *a, const uint8_t seed[MLDSA_SEEDBYTES], uint16_t nonce)
...
   unsigned int ctr, pos;
-  uint32_t t0, t1;
+  int32_t t0, t1;
   DBENCH_START();
 
   ctr = pos = 0;
   while (ctr < len && pos < buflen)
+  __loop__(
+    invariant(0 <= ctr && ctr <= len && pos <= buflen)
+    invariant(ctr > 0 ==> array_abs_bound(a, 0, ctr, MLDSA_ETA + 1))
```
mkannwischer: I think you do not need the implication here, do you? 
array_abs_bound should be true anyway if ctr=0. 
```suggestion
    invariant(array_abs_bound(a, 0, ctr, MLDSA_ETA + 1))
```

mldsa/poly.c
```diff
@@ -417,15 +417,37 @@ void poly_uniform(poly *a, const uint8_t seed[MLDSA_SEEDBYTES], uint16_t nonce)
...
+  requires(memory_no_alias(a, sizeof(int32_t) * len))
+  requires(memory_no_alias(buf, buflen))
+  assigns(memory_slice(a, sizeof(int32_t) * len))
+  ensures(return_value <= len)
+  ensures(array_abs_bound(a, 0, return_value, MLDSA_ETA + 1))
+)
 {
   unsigned int ctr, pos;
-  uint32_t t0, t1;
+  int32_t t0, t1;
```
mkannwischer: I don't think this is a good idea. `t0` and `t1` are conceptually unsigned values. 
Can you instead add an explicit cast to `int32_t` to the assignments to `a[...]`? That would be a cleaner solution.

hanno-becker (dismissed):
I believe that the computation of `t0,t1` itself _is_ unsigned; it's only when you convert to signed prior to `2 - t_i` and `4 - t_i` that you need an explicit cast, as otherwise that subtraction is computed in `unsigned`. Marking `t0,t1` as signed would not check for underflows in `t0 = t0 - (205 * t0 >> 10) * 5;` and `t1 = t1 - (205 * t1 >> 10) * 5;`, which we want CBMC to do.

mkannwischer (changes requested):
Thanks! 
I left two comments in the code. Also, could you please add the description to the commit message as well? The PR descriptions are hard to track down after the PR has been merged. You may just refer to the commit msg in the PR description.

mkannwischer:
@manastasova, should we add you as contributor to mldsa-native?
Then you have a few more rights in managing the issues, I think. It will also allow you to work in local branches which will be needed as soon as we are running CI on AWS EC2 machines (as that will not work from a fork).

mldsa/poly.c
manastasova: My understanding is that the validation through assertions should start from the second iteration onward, wherein we verify the correctness of both `a[0]` and `a[1]`, corresponding to subarrays spanning from `index_low = 0` to  `index_high = 2`, etc. However, as discussed with @rod-chapman ,`ctr > 0` would not change the validation when passing `index_low = 0` to  `index_high = 0`. 
I can confirm that removing the implication does not affect the successful outcome of the CBMC proof.
Should I update [mlkem-native](https://github.com/pq-code-package/mlkem-native/?blob/26ea08ca61ecc26d01b087afdbda186a5e7d53e0/mlkem/sampling.c#L41) similar implication?

mldsa/poly.c
manastasova: I considered casting the result, however, I am still seeing cbmc fails.
```
[rej_eta.overflow.18] line 459 arithmetic overflow on unsigned - in (unsigned int)2 - t0: FAILURE
[rej_eta.overflow.19] line 459 arithmetic overflow on unsigned to signed type conversion in (int32_t)((unsigned int)2 - t0): FAILURE
[rej_eta.overflow.26] line 464 arithmetic overflow on unsigned - in (unsigned int)2 - t1: FAILURE
[rej_eta.overflow.27] line 464 arithmetic overflow on unsigned to signed type conversion in (int32_t)((unsigned int)2 - t1): FAILURE
[rej_eta.overflow.14] line 468 arithmetic overflow on unsigned - in (unsigned int)(int32_t)4 - t0: FAILURE
[rej_eta.overflow.15] line 468 arithmetic overflow on unsigned to signed type conversion in (int32_t)((unsigned int)(int32_t)4 - t0): FAILURE
[rej_eta.overflow.29] line 468 arithmetic overflow on unsigned - in (unsigned int)(int32_t)4 - t0: FAILURE
[rej_eta.overflow.30] line 468 arithmetic overflow on unsigned to signed type conversion in (int32_t)((unsigned int)(int32_t)4 - t0): FAILURE
[rej_eta.overflow.19] line 470 arithmetic overflow on unsigned - in (unsigned int)(int32_t)4 - t1: FAILURE
[rej_eta.overflow.20] line 470 arithmetic overflow on unsigned to signed type conversion in (int32_t)((unsigned int)(int32_t)4 - t1): FAILURE
[rej_eta.overflow.34] line 470 arithmetic overflow on unsigned - in (unsigned int)(int32_t)4 - t1: FAILURE
[rej_eta.overflow.35] line 470 arithmetic overflow on unsigned to signed type conversion in (int32_t)((unsigned int)(int32_t)4 - t1): FAILURE
```
Casting the auxiliary values `t0, t1` before use solves the fails, however, I thought it is simpler and cleaner to change the type.

mldsa/poly.c
rod-chapman: Proofs looks good here without the `ctr > 0` guard, so I'd probably remove it.

rod-chapman:
Yes... Please add Mila as a contributor here, and please add her to the Discord channel. Thanks.

manastasova:
> @manastasova, should we add you as contributor to mldsa-native? Then you have a few more rights in managing the issues, I think. It will also allow you to work in local branches which will be needed as soon as we are running CI on AWS EC2 machines (as that will not work from a fork).

That would be great! Thanks @mkannwischer!

mldsa/poly.c
hanno-becker: You need to cast `t0` and `t1` to `int32_t` before doing the subtraction, so `a[..] = 2 - (int32_t) t`. Otherwise, the subtraction is computed in `uint32_t`, and does indeed unverflow.

hanno-becker: If you define `t0,t1` as `int32_t`, CBMC would not catch an underflow in `t0 = t0 - (205 * t0 >> 10) * 5;`, which makes the proof weaker.

mkannwischer:
> > @manastasova, should we add you as contributor to mldsa-native? Then you have a few more rights in managing the issues, I think. It will also allow you to work in local branches which will be needed as soon as we are running CI on AWS EC2 machines (as that will not work from a fork).
> 
> That would be great! Thanks @mkannwischer!

I open this PR to add you as a contributor: https://github.com/pq-code-package/tsc/pull/152
We'll get that merged in the next few days.

The PQCA Discord is open for everyone to join: https://pqca.org/get-involved/
Find us in the mldsa-native-developrs channel.

mldsa/poly.c
mkannwischer: You are right  - this can also be removed in mlkem-native.
I have opened a PR that does it: https://github.com/pq-code-package/mlkem-native/pull/943

mldsa/poly.c
manastasova: I see your point. 
My understanding here is that the` if (t0 < 15) `condition should safeguard against underflow scenarios in this context. With this constraint in place, whether the computation uses signed or unsigned values, the actual values would have the same binary representation and would not cause under/overflow in the given context.
If underflow isn't possible in the `t0 = t0 - (205 * t0 >> 10) * 5;` calculation and we already need an explicit cast to `int32_t` in `a[...] = 2 - (int32_t)t0`, then I thought it would be simpler to declare `t0` as `int32_t` from the beginning rather than using an unsigned type that requires casting later.

mkannwischer (approved):
Thanks @manastasova. LGTM!
I took the liberty to rebase this on top of main.

PR #89, opened 2025-04-03
CBMC: Add contract and proof for power2round 

mkannwischer:
Resolves https://github.com/pq-code-package/mldsa-native/issues/18

### DIFF
--- a/mldsa/rounding.h
+++ b/mldsa/rounding.h
@@ -10,7 +10,20 @@
 #include "params.h"
 
 #define power2round MLD_NAMESPACE(power2round)
-int32_t power2round(int32_t *a0, int32_t a);
+/*************************************************
+ * Name:        power2round
+ *
+ * Description: For finite field element a, compute a0, a1 such that
+ *              a mod^+ MLDSA_Q = a1*2^MLDSA_D + a0 with -2^{MLDSA_D-1} < a0 <=
+ *              2^{MLDSA_D-1}. Assumes a to be standard representative.
+ *
+ * Arguments:   - int32_t a: input element
+ *              - int32_t *a0: pointer to output element a0
+ *              - int32_t *a1: pointer to output element a1
+ *
+ * Reference: a1 is passed as a return value instead
+ **************************************************/
+void power2round(int32_t *a0, int32_t *a1, int32_t a);
 
 #define decompose MLD_NAMESPACE(decompose)
 int32_t decompose(int32_t *a0, int32_t a);
--- a/mldsa/rounding.h
+++ b/mldsa/rounding.h
@@ -9,6 +9,8 @@
 #include "cbmc.h"
 #include "params.h"
 
+#define MLD_2_POW_D (1 << MLDSA_D)
+
 #define power2round MLD_NAMESPACE(power2round)
 /*************************************************
  * Name:        power2round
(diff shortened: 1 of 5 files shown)


mkannwischer:
I tried `poly_power2round` as well, but I cannot get that to work as of now.

mldsa/rounding.h
```diff
@@ -10,7 +10,31 @@
...
+ *              a mod^+ MLDSA_Q = a1*2^MLDSA_D + a0 with -2^{MLDSA_D-1} < a0 <=
+ *              2^{MLDSA_D-1}. Assumes a to be standard representative.
+ *
+ * Arguments:   - int32_t a: input element
+ *              - int32_t *a0: pointer to output element a0
+ *              - int32_t *a1: pointer to output element a1
+ *
+ * Reference: a1 is passed as a return value instead
+ **************************************************/
+void power2round(int32_t *a0, int32_t *a1, int32_t a)
```
rod-chapman, line 28: Parameter "a" should be marked "const" both here and in the .c file.

rod-chapman (commented):
Proof looks good. One minor comment.

mldsa/rounding.h
mkannwischer: Isn't `const` meaningless for parameters passed by value? 

https://softwareengineering.stackexchange.com/questions/204500/when-and-for-what-purposes-should-the-const-keyword-be-used-in-c-for-variables
https://google.github.io/styleguide/cppguide.html#Use_of_const

suggest you should not use it for the function declarations as it provides no value to the caller. You can use it in the function definition, but here I am not sure if it adds any value as the function has just two lines that obviously do not modify `a`. 

I'd prefer to keep it as is.

hanno-becker: Agree. `const` for by-value parameters carries no information in declarations.

mldsa/rounding.h
```diff
@@ -10,7 +10,31 @@
...
+__contract__(
+  requires(memory_no_alias(a0, sizeof(int32_t)))
+  requires(memory_no_alias(a1, sizeof(int32_t)))
+  requires(a >= 0 && a < MLDSA_Q)
+  assigns(memory_slice(a0, sizeof(int32_t)))
+  assigns(memory_slice(a1, sizeof(int32_t)))
+  ensures(*a0 > -(1<<MLDSA_D) && *a0 <= (1<<MLDSA_D))
```
hanno-becker: General comment: Does it make sense to introduction a macro-abbreviation for `1<<MLDSA_D`?

hanno-becker (approved):

mldsa/rounding.h
rod-chapman: I stand corrected! I never realized it can be "const" in the definition, but not in the declaration! I'd still prefer it to be const in the defintion, since it prevents the implementation modifying it, and provides an optimization hint I presume.

mldsa/rounding.h
rod-chapman: It appears on 3 lines in rounding.h, so I'd say Yes...

mkannwischer: okay - I added `MLD_2_POW_D`

mldsa/rounding.h
mkannwischer: Okay - I added it to rounding.c

mkannwischer:
Plase take another look - my bounds on a0 were actually wrong.

rod-chapman:
Yes... I see it's "2^(MLDSA_D-1)" in the Description comment. Looks good and proof OK.

rod-chapman (approved):

PR #90, opened 2025-04-03
Add proof for poly_reduce

### LINKED ISSUE #64: CBMC: Prove `poly_reduce`
With `reduce32` (#15) being proved, we can prove `poly_reduce` using the `reduce32` contract.

praveksharma:
1. Adds spec for `poly_reduce` using proof for reduce32. Fixes #64.
2. Edits `reduce32` proof to use `cassert` instead of `ensures`.
3. Edits `cbmc.h` to ignore `cassert` statements unless doing a CBMC proof.
4. Moves `REDUCE_*` defines from `reduce.h`, `poly.h` to common header file `params.h` to avoid divergent values. The defines are renamed as per convention.

### DIFF
--- a/mldsa/poly.c
+++ b/mldsa/poly.c
@@ -35,6 +35,10 @@ void poly_reduce(poly *a)
   DBENCH_START();
 
   for (i = 0; i < MLDSA_N; ++i)
+  __loop__(
+    invariant(i <= MLDSA_N)
+    invariant(forall(k0, i, MLDSA_N, a->coeffs[k0] == loop_entry(*a).coeffs[k0]))
+    invariant(array_bound(a->coeffs, 0, i, -REDUCE_RANGE_MAX, REDUCE_RANGE_MAX)))
   {
     a->coeffs[i] = reduce32(a->coeffs[i]);
   }
--- a/mldsa/poly.h
+++ b/mldsa/poly.h
@@ -8,6 +8,7 @@
 #include <stdint.h>
 #include "cbmc.h"
 #include "params.h"
+#include "reduce.h"
 
 typedef struct
 {
@@ -15,7 +16,14 @@ typedef struct
 } poly;
 
 #define poly_reduce MLD_NAMESPACE(poly_reduce)
-void poly_reduce(poly *a);
+void poly_reduce(poly *a)
+__contract__(
+  requires(memory_no_alias(a, sizeof(poly)))
+  requires(forall(k0, 0, MLDSA_N, a->coeffs[k0] <= REDUCE_DOMAIN_MAX)) 
+  assigns(memory_slice(a, sizeof(poly)))
+  ensures(array_bound(a->coeffs, 0, MLDSA_N, -REDUCE_RANGE_MAX, REDUCE_RANGE_MAX))
+);
+
 #define poly_caddq MLD_NAMESPACE(poly_caddq)
 void poly_caddq(poly *a);
 
--- a/mldsa/reduce.c
+++ b/mldsa/reduce.c
@@ -76,6 +76,7 @@ int32_t reduce32(int32_t a)
 
   t = (a + (1 << 22)) >> 23;
   t = a - t * MLDSA_Q;
+  cassert((t - a) % MLDSA_Q == 0);
   return t;
 }
 
--- a/mldsa/reduce.h
+++ b/mldsa/reduce.h
@@ -38,7 +38,6 @@ __contract__(
   requires(a <= REDUCE_DOMAIN_MAX)
   ensures(return_value >= -REDUCE_RANGE_MAX)
   ensures(return_value <   REDUCE_RANGE_MAX)
-  ensures((return_value - a) % MLDSA_Q == 0)
 );
 
 #define caddq MLD_NAMESPACE(caddq)
--- a/proofs/cbmc/poly_reduce/Makefile
+++ b/proofs/cbmc/poly_reduce/Makefile
@@ -1,0 +1,54 @@
+# SPDX-License-Identifier: Apache-2.0
+
+include ../Makefile_params.common
+
+HARNESS_ENTRY = harness
+HARNESS_FILE = poly_reduce_harness
+
+# This should be a unique identifier for this proof, and will appear on the
+# Litani dashboard. It can be human-readable and contain spaces if you wish.
+PROOF_UID = poly_reduce
+
+DEFINES +=
+INCLUDES +=
+
+REMOVE_FUNCTION_BODY +=
+UNWINDSET +=
+
+PROOF_SOURCES += $(PROOFDIR)/$(HARNESS_FILE).c
+PROJECT_SOURCES += $(SRCDIR)/mldsa/poly.c
+
+CHECK_FUNCTION_CONTRACTS=$(MLD_NAMESPACE)poly_reduce
+USE_FUNCTION_CONTRACTS=$(MLD_NAMESPACE)reduce32
+APPLY_LOOP_CONTRACTS=on
+USE_DYNAMIC_FRAMES=1
+
+# Disable any setting of EXTERNAL_SAT_SOLVER, and choose SMT backend instead
+EXTERNAL_SAT_SOLVER=
+CBMCFLAGS=--smt2
+
+FUNCTION_NAME = poly_reduce
+
+# If this proof is found to consume huge amounts of RAM, you can set the
+# EXPENSIVE variable. With new enough versions of the proof tools, this will
+# restrict the number of EXPENSIVE CBMC jobs running at once. See the
+# documentation in Makefile.common under the "Job Pools" heading for details.
+# EXPENSIVE = true
+
+# This function is large enough to need...
+CBMC_OBJECT_BITS = 8
+
+# If you require access to a file-local ("static") function or object to conduct
+# your proof, set the following (and do not include the original source file
+# ("mldsa/poly.c") in PROJECT_SOURCES).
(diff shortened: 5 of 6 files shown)


mldsa/poly.c
```diff
@@ -4,6 +4,7 @@
  */
 #include "poly.h"
 #include <stdint.h>
+#include "cbmc.h"
```
manastasova: This header file is unnecessary here since it's already included by the `poly.h` header file.

mldsa/poly.c
```diff
@@ -35,6 +36,10 @@ void poly_reduce(poly *a)
   DBENCH_START();
 
   for (i = 0; i < MLDSA_N; ++i)
+  __loop__(
+    assigns(i, object_whole(a->coeffs))
```
manastasova: I'm not sure it's necessary to explicitly specify the loop's footprint since it already overlaps with the function's overall footprint. The memory regions accessed by the loop appear to be a subset of what the function itself accesses, making the separate footprint specification potentially redundant.

mldsa/poly.c
```diff
@@ -35,6 +36,10 @@ void poly_reduce(poly *a)
...
+    invariant(i <= MLDSA_N)
+    decreases(MLDSA_N - i))
```
manastasova: Since "decreases" is an optional contract element and appears inconsistent with the established pattern in mlkem-native and mldsa-native implementations, I'd recommend removing it for consistency, even though it's technically correct in this context.

manastasova: Introducing the assertion invariant`(array_bound(a->coeffs, 0, i, 0, MLDSA_Q))` would ensure that every element `a->coeffs` from `index 0 to i-1` remains within the range [0, MLSDA_Q-1] throughout each iteration of the loop.

mldsa/poly.h
```diff
@@ -9,13 +9,21 @@
...
 {
   int32_t coeffs[MLDSA_N];
 } poly;
 
 #define poly_reduce MLD_NAMESPACE(poly_reduce)
-void poly_reduce(poly *a);
+void poly_reduce(poly *a)
+__contract__(
+  requires(memory_no_alias(a, sizeof(poly)))
+  requires(forall(k0, 0, MLDSA_N, a->coeffs[k0] <= REDUCE_DOMAIN_MAX))
```
manastasova: You may want to add with a post-condition `ensures(array_bound(a->coeffs, 0, MLDSA_N, -REDUCE_RANGE_MAX, REDUCE_RANGE_MAX))` which will ensure the `a->coeffs` from index 0 to MLDSA_N-1 are all within the range [-REDUCE_RANGE_MAX, REDUCE_RANGE_MAX-1].
_Ensures_ clauses specifies the post-condition while _requires_ defines the preconditions.

manastasova (commented):
You might find it beneficial to check the mlkem-native CBMC proofs as a reference. They serve me as an excellent resource to validate and cross-check my proofs.

mldsa/poly.h
praveksharma: Since the proof for `poly_reduce` uses the function contract for `reduce32` I wasn't sure if repeating the conditions specified in the `reduce32` contract was necessary. I understand from this comment and [this one](https://github.com/pq-code-package/mldsa-native/pull/90#discussion_r2027864318) that this repetition is necessary, is that correct?

Would the condition `ensures(array_bound(a->coeffs, 0, MLDSA_N, -REDUCE_RANGE_MAX, REDUCE_RANGE_MAX + 1))` (with `REDUCE_RANGE_MAX = 6283009` defined in `reduce.h`) be more appropriate since `reduce32` ensures that `-REDUCE_RANGE_MAX <= reduce32(a->coeff[i]) < REDUCE_RANGE_MAX`?

praveksharma:
> You might find it beneficial to check the mlkem-native CBMC proofs as a reference. They serve me as an excellent resource to validate and cross-check my proofs.

Thank you for the detailed comments @manastasova, they are very helpful. 

I put up this incomplete PR because I wasn't making much progress working by myself. I'm not sure if you're on the Discord server so I'll repeat my comment from there here. The issue I am facing right now is that my proof is failing and CBMC reports an arithmetic overflow in `reduce32`. I'd appreciate any pointer to debug this.

mldsa/poly.c
hanno-becker: Current versions of CBMC are quite good at inferring the footprint, so while you should be prepared to add it yourself, I agree with @manastasova that you should try to remove it here.

hanno-becker: As a word of warning: `object_whole(...)` behaves 'nonlocally' in the sense that if you call this function with `a` pointing to a part of a struct, say, then `object_whole` will taint the whole struct. You almost never want this, so it's better to use `memory_slice`.

praveksharma: Thank you for the warning Hanno I'll keep that in mind.

praveksharma:
Thank you for your help @mkannwischer and the helpful comments @manastasova and @hanno-becker. I have ported over Matthias' proof from Discord and made a couple other changes. I'm marking the PR as ready.

mldsa/cbmc.h
```diff
@@ -13,6 +13,7 @@
 
 #define __contract__(x)
 #define __loop__(x)
+#define cassert(x)
```
mkannwischer: Can you change this to be in-line with what we do in mlkem-native? 
https://github.com/pq-code-package/mlkem-native/blob/c7fa785db37de10f41d6cb37517d155a7d268c5c/mlkem/debug.h#L82
Apologies

mldsa/params.h
```diff
@@ -54,6 +54,9 @@
 #define MLDSA_POLYT0_PACKEDBYTES 416
 #define MLDSA_POLYVECH_PACKEDBYTES (MLDSA_OMEGA + MLDSA_K)
 
+#define MLDSA_REDUCE_DOMAIN_MAX (INT32_MAX - (1 << 22))
+#define MLDSA_REDUCE_RANGE_MAX 6283009
```
mkannwischer: This is not a scheme parameter, but an implementation detail. 
Our convention is that scheme parameters start with `MLDSA_` and implementation details with `MLD_`
ALso I think we can keep this in reduce.h?

mkannwischer (changes requested):
Proof and spec itself looks good. Thanks @praveksharma!
Two small comments.

mldsa/cbmc.h
praveksharma: Apologies for the delay in getting back to you. I'm at a conference this week but I'll try get this done as soon as possible.

To clarify, is it the missing `mld_` namespacing which is the issue or would you like me to set up `debug.h` and `debug.c` as in mlkem-native?

mldsa/params.h
praveksharma: Thank you for the clarification regarding the `MLDSA_` and `MLD_` difference. My concern was that `REDUCE_DOMAIN_MAX` and `REDUCE_RANGE_MAX` would need to be defined separately in both `reduce.h` and `poly.h` (`poly.c` includes `reduce.h` but the `poly_reduce` contract in `poly.h` would need both defines again). Since `params.h` is common to both `reduce.h` and `poly.h` it seemed like the natural choice but I can see why that is incorrect. 

Are you suggesting I include `reduce.h` in `poly.h` or are the separate defines not an issue?

mldsa/cbmc.h
mkannwischer: Please add a debug.h/debug.c and define a mld_assert which get's turned into a run-time check if `MLD_DEBUG` is set, turned into a cassert if `CBMC` is set, and a no-op otherwise.

mldsa/params.h
mkannwischer: Including reduce.h in poly.h sounds the most sensible. Thank you.

mldsa/cbmc.h
mkannwischer: Actually, let's do this as a separate PR later - it's not essential now.

praveksharma: > Actually, let's do this as a separate PR later - it's not essential now.

Okay, that sounds good. Do you want me to open an issue to track this?

praveksharma:
Pushed another commit to switch from Bitwuzla to Z3.

mldsa/cbmc.h
mkannwischer: Yes please.

mkannwischer:
> Pushed another commit to switch from Bitwuzla to Z3.

Looking good to me. Can you please clean-up the commit history (e.g., by squashing your 6 commits into one)? Then this is good to merge.

mldsa/cbmc.h
praveksharma: Created #96.

praveksharma:
> Looking good to me. Can you please clean-up the commit history (e.g., by squashing your 6 commits into one)? Then this is good to merge.

Done!

mkannwischer (approved):
LGTM. Thanks @praveksharma for the work!

mldsa/reduce.c
```diff
@@ -23,6 +23,8 @@ int32_t montgomery_reduce(int64_t a)
 
   t = (int64_t)(int32_t)a * QINV;
   t = (a - (int64_t)t * MLDSA_Q) >> 32;
+
+  cassert((t - a) % MLDSA_Q == 0);
```
mkannwischer: This should be in `reduce32` not in `montgomery_reduce`

mkannwischer (changes requested):

mldsa/reduce.c
mkannwischer: I moved it.

mkannwischer (approved):

PR #91, opened 2025-04-04
CBMC: Add contract and proof for `montgomery_reduce` and `fqmul`

mkannwischer:
Resolves https://github.com/pq-code-package/mldsa-native/issues/14

This follows closely montgomery_reduce in mlkem-native: https://github.com/pq-code-package/mlkem-native/blob/83d85fe224bd6cf1b75f096a2b2fa01033b3dfda/mlkem/poly.h#L80

@hanno-becker @rod-chapman - can we prove anything else about this montgomery_reduction? The contract right now is rather weak. Do you remember why it is this weak in mlkem-native?

### DIFF
--- a/mldsa/ntt.c
+++ b/mldsa/ntt.c
@@ -7,6 +7,16 @@
 #include "params.h"
 #include "reduce.h"
 
+static int32_t mld_fqmul(int32_t a, int32_t b)
+__contract__(
+  requires(b > -MLDSA_Q_HALF && b < MLDSA_Q_HALF)
+  ensures(return_value > -MLDSA_Q && return_value < MLDSA_Q)
+)
+{
+  return montgomery_reduce((int64_t)a * (int64_t)b);
+  /* TODO: reason about bounds */
+}
+
 static const int32_t zetas[MLDSA_N] = {
     0,        25847,    -2608894, -518909,  237124,   -777960,  -876248,
     466468,   1826347,  2353451,  -359251,  -2091905, 3119733,  -2884855,
@@ -68,7 +78,7 @@ void ntt(int32_t a[MLDSA_N])
       zeta = zetas[++k];
       for (j = start; j < start + len; ++j)
       {
-        t = montgomery_reduce((int64_t)zeta * a[j + len]);
+        t = mld_fqmul(a[j + len], zeta);
         a[j + len] = a[j] - t;
         a[j] = a[j] + t;
       }
@@ -104,13 +114,13 @@ void invntt_tomont(int32_t a[MLDSA_N])
         t = a[j];
         a[j] = t + a[j + len];
         a[j + len] = t - a[j + len];
-        a[j + len] = montgomery_reduce((int64_t)zeta * a[j + len]);
+        a[j + len] = mld_fqmul(a[j + len], zeta);
       }
     }
   }
 
   for (j = 0; j < MLDSA_N; ++j)
   {
-    a[j] = montgomery_reduce((int64_t)f * a[j]);
+    a[j] = mld_fqmul(a[j], f);
   }
 }
--- a/mldsa/reduce.h
+++ b/mldsa/reduce.h
@@ -10,12 +10,27 @@
 #include "params.h"
 
 #define MONT -4186625 /* 2^32 % MLDSA_Q */
-#define QINV 58728449 /* q^(-1) mod 2^32 */
 #define REDUCE_DOMAIN_MAX (INT32_MAX - (1 << 22))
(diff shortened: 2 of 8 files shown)


rod-chapman:
The comment in MLKEM seems reasonable. There are only 2 call sites.
In one case, it's inlined, so the caller's post-condition embodies whatever is needed.
In the other case, no postcondition is required to prove type-safety.

rod-chapman:
Hmmm... we _will_ need a range for the output of montgomery_reduce() when we come to prove ntt() and invntt_tomont(), but perhaps the call to montgomery_reduce() can be inlined and wrapped in an new fqmul() function, as in MLKEM?

rod-chapman:
And similarly for poly_pointwise_montgomery() in poly.c

mkannwischer:
> Hmmm... we _will_ need a range for the output of montgomery_reduce() when we come to prove ntt() and invntt_tomont(), but perhaps the call to montgomery_reduce() can be inlined and wrapped in an new fqmul() function, as in MLKEM?

Okay - I'll try to add an `fqmul` as in mlkem-native and attempt to prove the (-q, q) bounds.

mkannwischer:
> > Hmmm... we _will_ need a range for the output of montgomery_reduce() when we come to prove ntt() and invntt_tomont(), but perhaps the call to montgomery_reduce() can be inlined and wrapped in an new fqmul() function, as in MLKEM?
> 
> Okay - I'll try to add an `fqmul` as in mlkem-native and attempt to prove the (-q, q) bounds.

This was suprisingly simple. I added the fqmul and it's proof.

rod-chapman:
Reviewing now

rod-chapman:
For efficiency, we want montgomery_reduce() to be amenable to inlining for compilation. In mlkem-native, it is declared _and_ defined in poly.h so it can be inlined into call sites. Should we do the same here, or leave performance work until later?

mldsa/reduce.h
```diff
@@ -10,12 +10,27 @@
...
+ *              -MLDSA_Q < r < MLDSA_Q.
+ *
+ * Arguments:   - int64_t: finite field element a
+ *
+ * Returns r.
+ **************************************************/
+int32_t montgomery_reduce(int64_t a)
+__contract__(
+  requires(a >= -MONTGOMERY_REDUCE_DOMAIN_MAX && a <= MONTGOMERY_REDUCE_DOMAIN_MAX)
+);
```
rod-chapman, line 33: Declare _and_ define here for inlining into call sites?

mldsa/ntt.c
```diff
@@ -7,6 +7,16 @@
 #include "params.h"
 #include "reduce.h"
 
+static int32_t mld_fqmul(int32_t a, int32_t b)
```
rod-chapman, line 10: Declare this in a .h file so it can called from poly.c?

rod-chapman (changes requested):

mkannwischer:
> For efficiency, we want montgomery_reduce() to be amenable to inlining for compilation. In mlkem-native, it is declared _and_ defined in poly.h so it can be inlined into call sites. Should we do the same here, or leave performance work until later?

I would like to have proper benchmarking in place before we do that. 

> In poly.c, we should probably call fqmul() in poly_pointwise_montgomery()
> 
> ```
>     c->coeffs[i] = montgomery_reduce((int64_t)a->coeffs[i] * b->coeffs[i]);
> ```
> 
> becomes
> 
> ```
>     c->coeffs[i] = fqmul(a->coeffs[i] * b->coeffs[i]);
> ```
> 
> right? That would mean declaring fqmul() somewhere visible, rather that hidden in ntt.c

Hmmmm - no. The pre-condition that we used for fqmul does not hold in the pointwise multiplication. I'd like to keep this separate at least until we have a proof for poly_pointwise_montgomery.

mldsa/reduce.h
mkannwischer: Let's do that later when benchmarking is in place.

mldsa/ntt.c
mkannwischer: Hmmmm - no. The pre-condition that we used for fqmul does not hold in the pointwise multiplication. I'd like to keep this separate at least until we have a proof for poly_pointwise_montgomery.

hanno-becker:
> Hmmmm - no. The pre-condition that we used for fqmul does not hold in the pointwise multiplication. I'd like to keep this separate at least until we have a proof for poly_pointwise_montgomery.

Agree.

rod-chapman:
Thanks for responses.

rod-chapman (approved):
Looks good now.

PR #87, opened 2025-04-02
CBMC: Add contract and proof for `decompose` and `poly_decompose`

mkannwischer:
Resolves #19

PR #92, opened 2025-04-06
CBMC: Add contract and proof for `polyveck_decompose`

mkannwischer:
Depends on #87


proofs/cbmc/poly_decompose/Makefile
```diff
@@ -26,6 +26,7 @@ USE_DYNAMIC_FRAMES=1
 # Disable any setting of EXTERNAL_SAT_SOLVER, and choose SMT backend instead
 EXTERNAL_SAT_SOLVER=
 CBMCFLAGS=--smt2
+CBMC_FLAG_UNSIGNED_OVERFLOW_CHECK=
```
mkannwischer: @rod-chapman - do you know if still should still be needed?

rod-chapman: Aha! Well spotted - let me restore the original and re-run...

rod-chapman: It's fine with the original flags, which is good news. Rebased, squashed, and pushed.

mkannwischer:
Thanks @rod-chapman - this looks good to me now.

rod-chapman (approved):
Looks good new. Please merge.

PR #93, opened 2025-04-07
Proofs of `use_hint()` and `poly_use_hint()`

### LINKED ISSUE #20: CBMC: Prove `use_hint` and `poly_use_hint`

rod-chapman:
Fixes #20 

Adds contracts and CBMC proofs of use_hint() and poly_use_hint()

polyveck_use_hint() will be attempted separately.

All tests, proofs, and lint OK.

### DIFF
--- a/mldsa/poly.h
+++ b/mldsa/poly.h
@@ -84,18 +84,31 @@ __contract__(
 );
 
 #define poly_use_hint MLD_NAMESPACE(poly_use_hint)
-void poly_use_hint(poly *b, const poly *a, const poly *h);
+void poly_use_hint(poly *b, const poly *a, const poly *h)
+__contract__(
+  requires(memory_no_alias(a,  sizeof(poly)))
+  requires(memory_no_alias(b, sizeof(poly)))
+  requires(memory_no_alias(h, sizeof(poly)))
+  requires(array_bound(a->coeffs, 0, MLDSA_N, 0, MLDSA_Q))
+  requires(array_bound(h->coeffs, 0, MLDSA_N, 0, 2))
+  assigns(memory_slice(b, sizeof(poly)))
+  ensures(array_bound(b->coeffs, 0, MLDSA_N, 0, (MLDSA_Q-1)/(2*MLDSA_GAMMA2)))
+);
 
 #define poly_chknorm MLD_NAMESPACE(poly_chknorm)
 int poly_chknorm(const poly *a, int32_t B);
+
 #define poly_uniform MLD_NAMESPACE(poly_uniform)
 void poly_uniform(poly *a, const uint8_t seed[MLDSA_SEEDBYTES], uint16_t nonce);
+
 #define poly_uniform_eta MLD_NAMESPACE(poly_uniform_eta)
 void poly_uniform_eta(poly *a, const uint8_t seed[MLDSA_CRHBYTES],
                       uint16_t nonce);
+
 #define poly_uniform_gamma1 MLD_NAMESPACE(poly_uniform_gamma1)
 void poly_uniform_gamma1(poly *a, const uint8_t seed[MLDSA_CRHBYTES],
                          uint16_t nonce);
+
 #define poly_challenge MLD_NAMESPACE(poly_challenge)
 void poly_challenge(poly *c, const uint8_t seed[MLDSA_CTILDEBYTES]);
 
--- a/mldsa/rounding.c
+++ b/mldsa/rounding.c
@@ -16,12 +16,22 @@ void power2round(int32_t *a0, int32_t *a1, const int32_t a)
 void decompose(int32_t *a0, int32_t *a1, int32_t a)
 {
   *a1 = (a + 127) >> 7;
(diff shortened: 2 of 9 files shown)


mldsa/poly.h
```diff
@@ -84,18 +84,31 @@ __contract__(
 );
 
 #define poly_use_hint MLD_NAMESPACE(poly_use_hint)
-void poly_use_hint(poly *b, const poly *a, const poly *h);
+void poly_use_hint(poly *b, const poly *a, const poly *h)
+__contract__(
+  requires(memory_no_alias(a,  sizeof(poly)))
+  requires(memory_no_alias(b, sizeof(poly)))
+  requires(memory_no_alias(h, sizeof(poly)))
+  requires(array_bound(a->coeffs, 0, MLDSA_N, 0, MLDSA_Q))
```
manastasova, line 92: If `array_bound(arr, idx_low, idx_high, value_low, value_high)` asserts that the values of the array `arr` within`[idx_low, ..., idx_high - 1]` are within the range `[value_low, ..., value_high]`, then for parameter `a`, shouldn't we use `requires(array_bound(a->coeffs, 0, MLDSA_N, 0, MLDSA_Q-1))` (also for `h->coeffs in [0;2)`)? 
I notice that mlkem-native uses a similar pattern with bounds `[0, MLKEM_Q]`. I'm wondering if there's something I'm missing about how these bounds should be specified.

mldsa/rounding.h
```diff
@@ -76,6 +76,11 @@ __contract__(
 );
 
 #define use_hint MLD_NAMESPACE(use_hint)
-int32_t use_hint(int32_t a, unsigned int hint);
+int32_t use_hint(int32_t a, unsigned int hint)
+__contract__(
+  requires(hint >= 0 && hint <= 1)
+  requires(a >= 0 && a < MLDSA_Q)
+  ensures(return_value >= 0 && return_value < (MLDSA_Q-1)/(2*MLDSA_GAMMA2))
```
manastasova, line 83: Based on Algorithm 40 in FIPS204, I believe the `return_value` should fall within the range `0 <= return_value <= (q-1)/(2*MLDSA_GAMMA2)`. To properly specify this constraint in the code, we would need to change the post-condition check to `return_value <=(MLDSA_Q-1)/(2*MLDSA_GAMMA2)`.
```suggestion
  ensures(return_value >= 0 && return_value <= (MLDSA_Q-1)/(2*MLDSA_GAMMA2))
```

mldsa/poly.h
rod-chapman: Look at the definition of array_bound_core macro in cbmc.h. It uses "<" for both the upper-bound of the quantified variable AND the range of the element.

mldsa/rounding.h
rod-chapman: I wrote and exhaustive test case for decompose(), and it only ever returns a1 in 0 .. 15 (inclusive) when MLDSA_MODE=3 or 0 .. 43 when MLDSA_MODE=2, so it looks like the reference implementation has a slightly stronger postcondition that the specification.

mldsa/poly.h
manastasova: Thank you! I actually just checked the macro itself in the` cbmc.h` file and now I understand the proof contract. 
My confusion is based on the https://github.com/pq-code-package/mlkem-native/blob/main/proofs/cbmc/proof_guide.md which defines the `array_bound` as inclusive for both lower and upper bounds.

mldsa/rounding.h
rod-chapman: The code is ugly (it's doing shifts and bitwise operations on signed integers which are, strictly speaking, implementation-defined), but if we assume the "obvious" 2's complement representation, then you can see that the final assignment to a1 in decompose() can only ever yield 0..15 or 0..43 for each case.

rod-chapman: I just added some intermediate assertions in decompose() that justify the final range of *a1
That, in turn, explains the post-condition of use_hint()

Does that make sense?

mldsa/poly.h
rod-chapman: Arrgh! Sorry.. that's a bug in proof_guide.md from earlier (when the upper-bound _was_ inclusive.). We changed it to exclusive upper bounds some time ago. I will raise a PR on the Proof Guide immediately.

mldsa/rounding.c
```diff
@@ -16,12 +16,22 @@ void power2round(int32_t *a0, int32_t *a1, const int32_t a)
 void decompose(int32_t *a0, int32_t *a1, int32_t a)
 {
   *a1 = (a + 127) >> 7;
+  /* We know a >= 0 and a > MLDSA_Q, so... */
```
mkannwischer:
```suggestion
  /* We know a >= 0 and a < MLDSA_Q, so... */
```

mkannwischer (approved):
LGTM - thanks @rod-chapman. One typo.

mldsa/rounding.c
rod-chapman: Fixed.

PR #95, opened 2025-04-08
Remove unused profiling from poly.c

mkannwischer:
The reference implementation comes with some built-in profiling that is inlined into most functions in poly.c.
This is unused in mldsa-native and incompatible with the way our component benchmarks work. For some functions the profiling has already been removed when CBMC proofs were added.
This commit removes all profiling.


rod-chapman (approved):
Looks good

PR #97, opened 2025-04-08
Move and clarify comments in rounding.c to rounding.h

rod-chapman:
Moves comments from rounding.c to their correct place in rounding.h
Clarifies wording in 2 places.

Lint OK. Tests OK.

### DIFF
--- a/mldsa/rounding.h
+++ b/mldsa/rounding.h
@@ -23,7 +23,8 @@
  *              - int32_t *a0: pointer to output element a0
  *              - int32_t *a1: pointer to output element a1
  *
- * Reference: a1 is passed as a return value instead
+ * In the reference implementation, a1 is passed as a
+ * return value instead.
  **************************************************/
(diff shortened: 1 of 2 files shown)


mldsa/rounding.h
```diff
@@ -23,7 +23,8 @@
  *              - int32_t *a0: pointer to output element a0
  *              - int32_t *a1: pointer to output element a1
  *
- * Reference: a1 is passed as a return value instead
+ * In the reference implementation, a1 is passed as a
```
hanno-becker: Keep the `Reference: ...` prefix for easy grepping

rod-chapman: Fixed

mkannwischer:
Thanks - I'll do a PR for all the other comments. That has been bothering me for some time.

hanno-becker (approved):

rod-chapman:
I'd like to have this one up-to-standard now so we can use it as an example for Jake, Mila, and other contributors.

PR #94, opened 2025-04-07
Add spec and proof for rej_uniform

manastasova:
This PR adds the contract and proof for rej_uniform function.

Notes::
- `rej_uniform` is static, thus, `CHECK_FUNCTION_CONTRACTS` is defined as `rej_uniform` instead of $`(MLD_NAMESPACE)rej_uniform` in the `Makefile`.
- `rej_uniform` is static, thus, additional assumptions, such as `requires(len <= MLDSA_N && buflen <= (POLY_UNIFORM_NBLOCKS * STREAM128_BLOCKBYTES))` are made. This accelerates the performance of CBMC proofs.

Solves https://github.com/pq-code-package/mldsa-native/issues/25.

PR #98, opened 2025-04-08
Move all docstrings to header files

mkannwischer:
Following https://github.com/pq-code-package/mldsa-native/pull/97, this commit moves _all_ comments documenting function signatures into the corresponding header files.
Note that there are still some functions that do not have a documenting comment.
I have not changed any comments except for alignment.

### DIFF
--- a/mldsa/poly.c
+++ b/mldsa/poly.c
@@ -10,14 +10,6 @@
 #include "rounding.h"
 #include "symmetric.h"
 
-/*************************************************
- * Name:        poly_reduce
- *
- * Description: Inplace reduction of all coefficients of polynomial to
- *              representative in [-6283008,6283008].
- *
- * Arguments:   - poly *a: pointer to input/output polynomial
- **************************************************/
 void poly_reduce(poly *a)
 {
   unsigned int i;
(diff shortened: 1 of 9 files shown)


mldsa/poly.c
```diff
@@ -359,18 +228,8 @@ __contract__(
...
- * Description: Sample polynomial with uniformly random coefficients
- *              in [0,MLDSA_Q-1] by performing rejection sampling on the
- *              output stream of SHAKE128(seed|nonce)
- *
- * Arguments:   - poly *a: pointer to output polynomial
- *              - const uint8_t seed[]: byte array with seed of length
- *MLDSA_SEEDBYTES
- *              - uint16_t nonce: 2-byte nonce
- **************************************************/
+#define POLY_UNIFORM_NBLOCKS \
```
rod-chapman: This macro appears to be defined twice now - lines 191 and 231. I think the earlier one is needed so it can be used in contracts?

mkannwischer: Good catch - that was moved in #94 and the rebase didn't work as intended.
I fixed it.

mkannwischer:
Rebased and addressed the comments.

rod-chapman:
Looks good.

rod-chapman (approved):
Looks good. Please merge.

PR #99, opened 2025-04-08
Add proofs of caddq() and freeze() in reduce.[hc]

### LINKED ISSUE #16: CBMC: Prove `caddq`

### LINKED ISSUE #17: CBMC: Prove `freeze`

rod-chapman:
Fixes #16 
Fixes #17 

Adds proofs of freeze() and caddq() in reduce.c
Also moves comments into reduce.h as appropriate.

Lint OK. Tests OK.


mldsa/reduce.h
```diff
@@ -41,9 +52,40 @@ __contract__(
...
+ *
+ * Arguments:   - int32_t: finite field element a
+ *
+ * Returns r.
+ **************************************************/
+int32_t freeze(int32_t a)
+__contract__(
+  requires(a <= REDUCE_DOMAIN_MAX)
+  ensures(return_value >= 0)
+  ensures(return_value < MLDSA_Q)
```
mkannwischer: Can we add a cassert at the end of the function body that the return value is actually congruent to a mod q?

proofs/cbmc/freeze/Makefile
```diff
@@ -0,0 +1,54 @@
...
+INCLUDES +=
+
+REMOVE_FUNCTION_BODY +=
+UNWINDSET +=
+
+PROOF_SOURCES += $(PROOFDIR)/$(HARNESS_FILE).c
+PROJECT_SOURCES += $(SRCDIR)/mldsa/reduce.c
+
+CHECK_FUNCTION_CONTRACTS=$(MLD_NAMESPACE)freeze
+USE_FUNCTION_CONTRACTS=
```
mkannwischer: Would it make sense to actually make use of the contracts for caddq and reduce32 as a sanity check? Or is it preferable to inline everything?

mkannwischer (changes requested):
Thanks @rod-chapman!

mldsa/reduce.h
rod-chapman: Possibly, but I can't work out how to express that using C's % operator...
We'd also need a stronger ensures contract on the spec of reduce32()

proofs/cbmc/freeze/Makefile
rod-chapman: Yes... my mistake. It should be using the contracts.

rod-chapman: Pushed.

mldsa/reduce.h
mkannwischer: `cassert((t - a) % MLDSA_Q == 0);`?
that works for `reduce32`

mkannwischer: ahhh but of course you need to add that to the contracts of `caddq` and `reduce32`

rod-chapman: Are we really going to need a strong postcondition on freeze()? Not sure...

rod-chapman: I can't find a single call to freeze() anywhere.... perhaps we should just delete it?!?

mkannwischer: Very good point. Please delete it.

mkannwischer (approved):
LGTM. Thanks! Good catch with the `freeze` not being used.

rod-chapman:
I have removed freeze(), squashed, and pushed.

PR #101, opened 2025-04-08
Refactor and prove type-safety of forward NTT

rod-chapman:
Refactors reference ntt() function into 3 functions, as per the pattern in mlkem-native:
  mld_ntt_butterfly_block()
  mld_ntt_layer()
  ntt()

Adds contracts and proofs of these 3 functions.
Adds and moves comments as appropriate.
Lint OK. Tests OK.


mkannwischer (approved):
Thanks @rod-chapman. LGTM.

PR #139, opened 2025-04-09
Simplification and proof of unpack_sig() and supporting functions.

rod-chapman:
Re-factors unpack_sig() to make it much simplifer, and isolates unpacking of hints into new, static unpack_hints() function

Adds supporting functions and their proofs:
 poly_clear() - zeroes a polynomial
 polyveck_clear() - zeroes a vector of K polynomials
 polyvecl_unpack_z() - unpacks the L z components of a signature
 unpack_hints() - the inverse of the "hint packing" algorithm in pack_sig()

All proofs OK. All tests OK. Lint OK.

### DIFF
--- a/mldsa/packing.c
+++ b/mldsa/packing.c
@@ -152,58 +152,105 @@ void pack_sig(uint8_t sig[CRYPTO_BYTES], const uint8_t c[MLDSA_CTILDEBYTES],
   }
 }
 
-int unpack_sig(uint8_t c[MLDSA_CTILDEBYTES], polyvecl *z, polyveck *h,
-               const uint8_t sig[CRYPTO_BYTES])
+/*************************************************
+ * Name:        unpack_hints
+ *
+ * Description: Unpack raw hint bytes into a polyveck
+ *              struct
+ *
+ * Arguments:   - polyveck *h: pointer to output hint vector h
+ *              - const uint8_t packed_hints[MLDSA_POLYVECH_PACKEDBYTES]:
+ *                raw hint bytes
+ *
+ * Returns 1 in case of malformed hints; otherwise 0.
+ **************************************************/
+static int unpack_hints(polyveck *h,
+                        const uint8_t packed_hints[MLDSA_POLYVECH_PACKEDBYTES])
+__contract__(
+  requires(memory_no_alias(packed_hints, MLDSA_POLYVECH_PACKEDBYTES))
+  requires(memory_no_alias(h, sizeof(polyveck)))
+  assigns(object_whole(h))
+  /* All returned coefficients are either 0 or 1 */
+  ensures(forall(k1, 0, MLDSA_K,
+    array_bound(h->vec[k1].coeffs, 0, MLDSA_N, 0, 2)))
+  ensures(return_value >= 0 && return_value <= 1)
+)
 {
-  unsigned int i, j, k;
+  unsigned int i, j;
+  unsigned int old_hint_count;
 
-  for (i = 0; i < MLDSA_CTILDEBYTES; ++i)
-  {
-    c[i] = sig[i];
-  }
-  sig += MLDSA_CTILDEBYTES;
+  /* Set all coefficients of all polynomials to 0.    */
+  /* Only those that are actually non-zero hints will */
+  /* be overwritten below.                            */
+  memset(h, 0, sizeof(polyveck));
 
-  for (i = 0; i < MLDSA_L; ++i)
-  {
-    polyz_unpack(&z->vec[i], sig + i * MLDSA_POLYZ_PACKEDBYTES);
-  }
-  sig += MLDSA_L * MLDSA_POLYZ_PACKEDBYTES;
-
-  /* Decode h */
-  k = 0;
+  old_hint_count = 0;
   for (i = 0; i < MLDSA_K; ++i)
+  __loop__(
+    invariant(i <= MLDSA_K)
+    /* Maintain the post-condition */
+    invariant(forall(k1, 0, MLDSA_K, array_bound(h->vec[k1].coeffs, 0, MLDSA_N, 0, 2)))
+  )
   {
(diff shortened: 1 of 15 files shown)


mldsa/packing.c
```diff
@@ -151,58 +151,136 @@ void pack_sig(uint8_t sig[CRYPTO_BYTES], const uint8_t c[MLDSA_CTILDEBYTES],
...
 
-  for (i = 0; i < MLDSA_CTILDEBYTES; ++i)
-  {
-    c[i] = sig[i];
-  }
-  sig += MLDSA_CTILDEBYTES;
+  /* Set all coefficients of all polynomials to 0.    */
+  /* Only those that are actually non-zero hints will */
+  /* be overwritten below.                            */
+  polyveck_clear(h);
```
mkannwischer: Replacing this by 
```
memset(h, 0, sizeof(polyveck));
```
results in the same proof time and we save quite a bit of code. Any reason for not doing that instead?

mldsa/packing.c
```diff
@@ -151,58 +151,136 @@ void pack_sig(uint8_t sig[CRYPTO_BYTES], const uint8_t c[MLDSA_CTILDEBYTES],
...
 
-  for (i = 0; i < MLDSA_L; ++i)
-  {
-    polyz_unpack(&z->vec[i], sig + i * MLDSA_POLYZ_PACKEDBYTES);
-  }
-  sig += MLDSA_L * MLDSA_POLYZ_PACKEDBYTES;
+  /* Assert that all values of h are either 0 or 1                   */
+  /* This establishes the post-condition before we fill in any hints */
+  cassert(
+      forall(k1, 0, MLDSA_K, array_bound(h->vec[k1].coeffs, 0, MLDSA_N, 0, 2)));
```
mkannwischer: This assertion appears to be not very useful? Obviously all coeffs are 0 after clearing them. 
Removing the cassert results in a slightly faster proof.

mkannwischer (changes requested):
Thanks @rod-chapman. This looks like a lot of work. A couple of suggestions.
Can we try to keep `unpack_hints` as close as possible to what the reference implementation is doing?
For example, the following version of the main loop is much simpler and closer to the reference implementation and proves in the same time as your version: 
```c
old_hint_count = 0;
  for (i = 0; i < MLDSA_K; ++i)
  __loop__(
    invariant(i <= MLDSA_K)
    /* Maintain the post-condition */
    invariant(forall(k1, 0, MLDSA_K, array_bound(h->vec[k1].coeffs, 0, MLDSA_N, 0, 2)))
  )
  {
    /* Grab the hint count for the i'th polynomial */
    const unsigned int new_hint_count = packed_hints[MLDSA_OMEGA + i];

    /* new_hint_count must increase or stay the same, but also remain */
    /* less than or equal to MLDSA_OMEGA                              */
    if (new_hint_count < old_hint_count || new_hint_count > MLDSA_OMEGA)
    {
      /* Error - new_hint_count is invalid */
      return 1;
    }

    /* If new_hint_count == old_hint_count, then this polynomial has */
    /* zero hints, so this loop executes zero times and we move      */
    /* straight on to the next polynomial.                           */
    for (j = old_hint_count; j < new_hint_count; ++j)
    __loop__(
        invariant(i <= MLDSA_K)
        /* Maintain the post-condition */
        invariant(forall(k1, 0, MLDSA_K, array_bound(h->vec[k1].coeffs, 0, MLDSA_N, 0, 2)))
      )
    {
      const uint8_t this_hint_index = packed_hints[j];

      /* Coefficients must be ordered for strong unforgeability */
      if (j > old_hint_count && this_hint_index <= packed_hints[j - 1])
      {
        return 1;
      }
      h->vec[i].coeffs[this_hint_index] = 1;
    }

    old_hint_count = new_hint_count;
  }
  ```

mldsa/packing.c
hanno-becker: Agree, I don't think we need dedicated `xxx_clear` functions. Once we ported `mlk_zeroize` to ML-DSA, this could be used alternatively to `memset`.

rod-chapman: The TrailOfBits audits that I have seen object to memset(). They (rightly) prefer memset_s(), but that's not available in C90. I prefer to wrap memset() calls in a proper abstraction that is named to indicate their purpose, especially if that code is repeated.

I am also not sure that mlk_zeroize() is really right here. We are setting *h to "all zero" because it's required for functional correctness, NOT for preventing unintentional data leakage. (Put another way, you could "zeroize" a chunk of memory all 0xFF and claim that's all good for non-functional security, but would be wrong for functional correctness. I don't think we should overload the two use cases.)

mldsa/packing.c
rod-chapman: Noted - this was more of a "documentation assertion" that one that is strictly required for proof. I will remove

mldsa/packing.c
rod-chapman: OK... this "clear a polyveck to all zero" pattern appears exactly once only, so a single call to memset() is reasonable.

rod-chapman:
Review comments addressed. Please take another look.

mldsa/packing.c
mkannwischer: I agree that `mld_zeroize()` may not be the right choice - that is specifically meant for stack cleansing. We probably want to add a `mld_memset` as a wrapper to memset or memset_s and add similar wrappers for other stdlib functions. There is an issue for this in mlkem-native (https://github.com/pq-code-package/mlkem-native/issues/550) and we should do something similar in mldsa-native. I see this as low prio though.

mkannwischer (changes requested):
Thanks @rod-chapman for the changes. It looks good to me now. 
Please squash your changes into one commit, then this is good to go.

rod-chapman:
Rebased, squashed and pushed. CI looks good.

mkannwischer (approved):
Thanks @rod-chapman!

PR #163, opened 2025-04-13
autogen: Autogenerate header guards

### LINKED ISSUE #155: Port autogeneration of header guards from mlkem-native
mlkem-native uses the `autogen` script to automatically generate guards for all headers. This should be ported to mldsa-native, taking care of part of #11.

hanno-becker:
* Resolves #155 

Following mlkem-native, this commit extends `scripts/autogen` to automatically generate guards for all headers in `mldsa/*`.


mkannwischer (approved):
LGTM @hanno-becker - thanks. I added it to linting as well - that was previously disabled.

Issue #83, opened 2025-03-31
Consider autogeneration of #else ... #endif comments (port from mlkem-native)


mkannwischer:
See https://github.com/pq-code-package/mlkem-native/issues/917 https://github.com/pq-code-package/mlkem-native/pull/921

PR #164, opened 2025-04-13
autogen: Autogenerate guard comments for `#if .. #else .. #endif` blocks

hanno-becker:
* Based on #163 
* Resolves https://github.com/pq-code-package/mldsa-native/issues/83

Following mlkem-native, this PR extends `scripts/autogen` with
autogeneration of guard-comments for `#if .. #else .. #endif` blocks.

There are multiple places in the code base where a parameter-set
preprocessor branch is conducted not by checking MLDSA_MODE but
the specific value of one of the derived constants, such as
MLDSA_GAMMA1 or MLDSA_GAMMA2. autogen cannot handle those.
This PR thus replaces those branch conditions by equality check
for MLDSA_MODE. It also reorders the branches so that MLDSA_MODE==2
comes first.


mkannwischer (approved):
Great progress. Thanks @hanno-becker!

PR #165, opened 2025-04-14
Add spec and proofs for load64 and store64

manastasova:
This PR adds the spec and proofs for load64 and store64.

Notes::
	- Change the signature of function `store64` from `static void store64(uint8_t x[8], uint64_t u)` to `static void store64(uint8_t x[8], const uint64_t u)`, marking the input as a `const`.
	-  Add explicit `& 0xFF` before casting `uint64_t u` to `uint8_t x[i]` to prevent overlflows

### DIFF
--- a/mldsa/fips202/fips202.c
+++ b/mldsa/fips202/fips202.c
@@ -25,11 +25,17 @@
  * Returns the loaded 64-bit unsigned integer
  **************************************************/
 static uint64_t load64(const uint8_t x[8])
+__contract__(
+  requires(memory_no_alias(x, sizeof(uint8_t) * 8))
+)
 {
   unsigned int i;
   uint64_t r = 0;
 
   for (i = 0; i < 8; i++)
+  __loop__(
+    invariant(i <= 8)
+  )
   {
     r |= (uint64_t)x[i] << 8 * i;
   }
@@ -47,12 +53,20 @@ static uint64_t load64(const uint8_t x[8])
  *              - uint64_t u: input 64-bit unsigned integer
  **************************************************/
 static void store64(uint8_t x[8], uint64_t u)
+__contract__(
+  requires(memory_no_alias(x, sizeof(uint8_t) * 8))
+  assigns(memory_slice(x, sizeof(uint8_t) * 8))
+)
 {
   unsigned int i;
 
   for (i = 0; i < 8; i++)
+  __loop__(
+    invariant(i <= 8)
+  )
   {
-    x[i] = u >> 8 * i;
+    /* Explicitly truncate to uint8_t */
+    x[i] = (uint8_t)((u >> (8 * i)) & 0xFF);
   }
 }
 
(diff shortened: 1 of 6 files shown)


mldsa/fips202/fips202.c
```diff
@@ -46,13 +52,21 @@ static uint64_t load64(const uint8_t x[8])
...
 {
   unsigned int i;
 
   for (i = 0; i < 8; i++)
+  __loop__(
+    invariant(i <= 8)
+  )
   {
-    x[i] = u >> 8 * i;
+    /* Make sure that u >> (8 * i) will not overflow uint8_t x[i] */
```
hanno-becker: Suggestion: 
```suggestion
    /* Explicitly truncate to uint8_t */
```

mldsa/fips202/fips202.c
```diff
@@ -46,13 +52,21 @@ static uint64_t load64(const uint8_t x[8])
  * Arguments:   - uint8_t *x: pointer to the output byte array (allocated)
  *              - uint64_t u: input 64-bit unsigned integer
  **************************************************/
-static void store64(uint8_t x[8], uint64_t u)
+static void store64(uint8_t x[8], const uint64_t u)
```
hanno-becker: `const` is not too common for arguments passed by value, not by reference. The qualification only restricts what the implementation can do, but it has no impact on the caller. I would prefer to leave it as it was, unless we want to make a general push to mark all unmodified by-value arguments `const` -- which I'm not sure is worth it.

proofs/cbmc/store64/store64_harness.c
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2025 The mldsa-native project authors
+// SPDX-License-Identifier: Apache-2.0
+
+#include "fips202/fips202.h"
+
+extern void store64(uint8_t x[8], const uint64_t u);
```
hanno-becker: As mentioned above, the `const` has no meaning for external signature because `u` is passed by value.

hanno-becker (changes requested):
Thanks @manastasova! Can you remove the `const` annotation and slightly clarify one of the comments, as indicated? Otherwise LGTM

mldsa/fips202/fips202.c
manastasova: I completely agree and just fixed it. Thank you!

mldsa/fips202/fips202.c
manastasova: Thanks!

proofs/cbmc/store64/store64_harness.c
manastasova: Thanks again!

manastasova:
> Thanks @manastasova! Can you remove the `const` annotation and slightly clarify one of the comments, as indicated?
> 
> Also, can you squash the commits, and make sure that lines in the body of the commit message are no longer than ~75 characters per line?
> 
> Otherwise LGTM

Thanks a lot @hanno-becker! Just addressed your comments.

hanno-becker (changes requested):
Thank you @manastasova, the code looks good now! Can you please fix the commit message (no lines > 75 characters, typos)? Then we're good to go.

hanno-becker (approved):

PR #167, opened 2025-04-14
Add spec and proofs for Keccak dependent functions

manastasova:
This PR adds the spec and proofs for different Keccak dependent functions including:  

 - `KeccakF1600_StatePermute`
 - `keccak_absorb_once`
 - `shake128_absorb_once`
 - `shake256_absorb_once`
 - `sha3_256`
 - `sha3_512`


mkannwischer:
This works fine on my machine, but fails in CI.
I rebased on top of https://github.com/pq-code-package/mldsa-native/pull/178 to check if running it on a bigger EC2 box solves the problem

PR #168, opened 2025-04-16
Add common.h and sys.h

### LINKED ISSUE #156: Port sys.h from mlkem-native
Port `sys.h` from mlkem-native and use the macros defined therein (`MLD_ALIGN`, `MLD_INLINE`, ...) throughout.

hanno-becker:
* Resolves #156 

Following mlkem-native, this commit adds two new header files, common.h and sys.h.

sys.h conducts system detection and sets various helper macros such as MLD_ALIGN accordingly. It is identical to the sys.h header in mlkem-native, modulo the MLK_ -> MLD_ renaming.

The common header file common.h should be included by all compilation units, including assembly files. It so far pulls in params.h and sys.h, but will be extended in the future.

All includes of params.h are replaced by includes of common.h.

### DIFF
--- a/mldsa/common.h
+++ b/mldsa/common.h
@@ -1,0 +1,12 @@
+/*
+ * Copyright (c) 2025 The mldsa-native project authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef MLD_COMMON_H
+#define MLD_COMMON_H
+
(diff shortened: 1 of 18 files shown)


mkannwischer (approved):
LGTM. Thanks @hanno-becker

PR #151, opened 2025-04-12
CBMC: Add contracts and proofs for various polyvec functions

mkannwischer:
- Depends on https://github.com/pq-code-package/mldsa-native/pull/140
- Resolves https://github.com/pq-code-package/mldsa-native/issues/116
- Resolves https://github.com/pq-code-package/mldsa-native/issues/117
- Resolves https://github.com/pq-code-package/mldsa-native/issues/124
- Resolves https://github.com/pq-code-package/mldsa-native/issues/126
- Resolves https://github.com/pq-code-package/mldsa-native/issues/127
- Resolves https://github.com/pq-code-package/mldsa-native/issues/136
- Resolves https://github.com/pq-code-package/mldsa-native/issues/135

`polyvecl_add`, `polyveck_add`, `polyveck_sub`,`polyveck_use_hint`, `polyveck_reduce`, `polyvecl_reduce` were too slow when using unrolling - I instead did a direct proof with the workaround from #92. I added them to #102 to remove the whole-struct assignments once the corresponding CBMC issue is resolved.

PR #171, opened 2025-04-16
CBMC: Add contract and proof for polyveck_reduce

jakemas:
resolves https://github.com/pq-code-package/mldsa-native/issues/124


mkannwischer:
That function is already proved in #151.

PR #178, opened 2025-04-22
CI: Add EC2 runners and move CBMC there

mkannwischer:


mkannwischer:
Moving CBMC to more beefy EC2 runners does not result in significantly faster proof time. No need to do that. Maybe we go straight to adding benchmarking.

mkannwischer:
This works fine on my machine, but fails in CI. 
I rebased on top of #178 to check if running it on a bigger EC2 box solves the problem.

mkannwischer:
> Moving CBMC to more beefy EC2 runners does not result in significantly faster proof time. No need to do that. Maybe we go straight to adding benchmarking.

I found out why it is not faster - this was actually recompiling CBMC in CI. This is because somehow the Arm64 cache entry got evicted. I'm rebuilding it now in https://github.com/pq-code-package/mldsa-native/actions/runs/14659779817/job/41141438281. Once that is done, we can rebase this PR and it should be quite a bit faster than on standard Github runners. It's also required to make the Keccak proofs in #181 work.

.github/workflows/ci.yml
```diff
@@ -274,34 +274,76 @@ jobs:
...
-      - uses: ./.github/actions/cbmc
-        with:
-          mldsa_mode: 3
-          gh_token: ${{ secrets.GITHUB_TOKEN }}
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    uses: ./.github/workflows/ci_ec2_reusable.yml
+    with:
+      name: CBMC (MLKEM-512)
```
hanno-becker: Copy pasta

mkannwischer: Fixed.

.github/workflows/ci_ec2_reusable.yml
```diff
@@ -0,0 +1,213 @@
...
+          acvp: ${{ inputs.acvptest }}
+      - name: CBMC
+        if: ${{ inputs.cbmc && (success() || failure()) }}
+        uses: ./.github/actions/cbmc
+        with:
+          nix-shell: ci-cbmc
+          nix-verbose: ${{ inputs.verbose }}
+          mldsa_mode: ${{ inputs.cbmc_mldsa_mode }}
+          gh_token: ${{ secrets.AWS_GITHUB_TOKEN }}
+      - name: SLOTHY
```
hanno-becker: Is this copy-pasta or does it do something?

hanno-becker (changes requested):
LGTM apart from a stray section on SLOTHY, which should be removed

hanno-becker:
@mkannwischer Should this still be marked `DO-NOT-MERGE`?

.github/workflows/ci_ec2_reusable.yml
mkannwischer: I commented it out, okay? Once we have SLOTHYed code, we should re-add it.

hanno-becker: Ok

hanno-becker (approved):

Issue #106, opened 2025-04-09
CBMC: Prove `poly_invntt_tomont`


mkannwischer:
Depends on #103

PR #179, opened 2025-04-22
CMBC: Refactor, add contract and proof for invntt_tomont

jakemas:
https://github.com/pq-code-package/mldsa-native/issues/103

PR #180, opened 2025-04-23
CBMC: Add contract and proof for poly_invntt_tomont

jakemas:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/106

### DIFF
--- a/mldsa/poly.h
+++ b/mldsa/poly.h
@@ -136,7 +136,13 @@ __contract__(
  *
  * Arguments:   - poly *a: pointer to input/output polynomial
  **************************************************/
-void poly_invntt_tomont(poly *a);
+void poly_invntt_tomont(poly *a)
+__contract__(
+  requires(memory_no_alias(a, sizeof(poly)))
+  requires(array_abs_bound(a->coeffs, 0, MLDSA_N, MLDSA_Q))
+  assigns(memory_slice(a, sizeof(poly)))
+  ensures(array_abs_bound(a->coeffs, 0, MLDSA_N, MLDSA_Q))
+);
 
 #define poly_pointwise_montgomery MLD_NAMESPACE(poly_pointwise_montgomery)
 /*************************************************
--- a/proofs/cbmc/poly_invntt_tomont/Makefile
+++ b/proofs/cbmc/poly_invntt_tomont/Makefile
@@ -1,0 +1,54 @@
+# SPDX-License-Identifier: Apache-2.0
+
+include ../Makefile_params.common
+
+HARNESS_ENTRY = harness
+HARNESS_FILE = poly_invntt_tomont_harness
+
+# This should be a unique identifier for this proof, and will appear on the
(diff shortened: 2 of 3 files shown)


mkannwischer (changes requested):
Please rebase this on top of of #179 (which has the `invntt_tomont` contract that you are using here). 

Doing a `USE_FUNCTION_CONTRACTS=` for a function that has no contract is a very bad idea. I'm quite surprised that you managed to proof an output bound for this.
@rod-chapman do you think this is intended behaviour?

mkannwischer:
After looking into #179, I realized that there there is actually no proof for `invntt_tomont` yet (only `invntt_layer`). The `invntt_tomont` proof should be done first to avoid having missing proofs later (in mlkem-native we noticed way too late that we were missing some proofs.)

mkannwischer:
#183 got merged, so this one can be merged as well now. I rebased on top of main.

mldsa/poly.h
```diff
@@ -136,7 +136,13 @@ __contract__(
  *
  * Arguments:   - poly *a: pointer to input/output polynomial
  **************************************************/
-void poly_invntt_tomont(poly *a);
+void poly_invntt_tomont(poly *a)
+__contract__(
+  requires(memory_no_alias(a, sizeof(poly)))
+  requires(array_abs_bound(a->coeffs, 0, MLDSA_N, MLDSA_Q))
+  assigns(object_whole(a))
```
hanno-becker: This should be `memory_slice(a, sizeof(poly))`.

hanno-becker (changes requested):
Can you change `object_whole` to `memory_slice`? Otherwise, we'll get issues later on when the function is being called on a poly-array.

proofs/cbmc/poly_invntt_tomont/Makefile
```diff
@@ -0,0 +1,55 @@
...
+
+DEFINES +=
+INCLUDES +=
+
+REMOVE_FUNCTION_BODY +=
+UNWINDSET +=
+
+PROOF_SOURCES += $(PROOFDIR)/$(HARNESS_FILE).c
+PROJECT_SOURCES += $(SRCDIR)/mldsa/poly.c
+PROJECT_SOURCES += $(SRCDIR)/mldsa/ntt.c
```
hanno-becker: ntt.c is not needed I think

hanno-becker (changes requested):
Minor not regarding an unused input source in CBMC

jakemas:
> Minor not regarding an unused input source in CBMC

Thanks Hanno, fixed!

proofs/cbmc/poly_invntt_tomont/Makefile
jakemas: fixed in [d303353](https://github.com/pq-code-package/mldsa-native/pull/180/commits/d3033535a19f0b6b9a7b24fb8ab25af372b74952)

mkannwischer (approved):

PR #181, opened 2025-04-25
Add spec and proofs for Keccak dependent functions 

mkannwischer:
Continuation of https://github.com/pq-code-package/mldsa-native/pull/167. 
These proofs work fine on my machine, but fail in CI. Testing here if they work if we run them on bigger EC2 boxes.

### DIFF
--- a/mldsa/fips202/fips202.c
+++ b/mldsa/fips202/fips202.c
@@ -475,7 +475,6 @@ static unsigned int keccak_squeeze(uint8_t *out, size_t outlen, uint64_t s[25],
   return pos;
 }
 
-
 /*************************************************
  * Name:        keccak_absorb_once
  *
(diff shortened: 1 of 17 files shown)


mkannwischer:
This confirms that it does work with sufficiently much RAM.
We need to merge #178 first.

mldsa/fips202/fips202.c
```diff
@@ -403,7 +422,32 @@ static unsigned int keccak_absorb(uint64_t s[25], unsigned int pos,
...
+    {
+      s[i / 8] ^= (uint64_t)*in++ << 8 * (i % 8);
+    }
+    inlen -= r;
+    KeccakF1600_StatePermute(s);
+  }
+
   for (i = pos; i < pos + inlen; i++)
+  __loop__(
+          assigns(i, in, memory_slice(s, sizeof(uint64_t) * MLD_KECCAK_LANES))
```
hanno-becker: Nit: Spacing

mldsa/fips202/fips202.c
```diff
@@ -394,15 +394,26 @@ static void keccak_init(uint64_t s[25])
...
+  requires(r >= 0 && r < sizeof(uint64_t) * MLD_KECCAK_LANES)
+  requires(pos < r && pos >= 0)
+  requires(memory_no_alias(s, sizeof(uint64_t) * MLD_KECCAK_LANES))
+  requires(memory_no_alias(in, inlen))
+  assigns(memory_slice(s, sizeof(uint64_t) * MLD_KECCAK_LANES)))
 {
   unsigned int i;
 
-  while (pos + inlen >= r)
+  if (pos + inlen >= r)
```
hanno-becker: I don't think the removal of the `while` loop is necessary. You can use the same invariant as you do below, just add `invariant(pos < r)`

hanno-becker: @manastasova I took the liberty to adjust this in a follow-up commit.

mldsa/fips202/fips202.c
hanno-becker: @manastasova I took the liberty to adjust this and similar instances in a follow-up commit.

mkannwischer:
Thanks @hanno-becker for the changes - this PR looks good to me now!

hanno-becker (approved):
Thanks @manastasova!

For the future, please see the additional commit regarding style changes. Also, please keep the character count in commit messages < 74.

PR #191, opened 2025-05-01
CBMC: Add proof and contract for shake128_init _absorb _finalize

jakemas:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/148
- Resolves https://github.com/pq-code-package/mldsa-native/issues/149
- Resolves https://github.com/pq-code-package/mldsa-native/issues/150

### DIFF
--- a/mldsa/fips202/fips202.h
+++ b/mldsa/fips202/fips202.h
@@ -29,11 +29,28 @@ typedef struct
 extern const uint64_t KeccakF_RoundConstants[];
 
 #define shake128_init FIPS202_NAMESPACE(shake128_init)
-void shake128_init(keccak_state *state);
+void shake128_init(keccak_state *state)
+__contract__(
+  requires(memory_no_alias(state, sizeof(keccak_state)))
+  assigns(memory_slice(state, sizeof(keccak_state)))
+  ensures(state->pos == 0)
+);
 #define shake128_absorb FIPS202_NAMESPACE(shake128_absorb)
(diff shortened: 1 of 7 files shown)


mldsa/fips202/fips202.h
```diff
@@ -29,11 +29,29 @@ typedef struct
 extern const uint64_t KeccakF_RoundConstants[];
 
 #define shake128_init FIPS202_NAMESPACE(shake128_init)
-void shake128_init(keccak_state *state);
+void shake128_init(keccak_state *state)
+__contract__(
+  requires(memory_no_alias(state, sizeof(keccak_state)))
+  assigns(memory_slice(state, sizeof(keccak_state)))
+  ensures(forall(k, 0, MLD_KECCAK_LANES, state->s[k] == 0))
```
manastasova: We decided to remove this post-condition for shake256 https://github.com/pq-code-package/mldsa-native/blob/e669f24be50aaa2bfb47f61eed2847a90041701e/mldsa/fips202/fips202.h#L50
Actually, I would think that since `keccak_init` does not have such post-condition, the proof for `shake128_init` should fail.

jakemas: Thank you! Addressed in [e944611](https://github.com/pq-code-package/mldsa-native/pull/191/commits/e944611549e756be986f47d16eb1263999c599d5)

mkannwischer (approved):
LGTM - thanks @jakemas.
I'm unsure if we really want the pre- and post-conditions on `pos` - but we can change that later if needed.

PR #199, opened 2025-05-02
Proof of poly_challenge() including contracts

### LINKED ISSUE #113: CBMC: Prove `poly_challenge`
Depends on #145 #146 #147 #144.

rod-chapman:
Fixes #113 

Adds contracts and proof support files for poly_challenge()

### DIFF
--- a/mldsa/poly.c
+++ b/mldsa/poly.c
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 #include <stdint.h>
+#include <string.h>
 
 #include "ntt.h"
 #include "poly.h"
@@ -390,8 +391,9 @@ void poly_uniform_gamma1(poly *a, const uint8_t seed[MLDSA_CRHBYTES],
 
 void poly_challenge(poly *c, const uint8_t seed[MLDSA_CTILDEBYTES])
 {
-  unsigned int i, b, pos;
+  unsigned int i, j, pos;
   uint64_t signs;
+  uint64_t offset;
   uint8_t buf[SHAKE256_RATE];
   keccak_state state;
 
@@ -400,32 +402,59 @@ void poly_challenge(poly *c, const uint8_t seed[MLDSA_CTILDEBYTES])
   shake256_finalize(&state);
   shake256_squeezeblocks(buf, 1, &state);
 
+  /* Convert the first 8 bytes of buf[] into an unsigned 64-bit value.   */
+  /* Each bit of that dictates the sign of the resulting challenge value */
   signs = 0;
   for (i = 0; i < 8; ++i)
+  __loop__(
+    assigns(i, signs)
+    invariant(i <= 8)
+  )
   {
     signs |= (uint64_t)buf[i] << 8 * i;
   }
   pos = 8;
 
-  for (i = 0; i < MLDSA_N; ++i)
-  {
-    c->coeffs[i] = 0;
-  }
+  memset(c, 0, sizeof(poly));
+
   for (i = MLDSA_N - MLDSA_TAU; i < MLDSA_N; ++i)
+  __loop__(
+    assigns(i, j, object_whole(buf), state, pos, memory_slice(c, sizeof(poly)), signs)
+    invariant(i >= MLDSA_N - MLDSA_TAU)
+    invariant(i <= MLDSA_N)
+    invariant(pos >= 1)
+    invariant(pos <= SHAKE256_RATE)
+    invariant(array_bound(c->coeffs, 0, MLDSA_N, -1, 2))
+  )
   {
     do
+    __loop__(
+      assigns(j, object_whole(buf), state, pos)
+    )
     {
       if (pos >= SHAKE256_RATE)
       {
         shake256_squeezeblocks(buf, 1, &state);
         pos = 0;
       }
+      j = buf[pos++];
+    } while (j > i);
 
-      b = buf[pos++];
-    } while (b > i);
+    c->coeffs[i] = c->coeffs[j];
 
-    c->coeffs[i] = c->coeffs[b];
-    c->coeffs[b] = 1 - 2 * (signs & 1);
+    /* Reference: Compute coefficent value here in two steps to */
+    /* mixinf unsigned and signed arithmetic with implicit      */
+    /* conversions, and so that CBMC can keep track of ranges   */
+    /* to complete type-safety proof here.                      */
+
+    /* The least-significant bit of signs tells us if we want -1 or +1 */
+    offset = 2 * (signs & 1);
+
+    /* offset has value 0 or 2 here, so (1 - (int32_t) offset) has
+     * value -1 or +1 */
+    c->coeffs[j] = 1 - (int32_t)offset;
+
+    /* Move to the next bit of signs for next time */
     signs >>= 1;
   }
 }
--- a/mldsa/poly.h
+++ b/mldsa/poly.h
@@ -343,7 +343,14 @@ void poly_uniform_gamma1(poly *a, const uint8_t seed[MLDSA_CRHBYTES],
  *              - const uint8_t mu[]: byte array containing seed of length
  *                MLDSA_CTILDEBYTES
  **************************************************/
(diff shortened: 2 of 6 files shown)


mldsa/poly.c
```diff
@@ -400,32 +402,59 @@ void poly_challenge(poly *c, const uint8_t seed[MLDSA_CTILDEBYTES])
...
 
-  for (i = 0; i < MLDSA_N; ++i)
-  {
-    c->coeffs[i] = 0;
-  }
+  memset(c, 0, sizeof(poly));
+
   for (i = MLDSA_N - MLDSA_TAU; i < MLDSA_N; ++i)
+  __loop__(
+    assigns(i, j, object_whole(buf), state, pos, memory_slice(c, sizeof(poly)), signs)
```
manastasova, line 422: I was wondering when do we prefer `object_whole(buf)` vs `memory_slice(buf, SHAKE256_RATE)`?

mldsa/poly.c
```diff
@@ -400,32 +402,59 @@ void poly_challenge(poly *c, const uint8_t seed[MLDSA_CTILDEBYTES])
...
         pos = 0;
       }
+      j = buf[pos++];
+    }
+    while (j > i) /* clang-format mangles layout owing to the contract above */
+      ;
+
+    c->coeffs[i] = c->coeffs[j];
+
+    /* The least-significant bit of signs tells is if we want -1 or +1 */
```
manastasova:
```suggestion
    /* The least-significant bit of signs tells us if we want -1 or +1 */
```

mldsa/poly.c
```diff
@@ -400,32 +402,59 @@ void poly_challenge(poly *c, const uint8_t seed[MLDSA_CTILDEBYTES])
...
+    invariant(i <= MLDSA_N)
+    invariant(pos >= 1)
+    invariant(pos <= SHAKE256_RATE)
+    invariant(array_bound(c->coeffs, 0, MLDSA_N, -1, 2))
+  )
   {
     do
+    __loop__(
+      assigns(j, object_whole(buf), state, pos)
+      invariant(pos >= 1)
```
manastasova: Based on `pos = 0`, line 441, isn't this invariant going to evaluate to false when the `if` branch is entered?

mldsa/poly.c
```diff
@@ -400,32 +402,59 @@ void poly_challenge(poly *c, const uint8_t seed[MLDSA_CTILDEBYTES])
...
+    )
     {
       if (pos >= SHAKE256_RATE)
       {
         shake256_squeezeblocks(buf, 1, &state);
         pos = 0;
       }
+      j = buf[pos++];
+    }
+    while (j > i) /* clang-format mangles layout owing to the contract above */
```
manastasova:
```suggestion
    while (j > i); /* clang-format mangles layout owing to the contract above */
```

mldsa/poly.c
```diff
@@ -400,32 +402,59 @@ void poly_challenge(poly *c, const uint8_t seed[MLDSA_CTILDEBYTES])
...
+      ;
```
manastasova:
```suggestion
```

mldsa/poly.c
```diff
@@ -400,32 +402,59 @@ void poly_challenge(poly *c, const uint8_t seed[MLDSA_CTILDEBYTES])
...
+    invariant(pos <= SHAKE256_RATE)
+    invariant(array_bound(c->coeffs, 0, MLDSA_N, -1, 2))
+  )
   {
     do
+    __loop__(
+      assigns(j, object_whole(buf), state, pos)
+      invariant(pos >= 1)
+      invariant(pos <= SHAKE256_RATE)
+      invariant(array_bound(c->coeffs, 0, MLDSA_N, -1, 2))
```
manastasova: Can we remove this invariant since this loop does not change the value of `c`?

mldsa/poly.c
```diff
@@ -400,32 +402,59 @@ void poly_challenge(poly *c, const uint8_t seed[MLDSA_CTILDEBYTES])
...
       {
         shake256_squeezeblocks(buf, 1, &state);
         pos = 0;
       }
+      j = buf[pos++];
+    }
+    while (j > i) /* clang-format mangles layout owing to the contract above */
+      ;
+
+    c->coeffs[i] = c->coeffs[j];
```
manastasova, line 443: After the while loop aren't the values of `i` and `j` the same?

mldsa/poly.h
```diff
@@ -5,6 +5,7 @@
 #ifndef MLD_POLY_H
 #define MLD_POLY_H
 
+#include <stdbool.h>
```
manastasova: Do we need booleans here?

hanno-becker: I don't think so

mldsa/poly.c
hanno-becker: In function contracts, we prefer `memory_slice(...)`. Otherwise, if you call the contract on a pointer to a subobject of a larger structure, the whole structure will be tainted, which is almost never what you want.

For local structures, it does not matter, and I think `object_whole` should be used -- including this case.

mldsa/poly.c
jakemas: Nit: "tells us"

mldsa/poly.c
```diff
@@ -400,32 +402,59 @@ void poly_challenge(poly *c, const uint8_t seed[MLDSA_CTILDEBYTES])
...
 
-      b = buf[pos++];
-    } while (b > i);
+    /* offset has value 0 or 2 here, so (1 - (int32_t) offset) has
+     * value -1 or +1 */
+    c->coeffs[j] = 1 - (int32_t)offset;
 
-    c->coeffs[i] = c->coeffs[b];
-    c->coeffs[b] = 1 - 2 * (signs & 1);
+    /* Move to the next bit of signs for next time */
```
jakemas, line 457: Curious about the deviation from the reference code here

hanno-becker: Agree. We should strive to deviate only if we need to, and if so, leave a comment in the commit message and a `/* Reference: ... */` comment.

mldsa/poly.c
manastasova: I see now, they will not! Sorry for the question!

mldsa/poly.c
rod-chapman: The original code mixed unsigned and signed arithmetic and relied on implicit conversions. As a result, CBMC was not able to reason about the range of the resulting expression, leading to a proof failure. As (re-written), it proves fine, it's probably clearer to a human reader what's going on.

mldsa/poly.c
mkannwischer: No, `pos` is incremented in line 443.

mldsa/poly.c
mkannwischer: You can stop `clang-format` from mangling your code here if you change `.clang-format` 
from 
` - __loop__(x)={}`
to 
` - __loop__(x)={} do`

A little hacky, but it does not seem to have any bad side-effects in mldsa-native and mlkem-native.
While you touching `.clang-format` already maybe fix the typo in artific(i)ally

rod-chapman: Done as suggested.

mldsa/poly.h
rod-chapman: Ah... my mistake. I initally has "ensures(true)" as a placeholder precondition, so needed stdbool.h, but this should be removed now.

mldsa/poly.c
rod-chapman: Fixed.

mldsa/poly.c
rod-chapman: Comment added.

mldsa/poly.c
rod-chapman: Fixed.

mldsa/poly.c
rod-chapman: Fixed by updated .clang-format as suggested below.

mldsa/poly.c
rod-chapman: Rather surprisingly, you're right. The "assigns" clause is required but the explicit invariants are not. Removed.

PR #200, opened 2025-05-02
CBMC: Add proof and contract for mldsa_shake128_stream_init

jakemas:
Resolves https://github.com/pq-code-package/mldsa-native/issues/141

### DIFF
--- a/mldsa/symmetric-shake.c
+++ b/mldsa/symmetric-shake.c
@@ -12,7 +12,8 @@ void mldsa_shake128_stream_init(keccak_state *state,
                                 uint16_t nonce)
 {
   uint8_t t[2];
-  t[0] = nonce;
+
+  t[0] = nonce & 0xFF;
   t[1] = nonce >> 8;
 
   shake128_init(state);
--- a/mldsa/symmetric.h
+++ b/mldsa/symmetric.h
@@ -6,6 +6,8 @@
 #define MLD_SYMMETRIC_H
 
 #include <stdint.h>
+#include "cbmc.h"
+#include "common.h"
 
 #include "fips202/fips202.h"
 
@@ -15,7 +17,12 @@ typedef keccak_state stream256_state;
 #define mldsa_shake128_stream_init MLD_NAMESPACE(mldsa_shake128_stream_init)
 void mldsa_shake128_stream_init(keccak_state *state,
                                 const uint8_t seed[MLDSA_SEEDBYTES],
-                                uint16_t nonce);
+                                uint16_t nonce)
+__contract__(
+  requires(memory_no_alias(state, sizeof(keccak_state)))
+  requires(memory_no_alias(seed, MLDSA_SEEDBYTES))
+  assigns(memory_slice(state, sizeof(keccak_state)))
+);
 
 #define mldsa_shake256_stream_init MLD_NAMESPACE(mldsa_shake256_stream_init)
 void mldsa_shake256_stream_init(keccak_state *state,
--- a/proofs/cbmc/mldsa_shake128_stream_init/mldsa_shake128_stream_init_harness.c
+++ b/proofs/cbmc/mldsa_shake128_stream_init/mldsa_shake128_stream_init_harness.c
@@ -1,0 +1,14 @@
+// Copyright (c) 2025 The mldsa-native project authors
+// SPDX-License-Identifier: Apache-2.0
+
+#include "params.h"
+#include "symmetric.h"
+
+void harness(void)
+{
+  keccak_state *s;
+  const uint8_t *seed;
+  uint16_t nonce;
+
+  mldsa_shake128_stream_init(s, seed, nonce);
+}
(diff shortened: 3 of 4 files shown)


mldsa/symmetric.h
```diff
@@ -15,7 +17,13 @@ typedef keccak_state stream256_state;
 #define mldsa_shake128_stream_init MLD_NAMESPACE(mldsa_shake128_stream_init)
 void mldsa_shake128_stream_init(keccak_state *state,
                                 const uint8_t seed[MLDSA_SEEDBYTES],
-                                uint16_t nonce);
+                                uint16_t nonce)
+__contract__(
+  requires(memory_no_alias(state, sizeof(keccak_state)))
+  requires(memory_no_alias(seed, MLDSA_SEEDBYTES))
+  requires(memory_no_alias(nonce, sizeof(nonce)))
```
hanno-becker: `nonce` is not a pointer, so this should be removed

jakemas:
> One unnecessary use of `memory_no_alias` in the preconditions.

When I remove this precondition I get an issue with CBMC overflow detection:

```
VERIFICATION FAILED

[MLD_65_ref_mldsa_shake128_stream_init.overflow.1] line 15 arithmetic overflow on unsigned to unsigned type conversion in (uint8_t)nonce: FAILURE
```

hanno-becker:
@jakemas Yes, CBMC does not like implicit truncation. You need to explicitly mask: `t[0] = nonce & 0xFF` etc. 

Have a look at https://github.com/pq-code-package/mlkem-native/blob/main/mlkem/compress.c#L101 where the same thing happened during compression in mlkem-native.

jakemas:
> @jakemas Yes, CBMC does not like implicit truncation. You need to explicitly mask: `t[0] = nonce & 0xFF`
> 
> Have a look at https://github.com/pq-code-package/mlkem-native/blob/main/mlkem/compress.c#L101 where the same thing happened during compression in mlkem-native.

Got it! Thanks!

mldsa/symmetric-shake.c
```diff
@@ -12,8 +12,9 @@ void mldsa_shake128_stream_init(keccak_state *state,
                                 uint16_t nonce)
 {
   uint8_t t[2];
-  t[0] = nonce;
-  t[1] = nonce >> 8;
+
+  t[0] = nonce & 0xFF;       
+  t[1] = (nonce >> 8) & 0xFF;
```
hanno-becker: The second truncation is not needed since the value is already within the range of `uint8_t`.

proofs/cbmc/mldsa_shake128_stream_init/mldsa_shake128_stream_init_harness.c
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2025 The mldsa-native project authors
+// SPDX-License-Identifier: Apache-2.0
+
+#include "params.h"
+#include "symmetric.h"
+
+void harness(void)
+{
+  keccak_state *s;
+  const uint8_t seed[MLDSA_SEEDBYTES];
```
hanno-becker: Blocker: This must be `const uint8_t *seed`. Otherwise, the harness always calls `mldsa_shake128_stream_init` with a valid `seed` pointer. Instead, it should call the function-under-proof with a priori unconstrained arguments, which are then _only_ constrained by the function preconditions.

To check, try removing the validity precondition for `seed` -- I suspect the proof will still go through.

We should double-check whether this is done elsewhere, too.

hanno-becker (changes requested):
Function arguments `T foo[N]` must be declared as `T* foo` in the harness, _not_ as `T foo[N]`. Otherwise, the function will only be considered for valid `foo`, _even_ if it does not have a precondition for `foo`.

@jakemas Can you please check existing harnesses to ensure we are not doing this elsewhere? We should never allocate an array in a harness.

proofs/cbmc/mldsa_shake128_stream_init/mldsa_shake128_stream_init_harness.c
jakemas: ok fixed!

hanno-becker (approved):
Thanks @jakemas!

mldsa/symmetric-shake.c
```diff
@@ -12,7 +12,8 @@ void mldsa_shake128_stream_init(keccak_state *state,
                                 uint16_t nonce)
 {
   uint8_t t[2];
-  t[0] = nonce;
+
+  t[0] = nonce & 0xFF;
```
manastasova, line 16: Wouldn't be a good practice to add explicit casts? Or it is not needed here?

hanno-becker: You don't need explicit cast to make CBMC happy, and IIRC we tend to not make them explicit. However, as mentioned above, you still need to make the value _truncation_ explicit.

PR #202, opened 2025-05-03
[DRAFT] CBMC: Add proof and contract for poly_uniform

jakemas:


hanno-becker (commented):
As mentioned offline: We've had non-termination issues around rejection sampling in mlkem-native, too. This can be resolved by making the `ctr` 'watermark' an explicit parameter to `rej_uniform`, thereby removing the pointer shift in the later calls. This (and the adjustment of spec+proof for `rej_uniform`) should be done in a separate PR first before continuing here.

jakemas:
> As mentioned offline: We've had proof non-termination issues around rejection sampling in mlkem-native, too. This can be resolved by making the `ctr` 'watermark' an explicit parameter to `rej_uniform`, thereby removing the pointer shift in the later calls. This (and the adjustment of spec+proof for `rej_uniform`) should be done in a separate PR first before continuing here.

https://github.com/pq-code-package/mldsa-native/pull/218

hanno-becker (changes requested):
@jakemas @mkannwischer I don't think the pre-existing code for `poly_uniform` makes sense. It should be fixed first.

1. Ignoring the actual value of `off == buflen % 3`, I think the inner loop would functionally do the wrong thing if it were ever called more than once: After the initial squeeze, we move the bytes `buflen - off, .. buflen - 1` forward and fill the rest with a fresh squeeze of `buflen` bytes. In particular, we will fill the excess bytes of index `>= buflen`. But if that's not enough, we will again move `buflen - off, .. buflen - 1` to the front, and squeeze again, which will _overwrite the excess bytes_ even though they were never used.

  It's quite surprising/shocking that this simple bug was not noted in the reference implementation before.

EDIT: My analysis above is wrong as it ignores that `buflen` is adjusted in the loop. The code is merely dead (for the ML-DSA parameter sets), but not buggy.

2. Luckily for us, the whole code seems pointless since `buflen % 3 == 0` (`STREAM128_BLOCKBYTES == 168`).

jakemas:
> @jakemas @mkannwischer I don't think the pre-existing code for `poly_uniform` makes sense. It should be fixed first.
> 
> 1. Ignoring the actual value of `off == buflen % 3`, I think the inner loop would functionally do the wrong thing if it were ever called more than once: After the initial squeeze, we move the bytes `buflen - off, .. buflen - 1` forward and fill the rest with a fresh squeeze of `buflen` bytes. In particular, we will fill the excess bytes of index `>= buflen`. But if that's not enough, we will again move `buflen - off, .. buflen - 1` to the front, and squeeze again, which will _overwrite the excess bytes_ even though they were never used.
> 
> It's quite surprising/shocking that this simple bug was not noted in the reference implementation before.
> 
> 2. Luckily for us, the whole code seems pointless since `buflen % 3 == 0` (`STREAM128_BLOCKBYTES == 168`).

Yup, it seems Kyber addresses this in the reference implementation here: https://github.com/pq-crystals/kyber/commit/ce492cd3d56bfca61bd941b6a73da3764e799aae in the similar function `gen_matrix`.

hanno-becker:
@jakemas Yes, though it was never acknowledged that even _if_ `off != 0`, the logic does not work.

jakemas:
See https://github.com/pq-code-package/mldsa-native/pull/219 for proof on refactored functions

hanno-becker:
@jakemas Thanks. Note that a new PR would not have been necessary: The present one could have been rebased.

PR #203, opened 2025-05-03
Add Neon NTT

mkannwischer:
- Resolves #197 

I have not run it through SLOTHY yet. I'll do that in a follow-up PR.

PR #204, opened 2025-05-05
Add Neon iNTT

### LINKED ISSUE #198: Neon backend: Add iNTT

mkannwischer:
- Resolves #198
- Depends on #203 

This adds https://github.com/slothy-optimizer/slothy/blob/main/examples/naive/aarch64/intt_dilithium_1234_5678.s. 
I cleaned it up in the similar way as the NTT.
So far the applied changes are as separate commits to ease review. Once approved, I propose to squash those into one commit.


mldsa/native/aarch64/src/intt.S
```diff
@@ -0,0 +1,409 @@
...
+        str q_data13, [in,  #(13*(512/8))]
+        str q_data14, [in,  #(14*(512/8))]
+        str q_data15, [in,  #(15*(512/8))]
+
+        // Scale half the coeffs by 1/n; for the other half, the scaling has
+        // been merged into the multiplication with the twiddle factor on the
+        // last layer.
+        mul_ninv data0, data1, data2, data3, data4, data5, data6, data7, data0, data1, data2, data3, data4, data5, data6, data7
+
+        canonical_reduce data0, modulus_half, neg_modulus_half, t2, t3
```
hanno-becker: I don't think we need a canonical reduction here, do we? The caller conducts further processing on the output of the invNTT which necessitates a canonical reduction at a later point.

mldsa/native/aarch64/src/intt.S
```diff
@@ -0,0 +1,409 @@
...
+
+        load_next_roots_6  root1, r1234_ptr
+        load_next_roots_56 root0, r1234_ptr
+
+        gs_butterfly data0, data1, root0, 0, 1
+        gs_butterfly data2, data3, root0, 2, 3
+        gs_butterfly data0, data2, root1, 0, 1
+        gs_butterfly data1, data3, root1, 0, 1
+
+        barrett_reduce_single data0
```
hanno-becker: Is this needed? We currently assume that the inputs to the iNTT are bound by Q, in which case no 32-bit overflow can happen.

hanno-becker (changes requested):
There seem to be inconsistencies between the C invNTT and the AArch64 invNTT in terms of their assumed input bounds (the C invNTT does not [need to] conduct any interim reductions)) and output bounds (the AArch64 NTT guarantees canonical representatives, the C invNTT doesn't). Those need to be ironed out.

mldsa/native/aarch64/src/intt.S
mkannwischer: Agreed. That was an artifact from the paper so that we could compare to a previous implementation.

mldsa/native/aarch64/src/intt.S
mkannwischer: I am not quite sure where this originates from - it's taken directly from the SLOTHY paper. Possibly we wrote that to support larger inputs, but I cannot recall right now.
I agree with you that it should not be needed. I have removed it.

hanno-becker (approved):
Thanks @mkannwischer!

PR #213, opened 2025-05-06
Optimize proof for `keccak_squeezeblocks`

manastasova:
This PR introduces the function `keccakf1600_extract_bytes` which optimizes the CBMC proof latency for `keccak_squeezeblocks` to 48s (previously ~150s).

Notes:
- Introduce `keccakf1600_extract_bytes` function as it is defined in mlkem-native - https://github.com/pq-code-package/mlkem-native/blob/ae2afe58380797f6b6789060b3ff8be2b2917fb5/mlkem/fips202/keccakf1600.c#L38 
- The main speedup was observed after adding support for both little- and big-endian.

### DIFF
--- a/mldsa/fips202/fips202.c
+++ b/mldsa/fips202/fips202.c
@@ -563,6 +563,46 @@ __contract__(
 }
 
 /*************************************************
+ * Name:        keccakf1600_extract_bytes
+ *
+ * Description: Extracts a specified number of bytes from the state
+ *              buffer starting at offset and stores them in the provided
+ *              data buffer
+ *
+ * Arguments:   - uint64_t *state: pointer to the state array
+ *              - unsigned char *data: pointer to the output byte array
+ *(allocated)
+ *              - unsigned offset: starting byte position in the state
+ *              - unsigned length: number of bytes to extract
+ **************************************************/
+static void keccakf1600_extract_bytes(uint64_t *state, unsigned char *data,
+                                      unsigned offset, unsigned length)
+__contract__(
+  requires(0 <= offset && offset <= MLD_KECCAK_LANES * sizeof(uint64_t) &&
+      0 <= length && length <= MLD_KECCAK_LANES * sizeof(uint64_t) - offset)
+  requires(memory_no_alias(state, sizeof(uint64_t) * MLD_KECCAK_LANES))
+  requires(memory_no_alias(data, length))
+  assigns(memory_slice(data, length)))
+{
+  unsigned i;
+#if defined(MLD_SYS_LITTLE_ENDIAN)
+  uint8_t *state_ptr = (uint8_t *)state + offset;
+  for (i = 0; i < length; i++)
+  __loop__(invariant(i <= length))
+  {
+    data[i] = state_ptr[i];
+  }
+#else  /* MLD_SYS_LITTLE_ENDIAN */
+  /* Portable version */
+  for (i = 0; i < length; i++)
+  __loop__(invariant(i <= length))
+  {
+    data[i] = (state[(offset + i) >> 3] >> (8 * ((offset + i) & 0x07))) & 0xFF;
+  }
+#endif /* !MLD_SYS_LITTLE_ENDIAN */
+}
+
+/*************************************************
  * Name:        keccak_squeezeblocks
  *
  * Description: Squeeze step of Keccak. Squeezes full blocks of r bytes each.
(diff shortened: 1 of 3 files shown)


mkannwischer:
Thanks @manastasova  for looking into this. 
What stops us from using [keccak_squeezeblocks from mlkem-native](https://github.com/pq-code-package/mlkem-native/blob/ae2afe58380797f6b6789060b3ff8be2b2917fb5/mlkem/fips202/fips202.c#L116) including it's proof?

manastasova:
Some of the functions used in mlkem-native are defined differently https://github.com/pq-code-package/mlkem-native/blob/ae2afe58380797f6b6789060b3ff8be2b2917fb5/mlkem/fips202/keccakf1600.c#L38. We could either adapt the functions or the proofs.

rod-chapman:
With this change, I still see an SMT2 file of c. 150MB and the "make result" run on my laptop takes 100s. I would like to see this re-factored so the inner loop of keccak_squeezeblocks() becomes a stand-alone static functions, as in mlkem-native.

manastasova:
Should we better bring all fips202 code from mlkem-native?

rod-chapman:
I might say "not yet"... the mlkem-native code has all the "x4" variants of Keccak - are these required for MLDSA, or does a different strategy offer better performance for MLDSA? I defer to Hanno and Matthias to comment on that matter.

mldsa/fips202/fips202.c
```diff
@@ -562,6 +562,45 @@ __contract__(
...
+ * Name:        keccak_squeezeblocks
+ *
+ * Description:
+ *
+ * Arguments:   - uint8_t *out: pointer to output blocks
+ *              - size_t nblocks: number of blocks to be squeezed (written to
+ *out)
+ *              - uint64_t *s: pointer to input/output Keccak state
+ *              - unsigned int r: rate in bytes (e.g., 168 for SHAKE128)
+ **************************************************/
```
jakemas: This documentation needs updating

mldsa/fips202/fips202.c
```diff
@@ -562,6 +562,45 @@ __contract__(
...
+                                      unsigned offset, unsigned length)
+__contract__(
+requires(0 <= offset && offset <= MLD_KECCAK_LANES * sizeof(uint64_t) &&
+    0 <= length && length <= MLD_KECCAK_LANES * sizeof(uint64_t) - offset)
+requires(memory_no_alias(state, sizeof(uint64_t) * MLD_KECCAK_LANES))
+requires(memory_no_alias(data, length))
+assigns(memory_slice(data, length)))
+{
+  unsigned i;
+#if defined(MLK_SYS_LITTLE_ENDIAN)
```
jakemas: `MLK_SYS_LITTLE_ENDIAN` I'm guessing from the prefix that this is an mlkem-native definition that doesn't exist in mldsa-native.

hanno-becker: It's `MLD_SYS_LITTLE_ENDIAN` in mldsa-native

mldsa/fips202/fips202.c
```diff
@@ -586,23 +625,16 @@ __contract__(
...
 {
-  unsigned int i;
-
   while (nblocks > 0)
   __loop__(
-    assigns(i, out, nblocks, memory_slice(s, sizeof(uint64_t) * MLD_KECCAK_LANES), memory_slice(out, nblocks * r))
+    assigns(out, nblocks, memory_slice(s, sizeof(uint64_t) * MLD_KECCAK_LANES),
+      memory_slice(out, nblocks * r))
     invariant(nblocks <= loop_entry(nblocks))
-    invariant(out == loop_entry(out) + r * (loop_entry(nblocks) - nblocks) && nblocks >= 0)
```
jakemas, line 595: Out of curiosity, why was this part of the invariant removed? `&& nblocks >= 0`

mldsa/fips202/fips202.c
manastasova: Sorry, I got lost between multiple github branches! I just updated it.

mldsa/fips202/fips202.c
manastasova: Sorry, I got lost between multiple github branches! I just updated it.

mldsa/fips202/fips202.c
manastasova: It could be skipped. It is not used in the mlkem-native proof
https://github.com/pq-code-package/mlkem-native/blob/ae2afe58380797f6b6789060b3ff8be2b2917fb5/mlkem/fips202/fips202.c#L131
so I assume we prefer to omit it here as well.

mldsa/fips202/fips202.c
```diff
@@ -562,6 +562,46 @@ __contract__(
...
+ *              - unsigned length: number of bytes to extract
+ **************************************************/
+static void keccakf1600_extract_bytes(uint64_t *state, unsigned char *data,
+                                      unsigned offset, unsigned length)
+__contract__(
+requires(0 <= offset && offset <= MLD_KECCAK_LANES * sizeof(uint64_t) &&
+    0 <= length && length <= MLD_KECCAK_LANES * sizeof(uint64_t) - offset)
+requires(memory_no_alias(state, sizeof(uint64_t) * MLD_KECCAK_LANES))
+requires(memory_no_alias(data, length))
+assigns(memory_slice(data, length)))
```
jakemas: Small nit: missing indentation!

manastasova: Thank you!

rod-chapman:
Can you also call keccakf1600_extract_bytes() from keccak_squeeze() as well? I think we also need proof files for keccak_squeeze() too.

rod-chapman:
Please rebase this branch against main before merge.

manastasova:
I changed locally the `keccak_squeeze` function to use `keccakf1600_extract_bytes` and it seems to work. However, I am not sure that there are enough test cases in the scope of mldsa to check the complete correctness (the value of position should be tracked among the `keccak_squeeze`, outlen + pos will need different branches when it is larger/smaller than the blocks size, etc.).
Do you think it will be better if we add this change in a different PR?

rod-chapman:
OK... let's reconsider it in a different PR.

manastasova:
Perfect! I just rebased the branch.

rod-chapman (approved):
Looks good here.

hanno-becker:
@rod-chapman This now requires a separate big-endian proof ... https://github.com/pq-code-package/mldsa-native/issues/226

PR #214, opened 2025-05-06
Debug: Add bounds assertions macros

hanno-becker:
Following mlkem-native, this commit adds debug.h and debug.c,
offering assertions for runtime checking of array bounds.
Those assertions provide the main source of confidence in the
validity of our bounds assumptions prior to completion of all
CBMC proofs up to the toplevel API.

All implementations of functions in poly.c have bounds assertions
for inputs and outputs added that match the CBMC specifications.

As for mlkem-native, if CBMC is enabled, the bounds assertions
are interpreted as CBMC proof obligations.

To make sure this is exercised in CI, another test workflow from mlkem-native
is ported to mldsa-native's CI.


mkannwischer (approved):
Thanks @hanno-becker. 
Where else should we add those assertions? polvec.c/reduce.c? Maybe we create issues for those?

PR #218, opened 2025-05-06
CBMC: Refactor rej_uniform and poly_uniform functions

jakemas:
Working on addressing Hanno's [comment](https://github.com/pq-code-package/mldsa-native/pull/202#pullrequestreview-2818915323) to refactor `rej_uniform` as was done in [mlkem-native](https://github.com/pq-code-package/mlkem-native/blob/main/mlkem/sampling.c) to facilitate the proof for `poly_uniform`.

Alligning the implementation of `poly_uniform` to be up to date with changes to the reference implementation of kyber (https://github.com/pq-crystals/kyber/commit/ce492cd3d56bfca61bd941b6a73da3764e799aae).

### DIFF
--- a/mldsa/poly.c
+++ b/mldsa/poly.c
@@ -259,13 +259,22 @@ int poly_chknorm(const poly *a, int32_t B)
  *              performing rejection sampling on array of random bytes.
  *
  * Arguments:   - int32_t *a: pointer to output array (allocated)
- *              - unsigned int len: number of coefficients to be sampled
- *              - const uint8_t *buf: array of random bytes
- *              - unsigned int buflen: length of array of random bytes
+ *              - unsigned int target:  requested number of coefficients to
+ *sample
+ *              - unsigned int offset:  number of coefficients already sampled
+ *              - const uint8_t *buf: array of random bytes to sample from
+ *              - unsigned int buflen: length of array of random bytes (must be
+ *                multiple of 3)
  *
  * Returns number of sampled coefficients. Can be smaller than len if not enough
  * random bytes were given.
  **************************************************/
+
+/* Reference: `rej_uniform()` in the reference implementation [@REF].
+ *            - Our signature differs from the reference implementation
+ *              in that it adds the offset and always expects the base of the
+ *              target buffer. This avoids shifting the buffer base in the
+ *              caller, which appears tricky to reason about. */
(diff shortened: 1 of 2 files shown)


hanno-becker (changes requested):
The code looks good, but can you write detailed commit messages here? The change to `rej_uniform` should be explained by referring to observed CBMC slowdown with the original signature; the change to `poly_uniform` should be explained pointing out that the code is dead, and perhaps even pointing out that it would be buggy if it weren't dead.

hanno-becker (changes requested):
Can you note in the commit message for the `rej_uniform` refactoring that the change was made to enable the CBMC proof?

Also, please add a `Reference: ...` note similar to what's in mlkem-native.

Also, I apologize for my earlier claim that the dead code was buggy: I don't think it is, owing to the adjustment of `buflen` within the loop. I think it is merely dead and should be removed, but it looks correct even if `off != 0` in other parameter sets. So the respective paragraph in your commit message can just go.

Finally, while this isn't a strict rule (I'm usually breaking somewhere between 70-80 characters), please strive to keep each line in the commit message body to 72 characters.

jakemas:
> Can you note in the commit message for the `rej_uniform` refactoring that the change was made to enable the CBMC proof?
> 
> Also, please add a `Reference: ...` note similar to what's in mlkem-native.
> 
> Also, I apologize for my earlier claim that the dead code was buggy: I don't think it is, owing to the adjustment of `buflen` within the loop. I think it is merely dead and should be removed, but it looks correct even if `off != 0` in other parameter sets. So the respective paragraph in your commit message can just go.
> 
> Finally, while this isn't a strict rule (I'm usually breaking somewhere between 70-80 characters), please strive to keep each line in the commit message body to 72 characters.

Okay:
- added note to the commit message to `rej_uniform` to state refactoring change was made to enable CBMC proof
- added Reference callouts in the code to state deviations from reference (will promote this change upstream) for both `rej_uniform` and `poly_uniform`
- edited commit messages to remove comments on functional correctness
- kept line character in commit messages <80 (didn't realise it was transparent here)

hanno-becker (approved):
LGTM, thank you @jakemas for adjusting the commit messages, this will be helpful for us and other readers in the future!

PR #221, opened 2025-05-07
Run Neon NTT+iNTT through SLOTHY

### LINKED ISSUE #206: Run Neon NTT/iNTT through SLOTHY
NTT/iNTT added in #203  and #204 are so far clean assembly. We should run them through [SLOTHY](https://slothy-optimizer.github.io/slothy/) similar as in https://github.com/pq-code-package/mlkem-native/blob/main/dev/aarch64_opt/src/Makefile

mkannwischer:
- Resolves #206 

This adds a Makefile that runs the Neon NTT through SLOTHY. To accomodate this the clean assembly is moved to dev/aarch64_clean/, while the mldsa/native/aarch64 contains the optimized assembly.

The main difference to mlkem-native is that we need set an explicit timeout as optimizing the second loop doesn't result reasonable performance, but a good solution is found within one minute on my Apple M4. I set the timeout to 2 minutes with the hope that it works on most platforms. We have have to increase that later.

For now the clean backend is not tested in CI - that's left for a follow-up PR. SLOTHY is also not run in CI, yet.
We probably want to put the assembly simplification scripts in place so we can follow the same structure as in mlkem-native.


mldsa/native/aarch64/src/Makefile
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) The mldsa-native project authors
+# SPDX-License-Identifier: Apache-2.0
```
hanno-becker: Apache-2.0 OR ISC OR MIT?

mkannwischer: Yes, good catch. Fixed.

hanno-becker (commented):
I have so far not seen SLOTHY produce _any_ solution for layer45678 in the forward NTT, unfortunately. This is on my local Apple M1.

mkannwischer:
I'm suspecting the ortools version...

PR #220, opened 2025-05-07
Add spec and proofs for all fips202 functions

manastasova:
This PR is a continuation of https://github.com/pq-code-package/mldsa-native/pull/187 and adds specs and proofs for the following functions: 

- keccak_squeeze
- shake128_squeeze
- shake256_squeeze
- shake128
- shake256

PR #224, opened 2025-05-07
CBMC: Add proof and contract for poly_uniform_eta

jakemas:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/111

### DIFF
--- a/mldsa/poly.c
+++ b/mldsa/poly.c
@@ -447,6 +447,11 @@ void poly_uniform_eta(poly *a, const uint8_t seed[MLDSA_CRHBYTES],
   ctr = rej_eta(a->coeffs, MLDSA_N, 0, buf, buflen);
   buflen = STREAM256_BLOCKBYTES;
   while (ctr < MLDSA_N)
+  __loop__(
+    assigns(ctr, state, memory_slice(a, sizeof(poly)), object_whole(buf))
+    invariant(ctr <= MLDSA_N)
+    invariant((state).pos <= SHAKE256_RATE)
+    invariant(array_abs_bound(a->coeffs, 0, ctr, MLDSA_ETA + 1)))
   {
     stream256_squeezeblocks(buf, 1, &state);
     ctr = rej_eta(a->coeffs, MLDSA_N, ctr, buf, buflen);
--- a/proofs/cbmc/poly_uniform_eta/Makefile
+++ b/proofs/cbmc/poly_uniform_eta/Makefile
@@ -1,0 +1,54 @@
+# SPDX-License-Identifier: Apache-2.0
+
+include ../Makefile_params.common
+
+HARNESS_ENTRY = harness
+HARNESS_FILE = poly_uniform_eta_harness
+
+# This should be a unique identifier for this proof, and will appear on the
+# Litani dashboard. It can be human-readable and contain spaces if you wish.
+PROOF_UID = poly_uniform_eta
+
+DEFINES +=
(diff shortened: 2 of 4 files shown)


proofs/cbmc/poly_uniform_eta/Makefile
```diff
@@ -0,0 +1,54 @@
...
+PROOF_UID = poly_uniform_eta
+
+DEFINES +=
+INCLUDES +=
+
+REMOVE_FUNCTION_BODY +=
+UNWINDSET +=
+
+PROOF_SOURCES += $(PROOFDIR)/$(HARNESS_FILE).c
+PROJECT_SOURCES += $(SRCDIR)/mldsa/poly.c $(SRCDIR)/mldsa/symmetric-shake.c $(SRCDIR)/mldsa/fips202/fips202.c
```
mkannwischer:
```suggestion
PROJECT_SOURCES += $(SRCDIR)/mldsa/poly.c
```

CBMC proofs shouldn't need the FIPS sources. They should be called by contract, not by inlining.

mkannwischer (changes requested):
Thanks @jakemas. Looks mostly good. 
The CBMC proof shouldn't require the FIPS sources as only the contracts are needed and those are included through the headers.

proofs/cbmc/poly_uniform_eta/Makefile
jakemas: Fixed!

jakemas:
Something merged within https://github.com/pq-code-package/mldsa-native/pull/220 seems to have broken these proofs? The function uses `shake256_squeezeblocks` so perhaps this is the cause?

jakemas:
> Thanks @jakemas. Looks mostly good. The CBMC proof shouldn't require the FIPS sources as only the contracts are needed and those are included through the headers.

Thanks, fixed!

mldsa/poly.c
```diff
@@ -447,6 +447,11 @@ void poly_uniform_eta(poly *a, const uint8_t seed[MLDSA_CRHBYTES],
   ctr = rej_eta(a->coeffs, MLDSA_N, 0, buf, buflen);
   buflen = STREAM256_BLOCKBYTES;
   while (ctr < MLDSA_N)
+  __loop__(
+    assigns(ctr, state, memory_slice(a, sizeof(poly)), object_whole(buf))
+    invariant(ctr <= MLDSA_N)
+    invariant((&state)->pos <= SHAKE256_RATE)
```
hanno-becker: Nit: `state.pos` instead of `(&state)->pos`

hanno-becker (changes requested):
LGTM except for a small nit regarding `(&state)->foo` instead of `state.foo`

jakemas:
> LGTM except for a small nit regarding `(&state)->foo` instead of `state.foo`

Thank you, fixed!

mldsa/poly.c
jakemas: hmm, I get `/mldsa-native/mldsa/poly.c:450:1: error: member operator requires structure type on left hand side but got 'stream256_state *'` when I try this.

jakemas: Fixed [fdd5ca9](https://github.com/pq-code-package/mldsa-native/pull/224/commits/fdd5ca90f244729b2c831aefe029e15f129ccbb3)

hanno-becker (approved):

PR #229, opened 2025-05-09
CBMC: Add proof and contract for polyvecl_uniform_eta

jakemas:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/114

Merge after https://github.com/pq-code-package/mldsa-native/pull/224

### DIFF
--- a/mldsa/polyvec.c
+++ b/mldsa/polyvec.c
@@ -42,10 +42,23 @@ void polyvecl_uniform_eta(polyvecl *v, const uint8_t seed[MLDSA_CRHBYTES],
                           uint16_t nonce)
 {
   unsigned int i;
+  uint16_t n = nonce;
 
   for (i = 0; i < MLDSA_L; ++i)
+  __loop__(
+    assigns(i, n, object_whole(v))
+    invariant(i <= MLDSA_L)
+    invariant(n == nonce + i) 
+    invariant(forall(k1, 0, i,
+      array_abs_bound(v->vec[k1].coeffs, 0, MLDSA_N, MLDSA_ETA + 1))))
   {
-    poly_uniform_eta(&v->vec[i], seed, nonce++);
+    poly t;
+    poly_uniform_eta(&t, seed, n);
+    n++;
+    /* Full struct assignment from local variables to simplify proof */
+    /* TODO: eliminate once CBMC resolves
+     * https://github.com/diffblue/cbmc/issues/8617 */
+    v->vec[i] = t;
   }
 }
 
--- a/mldsa/polyvec.h
+++ b/mldsa/polyvec.h
@@ -17,8 +17,29 @@ typedef struct
 } polyvecl;
 
 #define polyvecl_uniform_eta MLD_NAMESPACE(polyvecl_uniform_eta)
+/*************************************************
+ * Name:        polyvecl_uniform_eta
+ *
+ * Description: Sample vector of polynomials with uniformly random coefficients
+ *              in [-MLDSA_ETA, MLDSA_ETA] by performing rejection sampling on
+ *              the output stream from SHAKE256(seed|nonce). The vector has
+ *              MLDSA_L polynomials.
+ *
+ * Arguments:   - polyvecl *v: pointer to output polynomial vector
+ *              - const uint8_t seed[]: byte array with seed of length
+ *                MLDSA_CRHBYTES
+ *              - uint16_t nonce: 2-byte nonce
+ **************************************************/
 void polyvecl_uniform_eta(polyvecl *v, const uint8_t seed[MLDSA_CRHBYTES],
-                          uint16_t nonce);
+                          uint16_t nonce)
+__contract__(
+  requires(memory_no_alias(v, sizeof(polyvecl)))
+  requires(memory_no_alias(seed, MLDSA_CRHBYTES))
+  requires(nonce <= UINT16_MAX - MLDSA_L)
+  assigns(memory_slice(v, sizeof(polyvecl)))
+  ensures(forall(k0, 0, MLDSA_L,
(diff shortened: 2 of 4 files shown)


mldsa/polyvec.c
```diff
@@ -39,13 +39,33 @@ void polyvec_matrix_pointwise_montgomery(polyveck *t,
...
+    poly_uniform_eta(&t, seed, n);
+
+#ifdef CBMC
+#pragma CPROVER check push
+#pragma CPROVER check disable "unsigned-overflow"
+#endif
+    n++;
+#ifdef CBMC
+#pragma CPROVER check pop
+#endif
```
mkannwischer: The function is only called with `nonce=0`, so this can never overflow in practise. I'd prefer a stronger pre-condition over silencing the warning here.

mkannwischer (changes requested):
Thank you @jakemas!
One comment. Rest looks good.

mldsa/polyvec.c
jakemas: I've tried a few iterations of stronger pre-conditions, but I get stuck on `smt2` counter-example failure `SMT2 solver returned non-constant value for variable B2865`. When switching to `bitwuzla` the proof runs indefinitely and thus fails to provide a useful error message. I chatted with @rod-chapman on this, and he proposed this as the work around.

rod-chapman: Even with `requires(nonce==0)`, can we prove that this algorithm will terminate before `nonce==UINT_MAX`? (or whatever is the max value that will lead to overflow on next time...)

rod-chapman: Oh... the loop is bounded to MLDSA_L, so starting with nonce==0, it should be easy to prove it is bounded to MLDSA_L and therefore will never overflow, and then get rid of the pragmas.

jakemas: Got it! Thank you!

mldsa/polyvec.h
```diff
@@ -17,8 +17,29 @@ typedef struct
...
 void polyvecl_uniform_eta(polyvecl *v, const uint8_t seed[MLDSA_CRHBYTES],
-                          uint16_t nonce);
+                          uint16_t nonce)
+__contract__(
+  requires(memory_no_alias(v, sizeof(polyvecl)))
+  requires(memory_no_alias(seed, MLDSA_CRHBYTES))
+  requires(nonce == 0)
+  assigns(memory_slice(v, sizeof(polyvecl)))
+  ensures(forall(k0, 0, MLDSA_L,
+  array_abs_bound(v->vec[k0].coeffs, 0, MLDSA_N, MLDSA_ETA + 1)))
```
hanno-becker:
```suggestion
    array_abs_bound(v->vec[k0].coeffs, 0, MLDSA_N, MLDSA_ETA + 1)))
```

mldsa/polyvec.h
```diff
@@ -17,8 +17,29 @@ typedef struct
...
+ *                MLDSA_CRHBYTES
+ *              - uint16_t nonce: 2-byte nonce
+ **************************************************/
 void polyvecl_uniform_eta(polyvecl *v, const uint8_t seed[MLDSA_CRHBYTES],
-                          uint16_t nonce);
+                          uint16_t nonce)
+__contract__(
+  requires(memory_no_alias(v, sizeof(polyvecl)))
+  requires(memory_no_alias(seed, MLDSA_CRHBYTES))
+  requires(nonce == 0)
```
hanno-becker: @jakemas @mkannwischer What is the point of this parameter if it is always `0`?

hanno-becker: For uniformity with `polyveck_uniform_eta`, and reducing divergence from the ref implementation, let's keep it the parameter, but relax the precondition to `requires(none <= UINT16_MAX - MLDSA_L)`

hanno-becker (approved):

mldsa/polyvec.c
hanno-becker: @jakemas To make this uniform with `polyveck_uniform_eta`, where the nonce is _not_ 0, can you please use `UINT16_MAX - MLDSA_L` as the bound here?

hanno-becker (changes requested):
Looking at @mkannwischer's comment, I think we should use a `UINT16_MAX - MLDSA_L` as the precondition bound for `nonce` here, and follow the same approach in #230

mldsa/polyvec.h
jakemas: done [62c127a](https://github.com/pq-code-package/mldsa-native/pull/229/commits/62c127a3e1203212410c345a2a381e8dcb630a29)

mldsa/polyvec.h
jakemas: fixed [da5d51c](https://github.com/pq-code-package/mldsa-native/pull/229/commits/da5d51cec7e3a76d790f7e75042cb6c0596ccbcb)

mldsa/polyvec.h
jakemas: ok, done [da5d51c](https://github.com/pq-code-package/mldsa-native/pull/229/commits/da5d51cec7e3a76d790f7e75042cb6c0596ccbcb)

mldsa/polyvec.c
jakemas: done!

hanno-becker (approved):
Great, thanks @jakemas!

hanno-becker:
@jakemas Can you rebase the PR on top of `main`, so we don't have a merge commit in the history?

jakemas:
> @jakemas Can you rebase the PR on top of `main`, so we don't have a merge commit in the history?

aha, you caught me, I typed merge rather than rebase.. it's late!!

mkannwischer (approved):
LGTM. Thanks for the changes @jakemas!

PR #230, opened 2025-05-09
CBMC: Add proof and contract for polyveck_uniform_eta 

jakemas:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/123
- Depends on #224

### DIFF
--- a/mldsa/polyvec.c
+++ b/mldsa/polyvec.c
@@ -184,10 +184,23 @@ void polyveck_uniform_eta(polyveck *v, const uint8_t seed[MLDSA_CRHBYTES],
                           uint16_t nonce)
 {
   unsigned int i;
+  uint16_t n = nonce;
 
   for (i = 0; i < MLDSA_K; ++i)
+  __loop__(
+    assigns(i, n, object_whole(v))
+    invariant(i <= MLDSA_K)
+    invariant(n == nonce + i) 
+    invariant(forall(k1, 0, i,
+      array_abs_bound(v->vec[k1].coeffs, 0, MLDSA_N, MLDSA_ETA + 1))))
   {
-    poly_uniform_eta(&v->vec[i], seed, nonce++);
+    poly t;
+    poly_uniform_eta(&t, seed, n);
+    n++;
+    /* Full struct assignment from local variables to simplify proof */
+    /* TODO: eliminate once CBMC resolves
+     * https://github.com/diffblue/cbmc/issues/8617 */
+    v->vec[i] = t;
   }
 }
 
--- a/mldsa/polyvec.h
+++ b/mldsa/polyvec.h
@@ -165,8 +165,29 @@ typedef struct
 } polyveck;
 
 #define polyveck_uniform_eta MLD_NAMESPACE(polyveck_uniform_eta)
+/*************************************************
+ * Name:        polyveck_uniform_eta
+ *
+ * Description: Sample vector of polynomials with uniformly random coefficients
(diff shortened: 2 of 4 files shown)


mldsa/polyvec.c
```diff
@@ -201,16 +201,37 @@ int polyvecl_chknorm(const polyvecl *v, int32_t bound)
...
+    poly_uniform_eta(&t, seed, n);
+
+#ifdef CBMC
+#pragma CPROVER check push
+#pragma CPROVER check disable "unsigned-overflow"
+#endif
+    n++;
+#ifdef CBMC
+#pragma CPROVER check pop
+#endif
```
mkannwischer: The function is only called with nonce=L`, so this can never overflow in practice. I'd prefer a stronger pre-condition over silencing the warning here.

mkannwischer (changes requested):
Thanks @jakemas. One comment, rest looks good.

mldsa/polyvec.c
jakemas: Thank you, addressed in [f14848a](https://github.com/pq-code-package/mldsa-native/pull/230/commits/f14848aa1c77e7cf3379a0e4df9e94f473dc49f7)

mldsa/polyvec.c
```diff
@@ -39,13 +39,33 @@ void polyvec_matrix_pointwise_montgomery(polyveck *t,
...
```
hanno-becker: As in #229, let's use a precondition bound for `nonce` instead.

hanno-becker (changes requested):
We should add a precondition bound for `nonce` rather than disabling the overflow check.

@jakemas Please don't disable any standard macros without explicit discussion and comment in the code -- this is dangerous and undermines the claims we are making about the scope of the CBMC proofs.

mldsa/polyvec.c
jakemas: Ignore this, this PR is just staged ontop of the old one that has now been updated. This will be removed when I rebase

hanno-becker: Ah right, this was from the old PR -- I misread!

jakemas:
> We should add a precondition bound for `nonce` rather than disabling the overflow check. We have to be careful about changing the default set of CBMC checks, as it can render our messaging about the scope of CBMC inaccurate.

I did! I think you were reviewing `polyvecl_uniform_eta`. I rebased ontop of main now `poly_uniform_eta` is merged, and removed all other commits to remove this confusion.

mldsa/polyvec.h
```diff
@@ -165,8 +165,29 @@ typedef struct
...
+ *                MLDSA_CRHBYTES
+ *              - uint16_t nonce: 2-byte nonce
+ **************************************************/
 void polyveck_uniform_eta(polyveck *v, const uint8_t seed[MLDSA_CRHBYTES],
-                          uint16_t nonce);
+                          uint16_t nonce)
+__contract__(
+  requires(memory_no_alias(v, sizeof(polyveck)))
+  requires(memory_no_alias(seed, MLDSA_CRHBYTES))
+  requires(nonce == MLDSA_L)
```
hanno-becker: Can we have  `<= UINT16_MAX - MLDSA_K` here, as in #229?

jakemas: fixed

hanno-becker (approved):
Thanks @jakemas !

mkannwischer (approved):
LGTM. Thanks for the changes @jakemas!

PR #231, opened 2025-05-11
Re-factor keccak_squeeze() to improve proof time.

rod-chapman:
1. Avoid pointer arithmetic on "out" parameter, and use an integer-value out_offset instead, and array-index assignment.
2. Leave "outlen" parameter constant so it can be used unmodified in CBMC assigns() clauses.
3. Modify loop invariants appropriately.

Proof time on Apple M1 now 6 seconds instead of c. 130 seconds.

### DIFF
--- a/mldsa/fips202/fips202.c
+++ b/mldsa/fips202/fips202.c
@@ -475,7 +475,7 @@ __contract__(
  *              Modifies the state. Can be called multiple times to keep
  *              squeezing, i.e., is incremental.
  *
- * Arguments:   - uint8_t *out: pointer to output
+ * Arguments:   - uint8_t *out: pointer to output data
  *              - size_t outlen: number of bytes to be squeezed (written to out)
  *              - uint64_t *s: pointer to input/output Keccak state
  *              - unsigned int pos: number of bytes in current block already
@@ -498,14 +498,21 @@ __contract__(
   ensures(return_value <= r))
 {
   unsigned int i;
-  uint64_t lane;
-  uint8_t byte;
+  size_t out_offset = 0;
 
-  while (outlen > 0)
+  /* Reference: This code is re-factored from the reference implementation
+   * to facilitate proof with CBMC and to improve readability.
+   *
+   * Take a mutable copy of outlen to count down the number of bytes
+   * still to squeeze. The initial value of outlen is needed for the CBMC
+   * assigns() clauses. */
+  size_t bytes_to_go = outlen;
+
+  while (bytes_to_go > 0)
   __loop__(
-    assigns(i, outlen, pos, out, memory_slice(s, sizeof(uint64_t) * MLD_KECCAK_LANES), memory_slice(out, outlen))
-    invariant(outlen <= loop_entry(outlen))
-    invariant(out == loop_entry(out) + (loop_entry(outlen) - outlen))
+    assigns(i, bytes_to_go, pos, out_offset, memory_slice(s, sizeof(uint64_t) * MLD_KECCAK_LANES), memory_slice(out, outlen))
+    invariant(bytes_to_go <= outlen)
+    invariant(out_offset == outlen - bytes_to_go)
     invariant(pos <= r)
   )
   {
@@ -514,18 +521,18 @@ __contract__(
       KeccakF1600_StatePermute(s);
       pos = 0;
     }
-    for (i = pos; i < r && i < pos + outlen; i++)
+    for (i = pos; i < r && i < pos + bytes_to_go; i++)
     __loop__(
-      assigns(i, out, memory_slice(out, outlen))
-      invariant(i >= pos && i <= r && i <= pos + outlen)
-      invariant(out == loop_entry(out) + (i - pos))
+      assigns(i, out_offset, memory_slice(out, outlen))
+      invariant(i >= pos && i <= r && i <= pos + bytes_to_go)
+      invariant(out_offset == loop_entry(out_offset) + (i - pos))
     )
     {
-      lane = s[i / 8];
-      byte = (uint8_t)((lane >> (8 * (i % 8))) & 0xFF);
-      *out++ = byte;
+      const uint64_t lane = s[i / 8];
+      out[out_offset] = (uint8_t)((lane >> (8 * (i % 8))) & 0xFF);
+      out_offset++;
     }
-    outlen -= i - pos;
+    bytes_to_go -= i - pos;
     pos = i;
   }
 


mldsa/fips202/fips202.c
```diff
@@ -498,14 +498,18 @@ __contract__(
   ensures(return_value <= r))
 {
   unsigned int i;
-  uint64_t lane;
-  uint8_t byte;
+  size_t out_offset = 0;
+
+  /* Take a mutable copy of outlen to count down the number of bytes still to
+   * squeeze. The initial (const) value of outlen is needed for the CBMC
+   * assigns() clauses. */
```
hanno-becker, line 508: Can you not refer to this as `old(outlen)` or `loop_entry(...)`?

mldsa/fips202/fips202.c
```diff
@@ -484,9 +484,9 @@ __contract__(
  *
  * Returns new position pos in current block
  **************************************************/
-static unsigned int keccak_squeeze(uint8_t *out, size_t outlen,
+static unsigned int keccak_squeeze(uint8_t *const out, const size_t outlen,
```
hanno-becker: We don't tend to use `const` annotations for by-value parameters -- unless we want to make a concerted push for this, I don't think we should introduce this in isolated places. Can you please remove this change?

rod-chapman: I find "const" useful since it tells me (the developer), the reviewer and the maintainer "This is supposed to be immutable, so don't try to change it." It also enables to compiler to reject any such attempt, which I generally like.

In the interests of consistency with the rest of the code, though, I will remove it here.

mldsa/fips202/fips202.c
rod-chapman: Don't think so. old() can only be used withing an ensures() contract. loop_entry() is not what I want to express the correct assigns() contract for the loop.

mldsa/fips202/fips202.c
```diff
@@ -475,7 +475,7 @@ __contract__(
  *              Modifies the state. Can be called multiple times to keep
  *              squeezing, i.e., is incremental.
  *
- * Arguments:   - uint8_t *out: pointer to output
+ * Arguments:   - uint8_t *out: const pointer to mutable output data
```
hanno-becker: Remove "const"

rod-chapman: Done

manastasova:
The changes look good to me! Thank you, @rod-chapman!

manastasova (approved):

jakemas:
Looks good to me!

mldsa/fips202/fips202.c
```diff
@@ -498,14 +498,18 @@ __contract__(
   ensures(return_value <= r))
 {
   unsigned int i;
-  uint64_t lane;
-  uint8_t byte;
+  size_t out_offset = 0;
+
+  /* Take a mutable copy of outlen to count down the number of bytes still to
+   * squeeze. The initial (const) value of outlen is needed for the CBMC
```
jakemas:
```suggestion
   * squeeze. The initial value of outlen is needed for the CBMC
```
Nit: adjust comment to reflect the removal of the `const`

jakemas (approved):

mldsa/fips202/fips202.c
```diff
@@ -498,14 +498,18 @@ __contract__(
...
 
-  while (outlen > 0)
+  while (bytes_to_go > 0)
   __loop__(
-    assigns(i, outlen, pos, out, memory_slice(s, sizeof(uint64_t) * MLD_KECCAK_LANES), memory_slice(out, outlen))
-    invariant(outlen <= loop_entry(outlen))
-    invariant(out == loop_entry(out) + (loop_entry(outlen) - outlen))
+    assigns(i, bytes_to_go, pos, out_offset, memory_slice(s, sizeof(uint64_t) * MLD_KECCAK_LANES), memory_slice(out, outlen))
+    invariant(bytes_to_go <= loop_entry(bytes_to_go))
+    invariant(out_offset == loop_entry(out_offset) + (loop_entry(bytes_to_go) - bytes_to_go))
```
hanno-becker:
```suggestion
    invariant(out_offset == outlen - bytes_to_go)
```

mldsa/fips202/fips202.c
```diff
@@ -498,14 +498,18 @@ __contract__(
...
+  size_t bytes_to_go = outlen;
 
-  while (outlen > 0)
+  while (bytes_to_go > 0)
   __loop__(
-    assigns(i, outlen, pos, out, memory_slice(s, sizeof(uint64_t) * MLD_KECCAK_LANES), memory_slice(out, outlen))
-    invariant(outlen <= loop_entry(outlen))
-    invariant(out == loop_entry(out) + (loop_entry(outlen) - outlen))
+    assigns(i, bytes_to_go, pos, out_offset, memory_slice(s, sizeof(uint64_t) * MLD_KECCAK_LANES), memory_slice(out, outlen))
+    invariant(bytes_to_go <= loop_entry(bytes_to_go))
```
hanno-becker:
```suggestion
    invariant(bytes_to_go <= outlen)
```

mldsa/fips202/fips202.c
```diff
@@ -475,7 +475,7 @@ __contract__(
  *              Modifies the state. Can be called multiple times to keep
  *              squeezing, i.e., is incremental.
  *
- * Arguments:   - uint8_t *out: pointer to output
+ * Arguments:   - uint8_t *out: pointer to mutable output data
```
hanno-becker: Nit: I prefer the original formulation. Calling `out` out as 'mutable' suggests that the data stored at `out` at call-time is being operated on, but that's not the case: It's just overwritten.

mldsa/fips202/fips202.c
```diff
@@ -475,7 +475,7 @@ __contract__(
...
  *              - size_t outlen: number of bytes to be squeezed (written to out)
```
hanno-becker, line 479: Can you use add ` /* Reference: ... */` comment while at it, noting the change in the implementation?

hanno-becker (changes requested):
Nice performance improvement, @rod-chapman! Some nits.

mldsa/fips202/fips202.c
rod-chapman: Done

mldsa/fips202/fips202.c
rod-chapman: Good simplification! Done.

mldsa/fips202/fips202.c
rod-chapman: Done

mldsa/fips202/fips202.c
rod-chapman: I see the point. I guess the word "mutable" does distinguish between "write only" and "read/write". I will remove it.

mldsa/fips202/fips202.c
rod-chapman: DOne

hanno-becker (approved):

PR #232, opened 2025-05-13
Add spec and proof for `polyvec_matrix_expand`

manastasova:
Solves https://github.com/pq-code-package/mldsa-native/issues/137.

### DIFF
--- a/mldsa/polyvec.c
+++ b/mldsa/polyvec.c
@@ -11,14 +11,38 @@
 void polyvec_matrix_expand(polyvecl mat[MLDSA_K],
                            const uint8_t rho[MLDSA_SEEDBYTES])
 {
-  unsigned int i, j;
+  /* Reference: This code is re-factored from the reference implementation
+   * to facilitate proof with CBMC and to improve readability.
+   *
+   * The temporary polyvecl and poly variables simplify the CBMC
+   * memory model by removing nested access patterns. Loop invariant
+   * checks are performed on the local variable tmp_polyvecl before
+   * assigning it to the output matrix. */
+
+  unsigned int i;
 
   for (i = 0; i < MLDSA_K; ++i)
+  __loop__(
+    assigns(i, memory_slice(mat, MLDSA_K * sizeof(polyvecl)))
+    invariant(i <= MLDSA_K)
+    invariant(forall(k1, 0, i, forall(l1, 0, MLDSA_L, array_bound(mat[k1].vec[l1].coeffs, 0, MLDSA_N, 0, MLDSA_Q))))
+  )
   {
+    unsigned int j;
+    polyvecl tmp_polyvecl;
+
     for (j = 0; j < MLDSA_L; ++j)
+    __loop__(
+      assigns(j, memory_slice(&tmp_polyvecl, sizeof(polyvecl)))
+      invariant(j <= MLDSA_L)
+      invariant(forall(l1, 0, j, array_bound(tmp_polyvecl.vec[l1].coeffs, 0, MLDSA_N, 0, MLDSA_Q)))
+    )
     {
-      poly_uniform(&mat[i].vec[j], rho, (i << 8) + j);
+      poly temp_poly;
+      poly_uniform(&temp_poly, rho, (i << 8) + j);
+      tmp_polyvecl.vec[j] = temp_poly;
     }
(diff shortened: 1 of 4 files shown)


mldsa/polyvec.c
```diff
@@ -11,14 +11,38 @@
...
+   * The temporary polyvecl and poly variables simplify the CBMC
+   * memory model by removing nested access patterns. Loop invariant
+   * checks are performed on the local variable tmp_polyvecl before
+   * assigning it to the output matrix. */
+
+  unsigned int i;
 
   for (i = 0; i < MLDSA_K; ++i)
+  __loop__(
+    assigns(i, memory_slice(mat, MLDSA_K * sizeof(polyvecl)))
```
mkannwischer, line 26: If I remember correctly then this should be a `whole_object(mat)`

mldsa/polyvec.c
```diff
@@ -11,14 +11,38 @@
...
+    invariant(i <= MLDSA_K)
+    invariant(forall(k1, 0, i, forall(l1, 0, MLDSA_L, array_bound(mat[k1].vec[l1].coeffs, 0, MLDSA_N, 0, MLDSA_Q))))
+  )
   {
+    unsigned int j;
+    polyvecl tmp_polyvecl;
+
     for (j = 0; j < MLDSA_L; ++j)
+    __loop__(
+      assigns(j, memory_slice(&tmp_polyvecl, sizeof(polyvecl)))
```
mkannwischer, line 36: Same here? Not sure.

mldsa/polyvec.c
```diff
@@ -11,14 +11,38 @@
...
 
   for (i = 0; i < MLDSA_K; ++i)
+  __loop__(
+    assigns(i, memory_slice(mat, MLDSA_K * sizeof(polyvecl)))
+    invariant(i <= MLDSA_K)
+    invariant(forall(k1, 0, i, forall(l1, 0, MLDSA_L, array_bound(mat[k1].vec[l1].coeffs, 0, MLDSA_N, 0, MLDSA_Q))))
+  )
   {
+    unsigned int j;
+    polyvecl tmp_polyvecl;
```
mkannwischer, line 32: Is this a temporary workaround for https://github.com/diffblue/cbmc/issues/8617? If so, please add an appropriate comment with a TODO.

mldsa/polyvec.c
```diff
@@ -11,14 +11,38 @@
...
+
     for (j = 0; j < MLDSA_L; ++j)
+    __loop__(
+      assigns(j, memory_slice(&tmp_polyvecl, sizeof(polyvecl)))
+      invariant(j <= MLDSA_L)
+      invariant(forall(l1, 0, j, array_bound(tmp_polyvecl.vec[l1].coeffs, 0, MLDSA_N, 0, MLDSA_Q)))
+    )
     {
-      poly_uniform(&mat[i].vec[j], rho, (i << 8) + j);
+      poly temp_poly;
```
mkannwischer, line 41: Same here.

mkannwischer (changes requested):
Thanks @manastasova. 

The performance penalty for the copying is noticable (3% overall on some platforms). 
Is this a temporary workaround for https://github.com/diffblue/cbmc/issues/8617, or is this unrelated? If it is a temporary workaround, we can accept it now and revisit it later. Otherwise, I think we need to think about alternatives.

Another problem is that matrix_expand should really be using a 4-way batched poly_uniform (which then uses a 4-way batched Keccak). See #210. So this function will have to be rewritten anyway. Do you maybe want to do that already?

hanno-becker:
Thanks a lot @manastasova for tackling this!

> Another problem is that matrix_expand should really be using a 4-way batched poly_uniform (which then uses a 4-way batched Keccak). See https://github.com/pq-code-package/mldsa-native/issues/210. So this function will have to be rewritten anyway. Do you maybe want to do that already?

I think @mkannwischer raises an important point here. This is one of the main areas of code that we know will need to be rewritten, so we may want to do this first. What do you think, @manastasova?

manastasova:
Thank you @mkannwischer and @hanno-becker for the comments and suggestions!
I see the PR https://github.com/pq-code-package/mldsa-native/pull/233 already solved the x4 batched matrix generation. I am closing this PR.

PR #233, opened 2025-05-14
Sample matrix four polynomials at a time using 4-way batched Keccak

### LINKED ISSUE #210: Add `poly_uniform_4x`
https://github.com/pq-crystals/dilithium/blob/444cdcc84eb36b66fe27b3a2529ee48f6d8150c2/avx2/poly.h#L59

mkannwischer:
- Resolves #210 

This commit adds 4-way batched SHAKE128, which is used in poly_rej_uniform_4x.

polyvec_matrix_expand is then changed to sample 4 polynomials at a time. As L isn't always divisible by four, we use the approach of sampling across polynomial vector boundaries which is borrowed from mlkem-native.

This PR leaves some open TODOs:
 - Currently it is assumed that the compiler does not add any padding between polynomial vectors. That is dangerous and should be refactored. mlkem-native uses a separate data type for the matrix, but this requires to change a few routines.
 - The CBMC annotations for the added Keccak functions are left in place, but the proofs are not yet added.


jakemas:
Should we note that this also resolves issue https://github.com/pq-code-package/mldsa-native/issues/137?

mkannwischer:
> Should we note that this also resolves issue #137?

I have not done the proof yet. This will actually be a bit more work as you need to add the proofs for the used Keccak functions as well. Those can mostly be copied from mlkem-native, but I'd want to do that in a follow-up PR.

hanno-becker (changes requested):
I think we now have two copies of Keccak-F1600 now, one in `keccakf1600.c`, and one in `fips202.c` -- is that deliberate?

mkannwischer:
> I think we now have two copies of Keccak-F1600 now, one in `keccakf1600.c`, and one in `fips202.c` -- is that deliberate?

Well, it is "deliberate". The one in fips202.c should be removed to align with mlkem-native. But I think that is orthogonal to this PR.

mldsa/poly.c
```diff
@@ -346,6 +347,53 @@ void poly_uniform(poly *a, const uint8_t seed[MLDSA_SEEDBYTES], uint16_t nonce)
   }
 }
 
+void poly_rej_uniform_4x(poly *vec,
```
hanno-becker, line 351: I assume you have copied that from mlkem-native? Can you edit the copyright header?

mkannwischer: Fixed

mldsa/fips202/fips202x4.c
```diff
@@ -0,0 +1,143 @@
...
+ */
+
+/* References
+ * ==========
+ *
+ * - [FIPS203]
+ *   FIPS 203 Module-Lattice-Based Key-Encapsulation Mechanism Standard
+ *   National Institute of Standards and Technology
+ *   https://csrc.nist.gov/pubs/fips/203/final
+ */
```
jakemas: Should we add FIPS 204 here?

mldsa/fips202/fips202x4.c
```diff
@@ -0,0 +1,143 @@
...
+
+typedef mld_shake128x4ctx mld_shake256x4_ctx;
+
+/*TODO: Proof*/
+static void mld_keccak_absorb_once_x4(uint64_t *s, uint32_t r,
+                                      const uint8_t *in0, const uint8_t *in1,
+                                      const uint8_t *in2, const uint8_t *in3,
+                                      size_t inlen, uint8_t p)
+__contract__(
+  requires(memory_no_alias(s, sizeof(uint64_t) * MLK_KECCAK_LANES * MLK_KECCAK_WAY))
```
jakemas: I know the proofs aren't hooked up, but we use `MLD_KECCAK_LANES` rather than `MLK_*` here.

mldsa/fips202/fips202x4.c
```diff
@@ -0,0 +1,143 @@
...
+  requires(r <= sizeof(uint64_t) * MLK_KECCAK_LANES)
+  requires(memory_no_alias(in0, inlen))
+  requires(memory_no_alias(in1, inlen))
+  requires(memory_no_alias(in2, inlen))
+  requires(memory_no_alias(in3, inlen))
+  assigns(memory_slice(s, sizeof(uint64_t) * MLK_KECCAK_LANES * MLK_KECCAK_WAY)))
+{
+  while (inlen >= r)
+  __loop__(
+    assigns(inlen, in0, in1, in2, in3, memory_slice(s, sizeof(uint64_t) * MLK_KECCAK_LANES * MLK_KECCAK_WAY))
```
jakemas: `MLK_KECCAK_WAY` undefined

mldsa/poly.h
```diff
@@ -304,6 +304,22 @@ __contract__(
...
   ensures(array_bound(a->coeffs, 0, MLDSA_N, 0, MLDSA_Q))
 );
+#define poly_rej_uniform_4x MLD_NAMESPACE(poly_rej_uniform_4x)
+/*************************************************
+ * Name:        poly_rej_uniform_x4
+ *
+ * Description: Generate four polynomials using rejection sampling
+ *              on (pseudo-)uniformly random bytes sampled from a seed.
+ *
+ * Arguments:   - mlk_poly *vec:
```
jakemas:
```suggestion
 * Arguments:   - poly *vec:
```

mldsa/fips202/fips202x4.c
mkannwischer: good catch! Fixed.

mldsa/fips202/fips202x4.c
mkannwischer: good catch! Fixed.

mldsa/poly.h
mkannwischer: Fixed.

mldsa/fips202/fips202x4.c
mkannwischer: These references are auto-generated in mlkem-native. I've removed it for now.

We don't have stack zeroization yet. Once we add it we can add an appropriate reference and add the autogeneration of references.

mldsa/fips202/fips202x4.c
```diff
@@ -0,0 +1,134 @@
...
+
+#include <string.h>
+#include "fips202.h"
+#include "fips202x4.h"
+#include "keccakf1600.h"
+
+typedef mld_shake128x4ctx mld_shake256x4_ctx;
+
+/*TODO: Proof*/
+static void mld_keccak_absorb_once_x4(uint64_t *s, uint32_t r,
```
hanno-becker, line 16: Seeing you're introducing contracts and invariants already, is adding the proofs not merely a matter of adding harness and Makefile? Maybe we should do it in one go?

mldsa/fips202/keccakf1600.h
```diff
@@ -0,0 +1,105 @@
+/*
+ * Copyright (c) The mlkem-native project authors
+ * Copyright (c) The mldsa-native project authors
+ * SPDX-License-Identifier: Apache-2.0
```
hanno-becker: Apache-2.0 OR ISC OR MIT

mldsa/fips202/keccakf1600.c
```diff
@@ -0,0 +1,403 @@
+/*
+ * Copyright (c) The mlkem-native project authors
+ * Copyright (c) The mldsa-native project authors
+ * SPDX-License-Identifier: Apache-2.0
```
hanno-becker: Apache-2.0 OR ISC OR MIT

mldsa/fips202/fips202x4.h
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright (c) The mlkem-native project authors
+ * Copyright (c) The mldsa-native project authors
+ * SPDX-License-Identifier: Apache-2.0
```
hanno-becker: Apache-2.0 OR ISC OR MIT

mkannwischer: fixed.

mldsa/fips202/keccakf1600.c
mkannwischer: fixed.

mldsa/fips202/keccakf1600.h
mkannwischer: fixed.

mldsa/fips202/fips202x4.c
mkannwischer: I've added the proofs now. Please take a look.

mldsa/polyvec.c
```diff
@@ -3,22 +3,64 @@
...
     }
+
+    /*
+     * This call writes across polyvec boundaries for L=5 and L=7.
+     */
+    /* TODO: This is actually not safe; we cannot be sure that the compiler
+     * does not introduce any padding here. Need to refactor the data type to
+     * polymat as in mlkem-native.
+     */
+    poly_rej_uniform_4x(&mat[(i) / MLDSA_L].vec[(i) % MLDSA_L], seed_ext);
```
hanno-becker, line 51: The naming and signature for `poly_uniform` and `poly_rej_uniform_4x` differ now, which we should fix. Can you open an issue?

hanno-becker (approved):
We should unify naming and signature between `poly_uniform` and `poly_rej_uniform_4x`, but this can be a follow-up. Thanks a lot @mkannwischer!

mldsa/polyvec.c
mkannwischer: Okay.

mkannwischer: done: #241

PR #235, opened 2025-05-14
Switch these proofs to use Z3 rather than Bitwuzla

rod-chapman:
I have found 3 functions whose proof is slow with Bitwuzla, but much faster with Z3.

Results on Mac M1:

polyvecl_unpack_z() - from 90s to 5s
poly_invntt_tomont() - from 75s to 5s
mld_invntt_layer() - from 65s to 30s

These were some of the most expensive proofs in the library, so reduce wall-clock time for the full proof run, even on large multi-core machines.

### DIFF
--- a/proofs/cbmc/invntt_layer/Makefile
+++ b/proofs/cbmc/invntt_layer/Makefile
@@ -25,7 +25,7 @@ USE_DYNAMIC_FRAMES=1
 
 # Disable any setting of EXTERNAL_SAT_SOLVER, and choose SMT backend instead
 EXTERNAL_SAT_SOLVER=
-CBMCFLAGS=--bitwuzla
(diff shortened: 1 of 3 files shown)


rod-chapman:
Data: on r7gn with -j64, wall-clock time to prove all 105 functions:
main: 2m 52 seconds
this branch: 1m 52 seconds

manastasova:
This is an impressive speedup @rod-chapman!
I wonder if there are known scenarios where `z3` is faster than `bitwuzla` (e.g. nested loops, arrays/pointers, etc.) and vice versa.

manastasova (approved):

jakemas:
Thank you for investigating Rod!

jakemas (approved):

PR #244, opened 2025-05-16
Consolidate poly_uniform and poly_uniform_4x

mkannwischer:
This commit consolidates the previous inconsitency in the function names and semantics of poly_uniform and poly_(rej_)uniform_4x.

- poly_rej_uniform_4x is renamed to poly_uniform_4x
- The signature of poly_uniform is changes to not contain the nonce as a seperate uint16_t, but instead receiving an extended seed that already contains the 2 byte nonce packed (inline with that poly_uniform_4x gets)
- Consequently, the stream128_init function becomes unncessesary and gets removed along with it's CBMC proof.
- For consistency we also rename the stream128_ functions to mld_xof_ functions
- CBMC proofs are adjusted accordingly.

Resolves https://github.com/pq-code-package/mldsa-native/issues/241


mldsa/symmetric.h
```diff
@@ -39,15 +27,19 @@ __contract__(
...
-  shake128_squeezeblocks(OUT, OUTBLOCKS, STATE)
 #define stream256_init(STATE, SEED, NONCE) \
   mldsa_shake256_stream_init(STATE, SEED, NONCE)
 #define stream256_squeezeblocks(OUT, OUTBLOCKS, STATE) \
   shake256_squeezeblocks(OUT, OUTBLOCKS, STATE)
 
+#define mld_xof_ctx keccak_state
+#define mld_xof_init(CTX) shake128_init(CTX)
+#define mld_xof_absorb(CTX, IN, INBYTES) \
+  shake128_absorb(CTX, IN, INBYTES);     \
```
hanno-becker: Add `do { ... } while(0)` so there's no risk of breaking this in `for` or `if` block without `{ ... }` (which we don't use, but still).

mldsa/poly.c
```diff
@@ -323,16 +323,20 @@ __contract__(
...
-void poly_uniform(poly *a, const uint8_t seed[MLDSA_SEEDBYTES], uint16_t nonce)
+ *           - Modified rej_uniform interface to track offset directly.
+ *           - Pass nonce packed in the extended seed array instead of a third
+ *             argument.
+ * */
+void poly_uniform(poly *a, const uint8_t seed[MLDSA_SEEDBYTES + 2])
 {
   unsigned int ctr;
   unsigned int buflen = POLY_UNIFORM_NBLOCKS * STREAM128_BLOCKBYTES;
   uint8_t buf[POLY_UNIFORM_NBLOCKS * STREAM128_BLOCKBYTES];
```
hanno-becker: While at it, can you add an alignment directive here? Otherwise, the performance may vary based on where the compiler puts this.

hanno-becker (changes requested):
Great, thank you @mkannwischer! Two small suggestions for improvement.

mldsa/symmetric.h
mkannwischer: Yes, good point. Later we may want to refactor this to an absorb_once though.

mldsa/poly.c
mkannwischer: done

hanno-becker (approved):

PR #245, opened 2025-05-16
Sample `s1`, `s2` vectors using 4-way batched Keccak

mkannwischer:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/211
- Depends on #244 

This commit adds poly_uniform_eta_4x that samples four polynomials with
coefficients uniform in [-MLDSA_ETA,MLDSA_ETA].
This routine is required in key generation to sample the two vectors s1 and s2
of length L and K respectively.
For the 3 parameter sets we have to samples 4+4, 5+6, 7+8 polynomials in total.

For the first and last parameter set, we can batch for each vector separately:
4+4 is straight-forwardly batched using 2x4.
7+8 is batched using 2x4 for s2 and 2x4 for s1 with the latter wasting one
polynomial.

For the 5+6 parameter set, we combine the sampling and sample 3x4 polynomials
wasting one.
To accomodate the latter, we eliminate the polyveck_uniform_eta and
polyvecl_uniform_eta functions and instead call poly_uniform_eta_4x
directly in sign.c.

poly_uniform_eta is no longer needed as the batching (with losing at most
1 polynomial) is likely faster on the fast majority of platforms.
The proof for poly_uniform_eta is removed along with the proofs for
polyveck_uniform_eta and polyvecl_uniform_eta.

poly_uniform_eta_4x requires a shake256x4 API that allows for repeated
squeezing.
This is not present in mlkem-native, and, hence, is added newly in this commit.
CBMC proofs are deferred to a follow up PR.


hanno-becker (approved):
This looks great, thanks @mkannwischer!

PR #260, opened 2025-05-20
Add contracts and proof files for polyveck_chknorm()

rod-chapman:
Fixes #132 

Adds contracts and proof files for polyveck_chknorm()

Proofs, lint and tests OK.

PR #261, opened 2025-05-20
CBMC: Added proof and contract for polyvecl_chknorm

jakemas:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/122

### DIFF
--- a/mldsa/polyvec.h
+++ b/mldsa/polyvec.h
@@ -174,9 +174,14 @@ void polyvecl_pointwise_acc_montgomery(poly *w, const polyvecl *u,
  * Returns 0 if norm of all polynomials is strictly smaller than B <=
  *(MLDSA_Q-1)/8 and 1 otherwise.
  **************************************************/
-int polyvecl_chknorm(const polyvecl *v, int32_t B);
-
-
+int polyvecl_chknorm(const polyvecl *v, int32_t B)
+__contract__(
+  requires(memory_no_alias(v, sizeof(polyvecl)))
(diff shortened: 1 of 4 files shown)


mldsa/polyvec.h
```diff
@@ -150,9 +150,14 @@ void polyvecl_pointwise_acc_montgomery(poly *w, const polyvecl *u,
...
-int polyvecl_chknorm(const polyvecl *v, int32_t B);
-
-
+int polyvecl_chknorm(const polyvecl *v, int32_t B)
+__contract__(
+  requires(memory_no_alias(v, sizeof(polyvecl)))
+  requires(0 <= B && B <= (MLDSA_Q - 1) / 8)
+  requires(forall(k0, 0, MLDSA_L, 
+    array_bound(v->vec[k0].coeffs, 0, MLDSA_N, -REDUCE_RANGE_MAX, REDUCE_RANGE_MAX)))
+  ensures(return_value == 0 || return_value == 1)
```
mkannwischer, line 183: Can we add a post-condition here that return_value == 0 imples that all coefficients are within the bound?
Something like 
```
ensures((return_value == 0) == forall(k0, 0, MLDSA_L, 
    array_abs_bound(v->vec[k0].coeffs, 0, MLDSA_N, B)))
```

mkannwischer (changes requested):

mldsa/polyvec.h
mkannwischer: Discussed in #260  that we do not need it.

mkannwischer (approved):

PR #262, opened 2025-05-21
Refactor polyvecl_pointwise_acc_montgomery in preparation of native code

mkannwischer:
This commit refactors polyvecl_pointwise_acc_montgomery in preparation of pulling in native code for this function.

For the optimized backends, this function will perform the accumulation of the products on 64-bit words prior to reducing it back to 32-bit, e.g., using Montgomery reduction.

This requires a pre-condition that this accumulation cannot overflow an int64_t. This is the case as in all call sites the second input is the output of an NTT which is bounded by 9q.

As a side-effect this commit allows proving the bounds using CBMC. Before the function was simply calling poly_pointwise_montgomery which does not provide any bounds on the inputs/outputs. Consequently the proof of the accumulation of those outputs to not overflow would be impossible.

### DIFF
--- a/mldsa/polyvec.c
+++ b/mldsa/polyvec.c
@@ -191,14 +191,22 @@ void polyvecl_pointwise_poly_montgomery(polyvecl *r, const poly *a,
 void polyvecl_pointwise_acc_montgomery(poly *w, const polyvecl *u,
                                        const polyvecl *v)
 {
-  unsigned int i;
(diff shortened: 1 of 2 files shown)


hanno-becker (approved):
Great, thank you @mkannwischer -- nice performance bump! Let's prioritize getting the bounds formalized in CBMC.

PR #266, opened 2025-05-21
Refactor crypto_sign_signature_internal() to remove goto statements

### LINKED ISSUE #265: Re-factor crypto_sign_signature_internal() to use a loop statement rather than goto statements
The function crypto_sign_signature_internal() uses a label and goto statements to construct its main rejection-sampling loop.

This defies use of a loop invariant contract for CBMC. This Issue proposes to re-structure the code as a do/while() loop to facilitate proof, while retaining readability and reasonable correspondance with the specification/pseudo-code in FIPS 204.

rod-chapman:
Fixes #265 

Refactor crypto_sign_signature_internal() to remove goto statements and introduce a loop statement, so that an invariant can be inserted for later CBMC proof.

Tests and lint OK. Proof will follow after completion of proofs in polyvec.[hc]

### DIFF
--- a/mldsa/sign.c
+++ b/mldsa/sign.c
@@ -143,71 +143,85 @@ int crypto_sign_signature_internal(uint8_t *sig, size_t *siglen,
   polyveck_ntt(&s2);
   polyveck_ntt(&t0);
 
-rej:
-  /* Sample intermediate vector y */
-  polyvecl_uniform_gamma1(&y, rhoprime, nonce++);
-
-  /* Matrix-vector multiplication */
-  z = y;
-  polyvecl_ntt(&z);
-  polyvec_matrix_pointwise_montgomery(&w1, mat, &z);
-  polyveck_reduce(&w1);
-  polyveck_invntt_tomont(&w1);
-
-  /* Decompose w and call the random oracle */
-  polyveck_caddq(&w1);
-  polyveck_decompose(&w1, &w0, &w1);
-  polyveck_pack_w1(sig, &w1);
-
-  shake256_init(&state);
-  shake256_absorb(&state, mu, MLDSA_CRHBYTES);
-  shake256_absorb(&state, sig, MLDSA_K * MLDSA_POLYW1_PACKEDBYTES);
-  shake256_finalize(&state);
-  shake256_squeeze(sig, MLDSA_CTILDEBYTES, &state);
-  poly_challenge(&cp, sig);
-  poly_ntt(&cp);
-
-  /* Compute z, reject if it reveals secret */
-  polyvecl_pointwise_poly_montgomery(&z, &cp, &s1);
-  polyvecl_invntt_tomont(&z);
-  polyvecl_add(&z, &z, &y);
-  polyvecl_reduce(&z);
-  if (polyvecl_chknorm(&z, MLDSA_GAMMA1 - MLDSA_BETA))
+  /* Reference: This code is re-structured using a while(1),  */
+  /* with explicit "continue" statements (rather than "goto") */
+  /* to implement rejection of invalid signatures.            */
+  /* The loop statement also supplies a syntactic location to */
+  /* place loop invariants for CBMC.                          */
+  while (1)
+  __loop__(
+    invariant(true); /* TODO */
+  )
   {
-    goto rej;
-  }
+    /* Sample intermediate vector y */
+    polyvecl_uniform_gamma1(&y, rhoprime, nonce++);
 
-  /* Check that subtracting cs2 does not change high bits of w and low bits
-   * do not reveal secret information */
-  polyveck_pointwise_poly_montgomery(&h, &cp, &s2);
-  polyveck_invntt_tomont(&h);
-  polyveck_sub(&w0, &w0, &h);
-  polyveck_reduce(&w0);
-  if (polyveck_chknorm(&w0, MLDSA_GAMMA2 - MLDSA_BETA))
-  {
-    goto rej;
-  }
+    /* Matrix-vector multiplication */
+    z = y;
+    polyvecl_ntt(&z);
+    polyvec_matrix_pointwise_montgomery(&w1, mat, &z);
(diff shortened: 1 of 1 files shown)


mldsa/sign.c
```diff
@@ -2,6 +2,7 @@
  * Copyright (c) The mldsa-native project authors
  * SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
  */
+#include <stdbool.h>
```
mkannwischer: Can you please eliminate stdbool, please?
We don't use `stdbool.h` anywhere in mlkem-native/mldsa-native. It's not part of C90 and I don't see a good reason to use it here.

mkannwischer (changes requested):
Thanks @rod-chapman. 
One request - rest looks good to me.

mldsa/sign.c
```diff
@@ -143,71 +144,80 @@ int crypto_sign_signature_internal(uint8_t *sig, size_t *siglen,
   polyveck_ntt(&s2);
   polyveck_ntt(&t0);
 
-rej:
-  /* Sample intermediate vector y */
-  polyvecl_uniform_gamma1(&y, rhoprime, nonce++);
+  /* Reference: This code is re-structured using a while(true), */
+  /* with explicit "continue" statements rather than the "goto" */
+  /* so there is a syntactic location to place the loop         */
+  /* invariants for CBMC.                                       */
```
mkannwischer: nit: clarify that "continue" happens when a signature is rejected.

mldsa/sign.c
```diff
@@ -143,71 +144,80 @@ int crypto_sign_signature_internal(uint8_t *sig, size_t *siglen,
...
+    poly_ntt(&cp);
+
+    /* Compute z, reject if it reveals secret */
+    polyvecl_pointwise_poly_montgomery(&z, &cp, &s1);
+    polyvecl_invntt_tomont(&z);
+    polyvecl_add(&z, &z, &y);
+    polyvecl_reduce(&z);
+    if (polyvecl_chknorm(&z, MLDSA_GAMMA1 - MLDSA_BETA))
+    {
+      continue;
```
mkannwischer, line 187:
```suggestion
      /* reject */
      continue;
```

Similar for other continue statements.

hanno-becker (changes requested):
We should not spin forever trying to sign but should introduce a clean abort mechanism after a certain number of attempts. The current version will also not prove correct in CBMC because `nonce` will overflow at some point. See `Loop Bounds` in FIPS 204. The minimum allowed lower limit is 814.

jakemas:
> We should not spin forever trying to sign but should introduce a clean abort mechanism after a certain number of attempts. The current version will also not prove correct in CBMC because `nonce` will overflow at some point. See `Loop Bounds` in FIPS 204. The minimum allowed lower limit is 814.

I was thinking the same thing. FIPS requires a finite algorithms, these limits are outlined in Appendix C, Table 3.

Specifically for ML-DSA.Sign_internal, the limit is 814 iterations of the rejection sampling loop. This provides a probability of failure of less than 2^-256.

From Section 6.2 (footnote 10):
>*Implementations may limit the number of iterations in this loop to not exceed a finite maximum value. If this option is used and the maximum number of iterations is exceeded without producing a valid signature, the signing algorithm shall return a constant that represents an error and no other output, destroying the results of the unsuccessful signing attempts.*

Reasonable limit seems to be 814, it certainly can't be less.

mkannwischer:
> We should not spin forever trying to sign but should introduce a clean abort mechanism after a certain number of attempts. The current version will also not prove correct in CBMC because `nonce` will overflow at some point. See `Loop Bounds` in FIPS 204. The minimum allowed lower limit is 814.
> 

Good point. Agreed! I'd prefer to stick to UINT16_MAX / L (which is when the overflow happens inside of the sampling of the masking vector y). That is closer to the reference implementation.

mkannwischer:
> > We should not spin forever trying to sign but should introduce a clean abort mechanism after a certain number of attempts. The current version will also not prove correct in CBMC because `nonce` will overflow at some point. See `Loop Bounds` in FIPS 204. The minimum allowed lower limit is 814.
> 
> I was thinking the same thing. FIPS requires a finite algorithms, these limits are outlined in Appendix C, Table 3.

That is not true. It says "Implementations **should not** bound the number of iterations in these loops".
You can bound it if you want to, but you shoulnd't do it unless you have to. 
I actually think this is a bug in the specifation...

hanno-becker:
Thanks @mkannwischer! I agree. I think one must bound the loop to avoid overflow of the nonce. Also, from a practical perspective one should bound the loop to not spin forever if there's a bug in the signing procedure.

mldsa/sign.c
rod-chapman: It's needed to get the literal "true". I guess I could just replace that by "1".

rod-chapman: Didn't realize it wasn't in C90...

mkannwischer: Yes, please replace by 1.
stdbool got added in C99. Same applies to stdint - but that is basically impossible to eliminate.

mldsa/sign.c
rod-chapman: Done

mldsa/sign.c
rod-chapman: Done

rod-chapman:
Suggest we take the "don't spin forever" problem as a distinct issue. This one is about fixing the loop structure for CBMC so we can make progress on proof.

hanno-becker:
> Suggest we take the "don't spin forever" problem as a distinct issue. This one is about fixing the loop structure for CBMC so we can make progress on proof.

Note that we won't be able to do the proof without a bound on the loop, because the nonce _will_ overflow.

But I'm OK doing this separately nonetheless.

mkannwischer (approved):
LGTM! Thanks @rod-chapman.

PR #286, opened 2025-05-24
Update to nixpkgs 25.05

mkannwischer:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/285

Similar to https://github.com/pq-code-package/mlkem-native/pull/1041
Additionally, refactoring the CI to align with mlkem-native

[NixOS 25.05 ](https://nixos.org/blog/announcements/2025/nixos-2505/) has been released yesterday bringing various version updates also to nixpkgs.

Notable changes for us are: 
 - Default clang/llvm is now 19; we switch to the default in the core shells
 - Default gcc is now 14; we switch to it in the core shells
 - gcc 7, zig_0_10, zig_0_11 are not longer supported in 25.05; we instead
   take it from nixpkgs 24.05
 - zig_0_14 has been added; we test in CI now as well
 - Unicorn and protobuf versions are now compatible with SLOTHY simpliying the
   build
 - z3 has been updated to 4.15.0; we take it from 25.05
 - clang_20, bitwuzla, qemu now have suitable versions in 25.05; we no longer
   need to take them from unstable.
 - The CBMC build in nixpkgs includes 3 patches now instead of 2. The third patch is not compatible and not needed for cbmc 6.6.0. We, hence, have to copy the first two patches and apply them locally. This makes the patch slightly more complicated.
 - A few dependency updates for which we do not require specific versions

### DIFF
--- a/.github/workflows/all.yml
+++ b/.github/workflows/all.yml
@@ -1,0 +1,45 @@
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+
+name: CI
+permissions:
+  contents: read
+on:
+  workflow_dispatch:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["main"]
+    types: [ "opened", "synchronize" ]
+
+jobs:
+  base:
+    name: Base
+    permissions:
+      contents: 'read'
+      id-token: 'write'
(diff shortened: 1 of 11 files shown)


hanno-becker:
Thank you also for the CI restructuring, @mkannwischer !

PR #293, opened 2025-05-27
Refactor and simplify repeated hashing code in sign.c

rod-chapman:
Introduce mld_H() function (local, static in sign.c) to factor our repeated sequences of calls to
shake256_init/absorb/finalize/squeeze() functions.

Should make proof of top-level functions easier.

Also add CBMC proof files for this new function.

### DIFF
--- a/mldsa/sign.c
+++ b/mldsa/sign.c
@@ -91,6 +91,55 @@ int crypto_sign_keypair(uint8_t *pk, uint8_t *sk)
   return crypto_sign_keypair_internal(pk, sk, seed);
 }
 
+/*************************************************
+ * Name:        mld_H
+ *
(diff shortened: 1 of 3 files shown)


jakemas:
Seems like a reasonable change, adding the `benchmark` flag to see if this effects performance.

mkannwischer (approved):
Thanks @rod-chapman! Looks fine.

I think it may be better to introduce differnt variants for different numbers of inputs (i.e., mld_H1, mld_H2, mld_H3).
But we can proceed as is and consolidate that once CBMC proofs are done.

rod-chapman:
Thanks and noted. Let's merge this if we can.

PR #299, opened 2025-06-03
nix: Simplify CBMC build

mkannwischer:
Ports https://github.com/pq-code-package/mlkem-native/pull/1064


mkannwischer:
To save some CI time, I included this in #300

PR #302, opened 2025-06-04
Add native Keccak backend for x86_64 and AArch64

mkannwischer:


hanno-becker (approved):
Thanks @mkannwischer !

PR #310, opened 2025-06-06
CBMC: Improve proof performance of polyveck_add

mkannwischer:
polyveck_add is our slowest proofs taking ~1600 seconds in CI. This commit removes the post-condition of the called function poly_add. The post-condition is not needed anyway.


mkannwischer:
Improves total proof runtime for ML-DSA-44 from 30 minutes to 7 minutes. That will save us quite a bit of money in combination with #309.

mkannwischer:
> Do we really not need the post-condition, or did we just not need it _yet_?

The `polyveck_add` and `polyvecl_add` contract do currently not have that post-condition, so as it stands right now, it's not useful.

Coming up with the proper post-conditions is much harder than I thought. Let's discuss here: https://github.com/pq-code-package/mldsa-native/issues/311

mkannwischer:
Closed in favour of working out the proper post-condition in #311

PR #316, opened 2025-06-09
Strengthen post-condition of `polyveck_add` and `polyvecl_add`

### LINKED ISSUE #311: CBMC: Strengthen post-condition for `polyveck_add` and `polyvecl_add`
Currently our polynomial additions (`polyveck_add` and `polyvecl_add`) have no bounds on the outputs. However, we do require bounds in some places. It is not immediately clear how those should be defined. There are 3 places where we need additions:

 1) Sign: 
 ```
    polyvecl_invntt_tomont(&z);
    polyvecl_add(&z, &y);
    polyvecl_reduce(&z);
  ```
  
  - reduce requires inputs in (INT32_MIN, REDUCE_DOMAIN_MAX=INT32_MAX-(1 << 22))
  - z (output of intt) is bounded by (-q, +q)
  - y (output of uniform_gamm1) is bounded by (gamma1 <= 2^19)
  - easiest would be to require the inputs to the addition to be (-q,q), then the outputs are (-2q, +2q)
    
2) Sign:
```
    polyveck_add(&w0, &h);
    n = polyveck_make_hint(&h, &w0, &w1);
 ```
  - polyveck_make_hint has no requirements --> all good 

 3) KeyGen:
 ```
   polyveck_invntt_tomont(&t1);
   polyveck_add(&t1, &s2);
   polyveck_caddq(&t1);
   polyveck_power2round(&t1, &t0, &t1);
 ```
 - polyveck_power2round requires standard represenatitives.
 - caddq requires inputs in (-q, +q) to produce standard representatives.
 - s2 is bounded by a very small value (MLDSA_ETA = 2/4) 
  - t1 (output of intt) is currently bounded by (-q, +q).
  - It is not clear to me how this can be proved to be correct. 

This is looking like we need a stronger bound for INTT. Is that possible @hanno-becker?

mkannwischer:
- Resolves #311 

In https://github.com/pq-code-package/mldsa-native/issues/311, it was
identified that the current post-conditions are insufficent to complete
the higher level proofs.

There are 3 different places where polynomial additions are required
and they come with different input/output bounds:
Case 1: polyvecl_add in Signing
Inputs (-q, q) -- actual bound is tigher, but this suffices
Outputs (-2q, 2q)

Case 2: polyveck_add in Signing
Inputs (-q, q)
Outputs (-2q, 2q) -- no output bound required

Case 3: polyveck_add in Key Generation
Inputs (-4211139, 4211139) and (-4, 4) -- can relax to something like (-3q/4, 3q/4) and (-q/4, q/4)
Outputs (-q, q) -- this is a strict bound for the following to work.

It appears best to prove correctness of the addition itself, then the bound
reasoning can happen in the higher level functions.

This commit adds that functional correctness proof for polyvecl_add
and polyveck_add.
It only becomes feasible with a recent update to CBMC which also
allows removal of the workaround using whole-struct assignments.

### DIFF
--- a/mldsa/polyvec.c
+++ b/mldsa/polyvec.c
@@ -172,8 +172,9 @@ void polyvecl_add(polyvecl *u, const polyvecl *v)
   __loop__(
     assigns(i, memory_slice(u, sizeof(polyvecl)))
     invariant(i <= MLDSA_L)
-    invariant(forall(k0, i, MLDSA_L,
-              forall(k1, 0, MLDSA_N, u->vec[k0].coeffs[k1] == loop_entry(*u).vec[k0].coeffs[k1]))))
+    invariant(forall(k0, i, MLDSA_L, 
+              forall(k1, 0, MLDSA_N, u->vec[k0].coeffs[k1] == loop_entry(*u).vec[k0].coeffs[k1])))
+    invariant(forall(k4, 0, i, forall(k5, 0, MLDSA_N, u->vec[k4].coeffs[k5] == loop_entry(*u).vec[k4].coeffs[k5] + v->vec[k4].coeffs[k5]))))
   {
     poly_add(&u->vec[i], &v->vec[i]);
   }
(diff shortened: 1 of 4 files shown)


hanno-becker (approved):
LGTM, thanks @mkannwischer

PR #317, opened 2025-06-13
AVX2: Add native implementation of  `rej_uniform` and `rej_eta`

mkannwischer:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/304
- Resolves https://github.com/pq-code-package/mldsa-native/issues/305

Based on public-domain implementation from
https://github.com/pq-crystals/dilithium/blob/master/avx2/rejsample.c


mldsa/native/x86_64/src/rej_uniform_avx2.c
```diff
@@ -0,0 +1,75 @@
...
+#if defined(MLD_ARITH_BACKEND_X86_64_DEFAULT) && \
+    !defined(MLD_CONFIG_MULTILEVEL_NO_SHARED)
+
+#include <immintrin.h>
+#include <stdint.h>
+#include <string.h>
+#include "arith_native_x86_64.h"
+#include "consts.h"
+
+unsigned int mld_rej_uniform_avx2(int32_t *MLD_RESTRICT r, const uint8_t *buf)
```
hanno-becker: Can you document the desired buffer length, please? (e.g. by using `const uint8_t buf[N]`)

Also, in the upstream AVX2 code I see ominous `+8` in the buffer sizes which suggest some overreading. How is this being dealt with here? Can you confirm that those functions are exercised with asan and minimally allocated buffers, so we would catch overreads?

hanno-becker: In general, can you document the differences between this code and the upstream one?

hanno-becker (changes requested):
Thanks @mkannwischer! We need more documentation of the differences between this implementation and the upstream one please, as well as buffer lengths in the function signatures.

mldsa/native/x86_64/src/rej_uniform_avx2.c
mkannwischer: The +8 has been accounted for by adjusting the loop boundary of the first loop. No overreading occurs.
I added a comment explaining the differences and also added the buffer sizes explicitly.

mldsa/native/x86_64/src/rej_uniform_avx2.c
```diff
@@ -0,0 +1,84 @@
...
+
+#include <immintrin.h>
+#include <stdint.h>
+#include <string.h>
+#include "arith_native_x86_64.h"
+#include "consts.h"
+
+/*
+ * Reference: The pqcrystals implementation assumes a buffer that is 8 bytes
+ *.           larger as the first loop overreads by 8 bytes that are then dis
```
jakemas:
```suggestion
 *            larger as the first loop overreads by 8 bytes that are then
```

mldsa/native/x86_64/src/rej_uniform_avx2.c
```diff
@@ -0,0 +1,84 @@
...
+ *            discared. We instead to not pad the buffer and do not overread.
```
jakemas:
```suggestion
 *            discarded. We instead do not pad the buffer and do not overread.
```

mkannwischer:
Thanks @jakemas for catching these typos. I fixed them just now.

mkannwischer:
Will be merged together with #329

Issue #295, opened 2025-05-29
Remove NISTKAT tests


mkannwischer:
The NISTKAT tests introduce unnecessary dependencies. The testcoverage of the functional tests, KAT tests, and the ACVP tests should be sufficient. 
See https://github.com/pq-code-package/mlkem-native/pull/1053

PR #323, opened 2025-06-16
Remove the NISTKAT test

willieyz:
- Solves #295 

The NISTKAT tests introduce unnecessary dependencies. The testcoverage of the functional tests, KAT tests, and the ACVP tests should be sufficient.


scripts/tests
```diff
@@ -194,7 +194,6 @@ class SCHEME(Enum):
 class TEST_TYPES(Enum):
     FUNC = 1
     BENCH = 2
```
mkannwischer, line 196: Adjust the numbering to be 1,2,3,4... again

PR #343, opened 2025-06-27
Remove workarounds for CBMC issue 8617

rod-chapman:
Remove all but one instance of workaround for CBMC issue 8617 now that CBMC 6.7.0 has been rolled out.

The final workaround in polyvec_matrix_expand() requires some additional refactoring of the code, so will be dealt with under a separate PR.


mkannwischer (approved):
Beautiful. 8/10/10 minutes for the cbmc CI jobs looks great. in main I see 9/10/20.

PR #345, opened 2025-06-27
Refactor poly_uniform_4x() to explicitly return 4 polynomials

rod-chapman:
Refactor poly_uniform_4x() to explicitly return 4 polynomials via 4 parameters, not via 1 parameter.

This prevents the danger of writing across array element boundaries for a caller, and also simplifies proof of the caller, and allows for the removal of the work-around for CBMC Issue 8617 in polyvec_matrix_expand()

polyvec_matrix_expand() also avoids four whole-struct assignments, so a small performance win.


proofs/cbmc/poly_uniform_4x/poly_uniform_4x_harness.c
```diff
@@ -6,7 +6,10 @@
 void harness(void)
 {
   poly *r0;
-  uint8_t(*seed)[MLD_ALIGN_UP(MLDSA_SEEDBYTES + 2)];
+  poly *r1;
+  poly *r2;
+  poly *r3;
+  uint8_t seed[4][MLD_ALIGN_UP(MLDSA_SEEDBYTES + 2)];
```
mkannwischer: Can you change the type of seed back please?

hanno-becker: Agree, this is a soundness issue. With this harness, you can remove the memory validity preconditions from the function contract, and it would still pass. Thanks for spotting, @mkannwischer !

rod-chapman: Like this:
```
uint8_t(*seed)[MLD_ALIGN_UP(MLDSA_SEEDBYTES + 2)];
```
my clang-format kept changing it to:
```
uint8_t (*seed)[MLD_ALIGN_UP(MLDSA_SEEDBYTES + 2)];
```
(adding 1 space), then it failed the "lint" check.

I will change it back, though.

mkannwischer: that's odd. what's your `clang-format --version`?

rod-chapman: Hmm... inside the NIX shell, it says 19.1.7.

Outside NIX (and therefore in my EMACS buffer), it says 20.1.5

PR #346, opened 2025-07-02
AVX2/AArch64: Add native `poly_pointwise_montgomery`

### LINKED ISSUE #336: AArch64: Add `poly_pointwise_montgomery` assembly
#334 showed noticiable performance improvement when implementing `poly_pointwise_montgomery` using assembly.
We should add that in a separate PR. The accumulating variant is not needed for now.

### LINKED ISSUE #337: AVX2: Add `poly_pointwise_montgomery` assembly

mkannwischer:
- Resolves #336
- Resolves #337

This hoists out the native pointwise multiplication from #334 and also adds similar code for AVX2.
Note that it does not yet modify the matrix-vector multiplication - that one is hence not using this code yet. 
That will be done in a follow-up PR which will make it easier to understand the performance impact.


mldsa/native/x86_64/src/pointwise.S
```diff
@@ -0,0 +1,125 @@
+/*
+ * Copyright (c) The mlkem-native project authors
```
jakemas: This came from upstream dilithium repo? There is no pointwise x86 in mlkem-native yet?

mkannwischer: Good point. That was copy-pasta.

mkannwischer: fixed.

Issue #319, opened 2025-06-13
AArch64: Add native implementation of `rej_eta`


mkannwischer:
Following #317, we should implement the same for AArch64. I am not aware of any prior implementations of it. One can draw inspiration from https://github.com/pq-code-package/mlkem-native/blob/main/mlkem/src/native/aarch64/src/rej_uniform_asm.S.

PR #329, opened 2025-06-19
Add native implementation of `rej_uniform` (AArch64 + AVX2)

mkannwischer:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/318
- Resolves #304 

AArch64:
Written from scratch closely following
https://github.com/pq-code-package/mlkem-native/blob/main/dev/aarch64_clean/src/rej_uniform_asm.S

AVX2:
Based on public-domain implementation from
https://github.com/pq-crystals/dilithium/blob/master/avx2/rejsample.c

PR #348, opened 2025-07-08
Add native imlementation of rej_eta (AArch64 + AVX2)

mkannwischer:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/319
- Resolves https://github.com/pq-code-package/mldsa-native/issues/305
- Depends on #329 

AVX2:
Based on public-domain implementation from
https://github.com/pq-crystals/dilithium/blob/master/avx2/rejsample.c

AArch64:
- Implements mld_rej_uniform_eta2_aarch64() for eta=2 case
- Implements mld_rej_uniform_eta4_aarch64() for eta=4 case
- Autogenerates required look-up table

Note that we reject modulo a very small number (9 or 15). Doing that in 32-bits
appears wasteful. Doing it on 8-bit values makes the look-up table approach
infeasible as it would require a 2^16-entry table. We insetead opt for the
16-bit approch which is close to the ML-KEM rejection sampling.
Some additional arithmetic is required which we merge into the final
copying loop. That is also where the expansion from 16-bit values to
32-bit values happens.


mkannwischer:
Solid improvements for key generation. I'd like to merge this change.

mldsa/native/aarch64/src/arith_native_aarch64.h
```diff
@@ -27,6 +27,13 @@ extern const int32_t mld_aarch64_intt_zetas_layer78[];
 extern const int32_t mld_aarch64_intt_zetas_layer123456[];
 
 extern const uint8_t mld_rej_uniform_table[];
+extern const uint8_t mld_rej_uniform_eta_table[];
+
+#define MLD_AARCH64_REJ_UNIFORM_ETA2_BUFLEN \
+  (1 * 136) /* POLY_UNIFORM_ETA_NBLOCKS * SHAKE256_RATE */
```
hanno-becker: Typo: POLY_UNIFORM_ETA2_NBLOCKS? How is this calculated?

mldsa/native/aarch64/src/arith_native_aarch64.h
```diff
@@ -27,6 +27,13 @@ extern const int32_t mld_aarch64_intt_zetas_layer78[];
...
+
+#define MLD_AARCH64_REJ_UNIFORM_ETA4_BUFLEN \
+  (2 * 136) /* POLY_UNIFORM_ETA_NBLOCKS * SHAKE256_RATE */
```
hanno-becker: Typo: POLY_UNIFORM_ETA4_NBLOCKS? How is this calculated?

mldsa/native/aarch64/src/arith_native_aarch64.h
mkannwischer: I added comments explaining how these numbers came about. 
For eta=2, the parameter appears to be selected heuristically as one expects to need 136.5 bytes. It depends on the relative performance of Keccak x1 and Keccak x4 whether it makes sense to sample 1 or 2 blocks initially. Overall it does not appear to matter much.

mldsa/native/aarch64/src/arith_native_aarch64.h
mkannwischer: comments added.

mldsa/native/aarch64/src/rej_uniform_eta2_asm.S
```diff
@@ -0,0 +1,329 @@
...
+
+    // Always copy MLDSA_N coefficients from the stack to the destination
+    mov final_copy_count, #0
+    mov output_tmp, output_tmp_base
+rej_uniform_eta2_final_copy:
+        ldr val0q, [output_tmp], #32
+        ldr val2q, [output_tmp, #-16]
+
+        // Reference:
+        // t0 = t0 - (205 * t0 >> 10) * 5;
```
hanno-becker: This is a bit confusing here since we seem to be using a different 2-power compared to the reference implementation?

mkannwischer: We have to, to be able to use high multiplication? Or what am I missing?

mldsa/native/aarch64/src/rej_uniform_eta4_asm.S
```diff
@@ -0,0 +1,308 @@
...
+        and     tmp1.8b, buf0.8b, tmp0.8b     // Low nibbles [L0, L1, L2, L3, L4, L5, L6, L7]
+        ushr    tmp2.8b, buf0.8b, #4          // High nibbles [H0, H1, H2, H3, H4, H5, H6, H7]
+
+        // Interleave low and high nibbles: L0,H0,L1,H1,L2,H2,L3,H3,...
+        zip1    tmp0.8b, tmp1.8b, tmp2.8b     // First 8 nibbles interleaved [L0,H0,L1,H1,L2,H2,L3,H3]
+        zip2    tmp3.8b, tmp1.8b, tmp2.8b     // Next 8 nibbles interleaved [L4,H4,L5,H5,L6,H6,L7,H7]
+
+        // Convert to 16-bit values
+        uxtl    val0.8h, tmp0.8b
+        uxtl    val1.8h, tmp3.8b
```
hanno-becker, line 161: Why not run the code below on `tmp0.8b` and `tmp3.8b`? That is, keep everything as-is but apply it to 64-bit half-vectors only. That should not be slower (?) and avoid the extension to 16-bit.

mldsa/native/aarch64/src/rej_uniform_eta2_asm.S
hanno-becker: I didn't mean to suggest that the ASM is changed. I was only reporting that I found the comment not helpful/confusing because it suggests that what the ASM is doing matches the C code, which is not the case.

mldsa/native/aarch64/src/arith_native_aarch64.h
hanno-becker: Thanks!

hanno-becker (approved):
I checked that the AVX2 logic is the one from the PQCrystals implementation. I did not try to verify the AVX2 logic itself. I checked the logic of the AArch64 implementation, which looks good to me. There may be other approaches one could explore prior to attempting verification, but this is certainly a fine start that allows us to make progress on the goal of performance parity. Thanks a lot @mkannwischer.

mldsa/native/aarch64/src/rej_uniform_eta4_asm.S
hanno-becker: Can be explored another time.

mldsa/native/aarch64/src/rej_uniform_eta2_asm.S
hanno-becker: @mkannwischer Feel free to remove or clarify the comment before you merge -- as you see fit.

mkannwischer: I added a clarifying comment explaining the difference to the reference implementation.

PR #365, opened 2025-07-15
Consistently use mld_assert instead of cassert

mkannwischer:
cassert turns into a CBMC proof assertion inside of CBMC proofs, but it turns into a no-op otherwise.
mld_assert behaves the save inside of CBMC proofs, but also turns into a run-time assertion in debug builds.

### DIFF
--- a/mldsa/polyvec.c
+++ b/mldsa/polyvec.c
@@ -296,21 +296,7 @@ void mld_polyvecl_pointwise_acc_montgomery(mld_poly *w, const mld_polyvecl *u,
       t += (int64_t)u->vec[j].coeffs[i] * v->vec[j].coeffs[i];
     }
 
-    /* Substitute j == MLSDA_L into the loop invariant to get... */
-    cassert(j == MLDSA_L);
-    cassert(t >= -(int64_t)MLDSA_L * (MLDSA_Q - 1) * (MLD_NTT_BOUND - 1));
-    cassert(t <= (int64_t)MLDSA_L * (MLDSA_Q - 1) * (MLD_NTT_BOUND - 1));
-
-    /* ...and therefore... */
-    cassert(t >= -MONTGOMERY_REDUCE_STRONG_DOMAIN_MAX);
-    cassert(t < MONTGOMERY_REDUCE_STRONG_DOMAIN_MAX);
-
-    /* ...which meets the "strong" case of mld_montgomery_reduce() */
     r = mld_montgomery_reduce(t);
-
-    /* ...and therefore we can assert a stronger bound on r */
-    cassert(r > -MLDSA_Q);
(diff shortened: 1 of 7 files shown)


rod-chapman:
What happens when an mld_assert() contains a forall quantifier or something else that isn't executable?

mkannwischer:
> What happens when an mld_assert() contains a forall quantifier or something else that isn't executable?

You can make use of `mld_assert_bound` / `mld_assert_abs_bound`. If that is insufficient, one can consider adding additional `mld_assert_` functions.

jakemas (approved):

mldsa/polyvec.c
```diff
@@ -270,20 +271,20 @@ void mld_polyvecl_pointwise_acc_montgomery(mld_poly *w, const mld_polyvecl *u,
     }
 
     /* Substitute j == MLSDA_L into the loop invariant to get... */
-    cassert(j == MLDSA_L);
-    cassert(t >= -(int64_t)MLDSA_L * (MLDSA_Q - 1) * (MLD_NTT_BOUND - 1));
-    cassert(t <= (int64_t)MLDSA_L * (MLDSA_Q - 1) * (MLD_NTT_BOUND - 1));
+    mld_assert(j == MLDSA_L);
```
hanno-becker: @mkannwischer Can we use this opportunity to check if we actually need those asserts?

mldsa/rounding.c
```diff
@@ -17,20 +18,20 @@ void mld_decompose(int32_t *a0, int32_t *a1, int32_t a)
 {
   *a1 = (a + 127) >> 7;
   /* We know a >= 0 and a < MLDSA_Q, so... */
-  cassert(*a1 >= 0 && *a1 <= 65472);
```
hanno-becker: Same here

mldsa/polyvec.c
rod-chapman: While assertions like these might not be strictly needed for Z3 to prove everything, they sometimes serve as documentation for a new human reader. Imagine someone from (say) TrailOfBits looking at this code for the first time ... such assertions are a chance for the developer to say "Dear reader, here's something important that has to True at this point in the program that might not be obvious..."

Please consider the human reader in evaluating the merits of these cases.

hanno-becker: I agree in principle, but here were merely restate the loop invariant which is just a few lines above.

rod-chapman: Agreed in this case.

hanno-becker: Removing all `cassert`s to check if the CI is still happy.

mkannwischer:
Note that updates in main have introduced additional casserts. Those should be changed to mld_assert (or removed as well).

Issue #271, opened 2025-05-22
Add support for Pairwise Consistency Test (PCT)


mkannwischer:
[FIPS140-3](https://csrc.nist.gov/csrc/media/Projects/cryptographic-module-validation-program/documents/fips%20140-3/FIPS%20140-3%20IG.pdf) requires PCT for ML-DSA (even though FIPS204 does not mandate it). 
[As in mlkem-native,](https://github.com/pq-code-package/mlkem-native/blob/97bfbfa49c42672acb4c9acd90c9ce82fe18f231/mlkem/config.h#L375) we should add a configuration called `MLD_CONFIG_KEYGEN_PCT` (off by default) that enables PCT at the end of keygen.

PR #368, opened 2025-07-15
Add Pairwise Consistency Test (PCT) for FIPS-204 IG compliance

jakemas:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/271

### DIFF
--- a/mldsa/config.h
+++ b/mldsa/config.h
@@ -59,4 +59,42 @@
 #define MLD_CONFIG_FIPS202_BACKEND_FILE "fips202/native/auto.h"
 #endif
 
+/******************************************************************************
+ * Name:        MLD_CONFIG_KEYGEN_PCT
+ *
+ * Description: Compliance with @[FIPS140_3_IG, p.87] requires a
+ *              Pairwise Consistency Test (PCT) to be carried out on a freshly
+ *              generated keypair before it can be exported.
+ *
+ *              Set this option if such a check should be implemented.
+ *              In this case, crypto_sign_keypair_internal and
+ *              crypto_sign_keypair will return a non-zero error code if the
+ *              PCT failed.
+ *
+ *              NOTE: This feature will drastically lower the performance of
+ *              key generation.
+ *
+ *****************************************************************************/
+/* #define MLD_CONFIG_KEYGEN_PCT */
+
+/******************************************************************************
+ * Name:        MLD_CONFIG_KEYGEN_PCT_BREAKAGE_TEST
+ *
+ * Description: If this option is set, the user must provide a runtime
+ *              function `static inline int mld_break_pct() { ... }` to
+ *              indicate whether the PCT should be made fail.
+ *
+ *              This option only has an effect if MLD_CONFIG_KEYGEN_PCT is set.
+ *
+ *****************************************************************************/
+/* #define MLD_CONFIG_KEYGEN_PCT_BREAKAGE_TEST
+   #if !defined(__ASSEMBLER__)
+   #include "sys.h"
+   static MLD_INLINE int mld_break_pct(void)
+   {
+       ... return 0/1 depending on whether PCT should be broken ...
+   }
+   #endif
+*/
+
 #endif /* !MLD_CONFIG_H */
--- a/mldsa/sign.c
+++ b/mldsa/sign.c
@@ -14,6 +14,79 @@
 #include "sign.h"
 #include "symmetric.h"
 
+
+static int mld_check_pct(uint8_t const pk[CRYPTO_PUBLICKEYBYTES],
+                         uint8_t const sk[CRYPTO_SECRETKEYBYTES])
+__contract__(
+  requires(memory_no_alias(pk, CRYPTO_PUBLICKEYBYTES))
+  requires(memory_no_alias(sk, CRYPTO_SECRETKEYBYTES))
+  ensures(return_value == 0 || return_value == -1)
+);
+
+#if defined(MLD_CONFIG_KEYGEN_PCT)
+/*************************************************
+ * [FIPS 140-3 IG]
+ * (https://csrc.nist.gov/csrc/media/Projects/cryptographic-module-validation-program/documents/fips%20140-3/FIPS%20140-3%20IG.pdf)
+ *
+ * TE10.35.02: Pair-wise Consistency Test (PCT) for DSA keypairs
+ *
+ * Purpose: Validates that a generated public/private key pair can correctly
+ * sign and verify data. Test performs signature generation using the private
+ * key (sk), followed by signature verification using the public key (pk).
+ * Returns 0 if the signature was successfully verified, non-zero if it cannot.
+ *
+ * Note: FIPS 204 requires that public/private key pairs are to be used only for
+ * the calculation and/of verification of digital signatures.
+ **************************************************/
+static int mld_check_pct(uint8_t const pk[CRYPTO_PUBLICKEYBYTES],
+                         uint8_t const sk[CRYPTO_SECRETKEYBYTES])
+{
+  uint8_t message[1] = {0};
+  uint8_t signature[CRYPTO_BYTES];
+  uint8_t pk_test[CRYPTO_PUBLICKEYBYTES];
+  size_t siglen;
+  int ret;
+
+  /* Copy public key for testing */
+  memcpy(pk_test, pk, CRYPTO_PUBLICKEYBYTES);
+
+  /* Sign a test message using the original secret key */
+  ret = crypto_sign_signature(signature, &siglen, message, sizeof(message),
+                              NULL, 0, sk);
+  if (ret == 0)
+  {
+#if defined(MLD_CONFIG_KEYGEN_PCT_BREAKAGE_TEST)
+    /* Deliberately break public key for testing purposes */
+    if (mld_break_pct())
+    {
+      pk_test[0] = ~pk_test[0];
+    }
+#endif /* MLD_CONFIG_KEYGEN_PCT_BREAKAGE_TEST */
+
+    /* Verify the signature using the (potentially corrupted) public key */
+    ret = crypto_sign_verify(signature, siglen, message, sizeof(message), NULL,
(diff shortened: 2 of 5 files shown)


mkannwischer (changes requested):
Thanks @jakemas.

However, note that keygen is already proved using CBMC, and this addition, hence, introduces a proof gap. At this stage we should avoid adding proof gaps.

I will do a more thorough review later.

mldsa/config.h
```diff
@@ -59,4 +59,41 @@
...
+/******************************************************************************
+ * Name:        MLD_CONFIG_KEYGEN_PCT
+ *
+ * Description: Compliance with @[FIPS140_3_IG, p.87] requires a
+ *              Pairwise Consistency Test (PCT) to be carried out on a freshly
+ *              generated keypair before it can be exported.
+ *
+ *              Set this option if such a check should be implemented.
+ *              In this case, crypto_sign_keypair_internal and
+ *crypto_sign_keypair will return a non-zero error code if the PCT failed.
```
mkannwischer: not: alignment

mldsa/sign.c
```diff
@@ -14,6 +14,76 @@
 #include "sign.h"
 #include "symmetric.h"
 
+static int mld_check_pct(uint8_t const pk[CRYPTO_PUBLICKEYBYTES],
+                         uint8_t const sk[CRYPTO_SECRETKEYBYTES]);
```
mkannwischer: This is missing the CBMC contract. 
It should then be proven for the actual implementation - see [this](https://github.com/pq-code-package/mlkem-native/tree/main/proofs/cbmc/check_pct).

mldsa/sign.c
```diff
@@ -14,6 +14,76 @@
...
+
+#if defined(MLD_CONFIG_KEYGEN_PCT)
+/*************************************************
+ * [FIPS 140-3 IG]
+ * (https://csrc.nist.gov/csrc/media/Projects/cryptographic-module-validation-program/documents/fips%20140-3/FIPS%20140-3%20IG.pdf)
+ *
+ * VE10.35.02: Pair-wise Consistency Test (PCT) for DSA keypairs
```
mkannwischer: Should be TE rather than VE?

mldsa/sign.c
```diff
@@ -14,6 +14,76 @@
...
+ * key (sk), followed by signature verification using the public key (pk).
+ * Returns 0 if the signature was successfully verified, non-zero if it cannot.
+ *
+ * Note: FIPS 204 requires that public/private key pairs are to be used only for
+ * the calculation and/of verification of digital signatures.
+ **************************************************/
+static int mld_check_pct(uint8_t const pk[CRYPTO_PUBLICKEYBYTES],
+                         uint8_t const sk[CRYPTO_SECRETKEYBYTES])
+{
+  uint8_t message[1] = {0};
```
mkannwischer, line 44: Out of curiosity: Does the FIPS IG suggest signing a single 0 byte, or is up to us?

mkannwischer (changes requested):
Thanks @jakemas  for this very important addition. 
A couple of comments. I think we should already add the CBMC proof to not introduce a proof gap into key generation.

mldsa/sign.c
mkannwischer: This will rely on the `crypto_sign_signature`function which is not proven yet, but you can just add a contract for that function and then it should hopefully just go through.

mldsa/sign.c
jakemas: Vendor Evidence vs Test Evidence. Happy to go with TE10.35.02

mldsa/sign.c
jakemas: It doesn't specify what should be signed, it's up to us.

jakemas:
Current CBMC issue:
```
** 1 of 7 failed (2 iterations)
VERIFICATION FAILED

[MLD_65_ref_check_pct.assertion.1] line 8 undefined function should be unreachable: FAILURE
```

I think this has to do with the makefile. This error is much harder to understand because of the `ifdefs` around the specificaiton.

mldsa/sign.c
rod-chapman: I think crypto_sign_signature() is proven... I remember doing it!

mldsa/sign.c
```diff
@@ -14,6 +14,81 @@
 #include "sign.h"
 #include "symmetric.h"
 
+
+static int mld_check_pct(uint8_t const pk[CRYPTO_PUBLICKEYBYTES],
```
rod-chapman, line 18: Is there a good reason why mld_check_pct() needs a distinct declaration/contract from its definition below? Can't they just be merged?

mldsa/sign.c
```diff
@@ -14,6 +14,81 @@
...
+  size_t siglen;
+  int ret;
+
+  /* Copy public key for testing */
+  memcpy(pk_test, pk, CRYPTO_PUBLICKEYBYTES);
+
+  /* Sign a test message using the original secret key */
+  ret = crypto_sign_signature(signature, &siglen, message, sizeof(message),
+                              NULL, 0, sk);
+  if (ret != 0)
```
rod-chapman: I am aesthetically opposed to the use of "goto" here... especially since there's only one of them. Wouldn't an simple "if" statement here suffice?

mldsa/sign.c
mkannwischer: haha, yes, now it is proven. We should not introduce any proof gaps with new PRs.

proofs/cbmc/check_pct/Makefile
```diff
@@ -0,0 +1,56 @@
...
+PROJECT_SOURCES += $(SRCDIR)/mldsa/sign.c
+
+CHECK_FUNCTION_CONTRACTS=mld_check_pct
+USE_FUNCTION_CONTRACTS=$(MLD_NAMESPACE)signature $(MLD_NAMESPACE)verify
+APPLY_LOOP_CONTRACTS=on
+USE_DYNAMIC_FRAMES=1
+
+# Disable any setting of EXTERNAL_SAT_SOLVER, and choose SMT backend instead
+EXTERNAL_SAT_SOLVER=
+CBMCFLAGS=--bitwuzla
```
rod-chapman: Why bitwuzla? Did you try Z3? When it works, Z3 is usually faster.. and we have more assurance in Z3 4.12.6.

mldsa/sign.c
```diff
@@ -14,6 +14,81 @@
 #include "sign.h"
 #include "symmetric.h"
 
+
+static int mld_check_pct(uint8_t const pk[CRYPTO_PUBLICKEYBYTES],
+                         uint8_t const sk[CRYPTO_SECRETKEYBYTES])
+__contract__(
+  requires(memory_no_alias(pk, CRYPTO_PUBLICKEYBYTES))
+  requires(memory_no_alias(sk, CRYPTO_SECRETKEYBYTES))
+);
```
rod-chapman, line 24: Explain the meaning of the return value (in the comment), and add post-condition here to say it can only be 0 or -1.

mldsa/sign.c
```diff
@@ -96,6 +171,12 @@ int crypto_sign_keypair_internal(uint8_t *pk, uint8_t *sk,
   /* Compute H(rho, t1) and write secret key */
   shake256(tr, MLDSA_TRBYTES, pk, CRYPTO_PUBLICKEYBYTES);
   mld_pack_sk(sk, rho, tr, key, &t0, &s1, &s2);
+
+  /* Pairwise Consistency Test (PCT) @[FIPS140_3_IG, p.87] */
+  if (mld_check_pct(pk, sk))
+  {
+    return -1;
```
rod-chapman, line 176: You need to update the comments and contracts on crypto_sign_keypair_internal() and crypto_sign_keypair() to indicate that they can return 0 (success) or (now) -1 (indicating that the PCT has failed).

mldsa/sign.c
jakemas: I was setting up the PCT as was done in mlkem-native, see the same set up here: https://github.com/pq-code-package/mlkem-native/blob/main/mlkem/src/kem.c#L110-L171

mldsa/sign.c
jakemas: Ok, changed to `if`

proofs/cbmc/check_pct/Makefile
jakemas: changed to `--smt2`

mldsa/sign.c
jakemas: Ok, updated comments (in `sign.h`) and contracts to accept 0 or -1

mldsa/sign.c
jakemas: Done:
- https://github.com/pq-code-package/mldsa-native/pull/368/files#diff-fd39060a4fa7e44d4d22ef721e140fc366d1646edefce715a9c5cf1b4a780828R32-R35
- https://github.com/pq-code-package/mldsa-native/pull/368/files#diff-d12e4a0d5f3866ace5acfdef8dddc001d184ecd24178119878189a9c5f242290R21-R22
- https://github.com/pq-code-package/mldsa-native/pull/368/files#diff-d12e4a0d5f3866ace5acfdef8dddc001d184ecd24178119878189a9c5f242290R50-R51

rod-chapman (approved):
Looks good to me now...

jakemas:
> Thanks. The code changes look good to me!
> 
> What was missed is adding the mld_check_pct to CBMC proof of `crypto_sign_keypair_internal` via `USE_FUNCTION_CONTRACTS`. Consequently that CBMC proof inlines the empty mld_check_pct function and ignoring the contract providing no meaning whatsoever. Please add it.

Thank you! [Added](https://github.com/pq-code-package/mldsa-native/pull/368/files#diff-82fbc0212253f7ea265d8bbea19ec630e559a1d05aa7f0c4b48636f05faa06c5R35) :)

mkannwischer (approved):
Thanks @jakemas for the update! LGTM.

This should be exercised in CI - I have opened #380 for that.

PR #370, opened 2025-07-16
Downgrade to z3 4.12.6 and adjust CBMC proofs

mkannwischer:
We have noticed a severe performance degradation of some CBMC proofs when moving from z3 4.12 to 4.13/4.14/4.15. This is currently blocking the proof of the signing functions.
This PR downgrades back to z3 4.12.6 - the last release prior to the performance regression. A few minor changes to the other CBMC proofs are required which are applied as part of this PR.

### DIFF
--- a/mldsa/poly.c
+++ b/mldsa/poly.c
@@ -1106,13 +1106,13 @@ void mld_polyz_unpack(mld_poly *r, const uint8_t *a)
   mld_assert_bound(r->coeffs, MLDSA_N, -(MLDSA_GAMMA1 - 1), MLDSA_GAMMA1 + 1);
 }
 
-void mld_polyw1_pack(uint8_t *r, const mld_poly *a)
+void mld_polyw1_pack(uint8_t r[MLDSA_POLYW1_PACKEDBYTES], const mld_poly *a)
 {
   unsigned int i;
 
-#if MLDSA_MODE == 2
-  mld_assert_bound(a->coeffs, MLDSA_N, 0, 44);
+  mld_assert_bound(a->coeffs, MLDSA_N, 0, (MLDSA_Q - 1) / (2 * MLDSA_GAMMA2));
 
+#if MLDSA_MODE == 2
   for (i = 0; i < MLDSA_N / 4; ++i)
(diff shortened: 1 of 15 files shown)


mkannwischer:
Changes in proof performance:
- ML-DSA-44: 12 minutes -> 21 minutes
- ML-DSA-65: 12 minutes -> 8 minutes
- ML-DSA-87: 13 minutes -> 8 minutes

Major bottleneck appears to be `polyvecl_pointwise_acc_montgomery` proof
 - ML-DSA-44: 268s -> 1068s
 - ML-DSA-65: 230s -> 269s
 - ML-DSA-87: 307s -> 238s
 
 Also `poly_pointwise_montgomery`:
 - ML-DSA-44: 71s -> 285s
 - ML-DSA-65: 56s -> 72s
 - ML-DSA-87: 104s -> 99s

mldsa/poly.c
```diff
@@ -1106,13 +1106,13 @@ void mld_polyz_unpack(mld_poly *r, const uint8_t *a)
...
 }
 
-void mld_polyw1_pack(uint8_t *r, const mld_poly *a)
+void mld_polyw1_pack(uint8_t r[MLDSA_POLYW1_PACKEDBYTES], const mld_poly *a)
 {
   unsigned int i;
 
-#if MLDSA_MODE == 2
-  mld_assert_bound(a->coeffs, MLDSA_N, 0, 44);
+  mld_assert_bound(a->coeffs, MLDSA_N, 0, (MLDSA_Q - 1) / (2 * MLDSA_GAMMA2));
```
hanno-becker, line 1113: Is that necessary? This is just the precondition.

hanno-becker (approved):
LGTM

mldsa/poly.c
hanno-becker: Can be done separately to not re-trigger CI.

mkannwischer: Agreed - it should be possible to remove that. Let's do that in a follow-up. We should re-visit all the asserts and check if they add any value.

Issue #364, opened 2025-07-15
Introduce constant-time barriers for caddq, decompose, and power2round


mkannwischer:
- Ralated to #152 

caddq, decompose, and power2round use patterns such as 
```
  a += (a >> 31) & MLDSA_Q;
  return a;
``` 
that a compiler may turn into a conditional when inlined (either through lto or via explicit inlining such as #363).
We should introduce appropriate constant-time barriers to ensure that these functions remain constant-time also with lto. 
Note that `power2round` is less important as it only processes public data. If the performance overhead is small, however, we may just consistently use constant-time barriers.

Issue #152, opened 2025-04-13
Introduce constant-time primitive for absolute value computation


hanno-becker:
`poly_chknorm` uses the pattern

```c
    t = a->coeffs[i] >> 31;
    t = a->coeffs[i] - (t & 2 * a->coeffs[i]);
```

to compute the absolute value of an `int32_t`. This should be wrapped in a constant-time function guarded by barriers making compilation into a branch unlikely.

PR #371, opened 2025-07-16
 Constant-time hardening for `caddq`, `poly_chknorm`, `decompose`

mkannwischer:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/364
- Resolves https://github.com/pq-code-package/mldsa-native/issues/152

This PR adds constant-time value barriers to 3 functions: `caddq`, `poly_chknorm`, `decompose`. 
The same approach as in mlkem-native is followed: If available, by default value barriers are implemeted using an empty inline asm block which marks the target value as clobbered. If inline assembly is not available (e.g., on Windows), we XOR with the value of a volatile global that's set to 0 ([as proposed by Bernstein here](https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/hqbtIGFKIpU/m/H14H0wOlBgAJ)).

CBMC proofs are added for the added auxilary functions and the proofs of `caddq`, `poly_chknorm`, `decompose`.  A config variation test is added to CI to test the volatile global optblocker config option.

Performance impact is considerable (up to 20% when using link-time optimization).


mldsa/ct.h
```diff
@@ -68,6 +66,18 @@ __contract__(ensures(return_value == b))
...
+ * The mld_ct_sel_int32 function below makes deliberate use of unsigned
+ * to signed integer conversion, which is implementation-defined
+ * behaviour. Here, we assume that uint16_t -> int16_t is inverse
+ * to int16_t -> uint16_t.
+ */
+#ifdef CBMC
+#pragma CPROVER check push
+#pragma CPROVER check disable "conversion"
+#endif
+
```
hanno-becker, line 125: Spurious /* TODO: proof */ below. Multiple times.

hanno-becker: Can be done in a follow-up.

Issue #356, opened 2025-07-10
CBMC: Extend crypto_sign_open proof to cover s=sm


mkannwischer:
Currently the contract of crypto_sign_open requires the message buffer m and the signed message buffer sm to be distinct. However, the API explicitly allows s=sm. We should extend the contract and proof to cover this special case - the code should already be supporting it.

Same applied to crypto_sign

PR #372, opened 2025-07-17
Refactor and add proof files for crypto_sign_signature_internal()

### LINKED ISSUE #272: CBMC: Add proof for `crypto_sign_signature_internal`

rod-chapman:
Fixes #272 

Refactor and add proof files for crypto_sign_signature_internal()

Introduce new local function mld_attempt_signature_generation() to simplify and stabilize proofs.

Strengthen contracts for
  mld_poly_make_hint()
  mld_polyveck_make_hint()
  mld_pack_sig()
in support of that goal.

Adjust CBMC flags for proof stability on macOS and Linux

### DIFF
--- a/mldsa/sign.c
+++ b/mldsa/sign.c
@@ -155,19 +155,153 @@ __contract__(
   shake256_squeeze(out, outlen, &state);
 }
 
+/* Reference: The reference implementation does not explicitly   */
+/* check the maximum nonce value, but instead loops indefinitely */
+/* (even when the nonce would overflow). Internally,             */
+/* sampling of y uses (nonceL), (nonceL+1), ... (nonce*L+L-1).   */
+/* Hence, there are no overflows if nonce < (UINT16_MAX - L)/L.  */
+/* Explicitly checking for this explicitly allows us to prove    */
+/* type-safety. Note that FIPS204 explicitly allows an upper-    */
+/* bound this loop of 814 (< (UINT16_MAX - L)/L) - see           */
+/* Appendix C of FIPS204.                                        */
+#define NONCE_UB ((UINT16_MAX - MLDSA_L) / MLDSA_L)
+
+/*************************************************
+ * Name:        attempt_signature_generation
+ *
+ * Description: Attempts to generate a single signature.
+ *
+ * Arguments:   - uint8_t *sig: pointer to output signature
+ *              - const uint8_t *mu: pointer to message or hash
+ *                                   of exactly MLDSA_CRHBYTES bytes
+ *              - const uint8_t *rhoprime: pointer to randomness seed
+ *              - uint16_t nonce: current nonce value
+ *              - const polyvecl mat[MLDSA_K]: expanded matrix
+ *              - const polyvecl *s1: secret vector s1
+ *              - const polyveck *s2: secret vector s2
+ *              - const polyveck *t0: vector t0
+ *
+ * Returns 0 if signature generation succeeds, -1 if rejected
+ *
+ * Reference: This code differs from the reference implementation
+ *            in that it factors out the core signature generation
+ *            step into a distinct function here in order to improve
+ *            efficiency of CBMC proof.
+ **************************************************/
+static int mld_attempt_signature_generation(
+    uint8_t *sig, const uint8_t *mu, const uint8_t rhoprime[MLDSA_CRHBYTES],
+    uint16_t nonce, const mld_polyvecl mat[MLDSA_K], const mld_polyvecl *s1,
+    const mld_polyveck *s2, const mld_polyveck *t0)
+__contract__(
+  requires(memory_no_alias(sig, CRYPTO_BYTES))
+  requires(memory_no_alias(mu, MLDSA_CRHBYTES))
+  requires(memory_no_alias(rhoprime, MLDSA_CRHBYTES))
+  requires(memory_no_alias(mat, MLDSA_K * sizeof(mld_polyvecl)))
+  requires(memory_no_alias(s1, sizeof(mld_polyvecl)))
+  requires(memory_no_alias(s2, sizeof(mld_polyveck)))
+  requires(memory_no_alias(t0, sizeof(mld_polyveck)))
+  requires(nonce <= NONCE_UB)
+  requires(forall(k1, 0, MLDSA_K, forall(l1, 0, MLDSA_L,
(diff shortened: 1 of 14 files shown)


mldsa/sign.c
```diff
@@ -204,74 +325,36 @@ int crypto_sign_signature_internal(uint8_t *sig, size_t *siglen,
   /* place loop invariants for CBMC.                          */
```
jakemas: Can remove this todo
```suggestion
 
```

mldsa/sign.c
```diff
@@ -155,19 +155,140 @@ __contract__(
...
+
+/*************************************************
+ * Name:        attempt_signature_generation
+ *
+ * Description: Attempts to generate a single signature.
+ *              This function encapsulates one iteration of the
+ *              rejection sampling process to help CBMC analysis.
+ *
+ * Arguments:   - uint8_t *sig: pointer to output signature
+ *              - const uint8_t *mu: pointer to message hash
```
jakemas, line 168: `message hash` is a little imprecise here and may confuse reader. Mu is not just a message hash, but a prepared premessage string that's hashed. But also, this isn't always `Mu`, this function gets called from sign internal, and that can sign either `Mu` or raw message types.

jakemas (dismissed):
One small comment here

mldsa/sign.c
```diff
@@ -155,19 +155,140 @@ __contract__(
...
+ *              - const uint8_t *rhoprime: pointer to randomness seed
+ *              - uint16_t nonce: current nonce value
+ *              - const polyvecl mat[MLDSA_K]: expanded matrix
+ *              - const polyvecl *s1: secret vector s1
+ *              - const polyveck *s2: secret vector s2
+ *              - const polyveck *t0: secret vector t0
+ *
+ * Returns 0 if signature generation succeeds, -1 if rejected
```
mkannwischer, line 184: Please add a ` Reference:` comment here on how this differs from the reference implementation.
I remember you did make a couple of other changes as well. Maybe now is a good time to write them down if you still remember them.
That will save us some time further down the line.

mldsa/sign.c
```diff
@@ -155,19 +155,140 @@ __contract__(
...
+ *              rejection sampling process to help CBMC analysis.
+ *
+ * Arguments:   - uint8_t *sig: pointer to output signature
+ *              - const uint8_t *mu: pointer to message hash
+ *              - const uint8_t *rhoprime: pointer to randomness seed
+ *              - uint16_t nonce: current nonce value
+ *              - const polyvecl mat[MLDSA_K]: expanded matrix
+ *              - const polyvecl *s1: secret vector s1
+ *              - const polyveck *s2: secret vector s2
+ *              - const polyveck *t0: secret vector t0
```
mkannwischer: t0 is not secret.

mldsa/sign.c
```diff
@@ -204,74 +325,36 @@ int crypto_sign_signature_internal(uint8_t *sig, size_t *siglen,
...
+    int result;
 
-    /* Check that subtracting cs2 does not change high bits of w and low bits
-     * do not reveal secret information */
-    mld_polyveck_pointwise_poly_montgomery(&h, &cp, &s2);
-    mld_polyveck_invntt_tomont(&h);
-    mld_polyveck_sub(&w0, &h);
-    mld_polyveck_reduce(&w0);
-    if (mld_polyveck_chknorm(&w0, MLDSA_GAMMA2 - MLDSA_BETA))
+    if (nonce == NONCE_UB)
```
mkannwischer, line 358: Please add a comment what is going on here and that this should never happen - also describe how it differs from the reference implementation.

mldsa/sign.h
```diff
@@ -83,7 +83,20 @@ int crypto_sign_signature_internal(uint8_t *sig, size_t *siglen,
                                    const uint8_t *m, size_t mlen,
                                    const uint8_t *pre, size_t prelen,
                                    const uint8_t rnd[MLDSA_RNDBYTES],
-                                   const uint8_t *sk, int externalmu);
+                                   const uint8_t *sk, int externalmu)
+__contract__(
+  requires(memory_no_alias(sig, CRYPTO_BYTES))
+  requires(memory_no_alias(siglen, sizeof(size_t)))
+  requires(memory_no_alias(m, mlen))
```
mkannwischer, line 97: This currently does not support aliasing sig=m. We will need to support that at some point, but this can be done as a follow-up. Please open an issue in that case - just like https://github.com/pq-code-package/mldsa-native/issues/356.

mkannwischer (changes requested):
Thanks @rod-chapman. Awesome progress!
A couple of comments - mostly on the documentation.

mldsa/sign.c
rod-chapman: Done

mldsa/sign.c
rod-chapman: Done. Also added Reference: comment in sign.h for crypto_sign_signature_internal() to mention it can return -1 in the case of nonce exhaustion.

mldsa/sign.h
rod-chapman: I have this worked out in the proof of crypto_sign. To come in next PR.

mldsa/sign.c
```diff
@@ -200,78 +329,44 @@ int crypto_sign_signature_internal(uint8_t *sig, size_t *siglen,
...
-     * do not reveal secret information */
-    mld_polyveck_pointwise_poly_montgomery(&h, &cp, &s2);
-    mld_polyveck_invntt_tomont(&h);
-    mld_polyveck_sub(&w0, &h);
-    mld_polyveck_reduce(&w0);
-    if (mld_polyveck_chknorm(&w0, MLDSA_GAMMA2 - MLDSA_BETA))
+    /* Reference: this code explicitly checks for exhaustion of nonce     */
+    /* values to provide predictable termination and results in that case */
+    /* Checking here also means that incrementing nonce below can also    */
+    /* be proven to be type-safe.                                         */
```
mkannwischer, line 357: Maybe something like this would be more clear?

Reference: The reference implementation does not explicitly check the maximum nonce value, but instead loops indefinitely (even when the nonce would overflow).
Internally, sampling of y uses (nonce*L), (nonce*L+1), ... (nonce*L+L-1), and, hence, there are no overflows if nonce < (UINT16_MAX - L)/L.
Explicitly checking for this explicitly allows us to prove type-safety.
Note that FIPS204 explicitly allows upper-bounding this loop by 814 (< (UINT16_MAX - L)/L) - see Appendix C of FIPS204.

PR #387, opened 2025-07-25
Fix mac os sha256sum #386

### LINKED ISSUE #386: Tests fail on macOS - sha256sum command not found
## What's Wrong

  When I run `make test` on macOS, I get this error:
  /bin/sh: sha256sum: command not found

  The tests seem to pass anyway, but this looks broken.

  ## How to See the Problem

  1. On macOS, run: `make test`
  2. You'll see the error message above
  3. The tests continue and report "Everything checks fine!" but something is clearly wrong.

Image

## Why This Happens

On searching, I found that macOS doesn't have the `sha256sum` command. Instead, it has `shasum -a 256` which does the same thing.

sparrowjack234:
Fixes #386 - macOS compatibility issue where `sha256sum` command is not available by default, causing silent hash validation failures during KAT tests.

Testing successful on MacOS-
Screenshot 2025-07-26 at 3 45 46 AM

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ all: build
 
 W := $(EXEC_WRAPPER)
 
-# Platform-specific hash command detection for cross-platform compatibility
+# Platform-specific hash command
 UNAME_S := $(shell uname -s)
 ifeq ($(UNAME_S),Darwin)
     SHA256_CMD = shasum -a 256
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,14 @@ all: build
 
 W := $(EXEC_WRAPPER)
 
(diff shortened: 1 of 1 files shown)


sparrowjack234:
> Thanks for your interest in mldsa-native and for opening a PR! However, we cannot accept contributions by someone we do not know. If you are interested in participating in the LFX mentorship program, please send your application through LFX platform: https://mentorship.lfx.linuxfoundation.org/project/4202ff8f-c683-497b-8d06-d801d02943f0

Thank you for the feedback. I understand the security considerations for cryptography projects. I'll apply through the official LFX mentorship channel till then I will try to understand and test the codebase and also look for contributing in other C libraries.

Thank You!!!

PR #392, opened 2025-07-29
CT: Make poly_chknorm constant flow 

### LINKED ISSUE #153: Make `poly_chknorm` constant flow
The current implementation of `poly_chknorm`,

```c
  for (i = 0; i < MLDSA_N; ++i)
  {
    /* Absolute value */
    t = a->coeffs[i] >> 31;
    t = a->coeffs[i] - (t & 2 * a->coeffs[i]);

    if (t >= B)
    {
      return 1;
    }
  }
```

aborts upon the first entry which exceeds the given norm.

While there is a comment indicating why this does not leak anything sensitive, the input data _is_ sensitive (from what I understand), so it seems safer to use a constant-flow implementation here.

mkannwischer:
- Resolves #153 
- Depends on #371

The reference implementation implements poly_chknorm in variables time.
It argues that while the input coefficients itself are secret in some call
sites, it is okay to leak which coefficient lead to rejection.
It, hence, does absolute value computation in constant-time and then checks
the bound using a conditional.

This approach appears safe, but somewhat unclean as it is still operating on
_secret data_. When performing constant-time testing it also requires a
number of declassifications.

This commit takes a more conservative approach and changes poly_chknorm to
a constant-time implementation in the hope tha the performance penalty is
acceptable.

A minor change is that the API of poly_chknorm is changed to returning
0xFFFFFFFF in the case of failure to be able to re-use existing constant-time
primitives.
CBMC proofs are adjusted accordingly.


hanno-becker (approved):
LGTM. I support the attempt to make the functions constant-time to avoid refined reasoning about partial declassification.

Issue #354, opened 2025-07-10
Update api.h to reflect all supported API functions (and rename to mldsa_native.h)


mkannwischer:
Currently `crypto_sign_verify_internal` and `crypto_sign_signature_internal` are non-static and exposed externally (but they are currently not part of the api.h). We should reconsider that and either make them static or add them to the api.h (which should be renamed mldsa_native.h). 

It would be good to learn from potential consumers if those functions are required or if the other functions are sufficient. 
@praveksharma could you comment on this for liboqs? 
@hanno-becker, did you check yet what APIs would be required by AWS-LC?

hanno-becker:
I will check.

jakemas:
https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/ml_dsa/ml_dsa.h

In aws-lc we use `*_internal` functions to perform continuous testing against the ACVP test vectors and KATs. See references https://github.com/aws/aws-lc/blob/main/crypto/evp_extra/p_pqdsa_test.cc#L1564-L1566 -- it includes internal keygen, sign, and verify (as we need to use the internal deterministic functions)

praveksharma:
liboqs also uses `*_internal` functions similarly.

Issue #162, opened 2025-04-13
Integrate mldsa-native into libOQS


hanno-becker:

hanno-becker:
@praveksharma Would you like to handle this?

praveksharma:
> [@praveksharma](https://github.com/praveksharma) Would you like to handle this?

Yes please.

hanno-becker:
@praveksharma Great. I assigned this to you, then.

mkannwischer:
@praveksharma: I think now would be a good time to prototype a liboqs integration to identify any issues early on.
Do you have time to look into that?

mkannwischer:
Any blockers? The goal is to have no patches whatsoever and follow the same structure as for mlkem-native.

mkannwischer:
Thank you for looking into this @praveksharma. 

We need to consolidate the FIPS202 implementations in mldsa-native to match the naming, API, and semantics in mlkem-native. Let's do that as a separate PR first. Can you give it a try @praveksharma?

mkannwischer:
@praveksharma, are you still working on this one, or do you need help with something? We have @jammychiou1 now who could possibly help with this soon.

mkannwischer:
As there has been no progress, I will give this a try myself now.

PR #403, opened 2025-07-31
Add liboqs integration

praveksharma:
This adds integration code for liboqs. The integration code is being tested here: https://github.com/open-quantum-safe/liboqs/pull/2222. Please note, that the current integration does not integrate liboqs' FIPS202 into mldsa-native (earlier mentioned here: https://github.com/pq-code-package/mldsa-native/issues/162#issuecomment-3141093735) 

Errors, as recorded at the link above and observed from testing locally, are primarily because of two issues:
1. All FIPS202 code is namespaced as `mldsa_*` which isn't presently a valid liboqs namespace. This is controlled by the [`FIPS202_NAMESPACE`](https://github.com/pq-code-package/mldsa-native/blob/77976820d20cf173c7fe466a6f272565764ec5cf/mldsa/fips202/fips202.h#L20) macro. If integration of liboqs' FIP202 into mldsa-native is not desirable/feasible at the moment I will add the required configuration option to `config.h` and switch these functions to a `PQCP_` namespace.
2. ACVP tests are disabled because mldsa-native does not expose internal function. This can be fixed once #354 is resolved.

### DIFF
--- a/integration/liboqs/ML-DSA-44_META.yml
+++ b/integration/liboqs/ML-DSA-44_META.yml
@@ -1,0 +1,63 @@
+name: ML-DSA-44
+type: signature
+claimed-nist-level: 2
+claimed-security: SUF-CMA
+length-public-key: 1312
+length-secret-key: 2560
+length-signature: 2420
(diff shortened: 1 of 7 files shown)


praveksharma:
Closing this as I continue to work on #404.

PR #410, opened 2025-08-04
Inline rounding.c and reduce.c

mkannwischer:
- Resolves #217 

With constant-time hardening in place, we should be able to safely inline reduce.c and rounding.c.
This should vastly improve non-LTO performance. As CI is using LTO, I do not expect much improvement there.

I performed benchmarks on Apple M1 (using ` ./scripts/tests bench -c MAC -r`, clang16, i.e., no LTO):
### ML-DSA-44
| Version | Keypair | Sign | Verify |
|---------|---------|------|--------|
| Main | 133,231 | 600,444 | 159,982 |
| Commit 1 | 130,432 | 595,679 | 157,727 |
| Commit 2 | 117,094 | 463,152 | 138,418 |

### ML-DSA-65
| Version | Keypair | Sign | Verify |
|---------|---------|------|--------|
| Main | 224,854 | 985,587 | 251,728 |
| Commit 1 | 219,835 | 975,350 | 248,134 |
| Commit 2 | 200,026 | 749,140 | 219,666 |

### ML-DSA-87
| Version | Keypair | Sign | Verify |
|---------|---------|------|--------|
| Main | 366,192 | 1,208,165 | 403,781 |
| Commit 1 | 360,544 | 1,195,145 | 398,647 |
| Commit 2 | 330,288 | 932,454 | 357,744 |

### Commit 1 vs Main (inline rounding.c)
| Algorithm | Keypair | Sign | Verify |
|-----------|---------|------|--------|
| ML-DSA-44 | **1.02x** | **1.01x** | **1.01x** |
| ML-DSA-65 | **1.02x** | **1.01x** | **1.01x** |
| ML-DSA-87 | **1.02x** | **1.01x** | **1.01x** |

### Commit 2 vs Main (inline reduce.c)
| Algorithm | Keypair | Sign | Verify |
|-----------|---------|------|--------|
| ML-DSA-44 | **1.14x** | **1.30x** | **1.16x** |
| ML-DSA-65 | **1.12x** | **1.32x** | **1.15x** |
| ML-DSA-87 | **1.11x** | **1.30x** | **1.13x** |

### Commit 2 vs Commit 1 (cumulative improvement)
| Algorithm | Keypair | Sign | Verify |
|-----------|---------|------|--------|
| ML-DSA-44 | **1.11x** | **1.29x** | **1.14x** |
| ML-DSA-65 | **1.10x** | **1.30x** | **1.13x** |
[...]

PR #411, opened 2025-08-04
Add native implementation of `poly_decompose`

mkannwischer:
- Depends on #410


mldsa/poly.c
```diff
@@ -174,9 +174,19 @@ void mld_poly_power2round(mld_poly *a1, mld_poly *a0, const mld_poly *a)
 
 void mld_poly_decompose(mld_poly *a1, mld_poly *a0, const mld_poly *a)
 {
+#if defined(MLD_USE_NATIVE_POLY_DECOMPOSE_88) && MLDSA_MODE == 2
+  /* TODO: proof */
+  mld_assert_bound(a->coeffs, MLDSA_N, 0, MLDSA_Q);
+  mld_poly_decompose_88_native(a1->coeffs, a0->coeffs, a->coeffs);
+#elif defined(MLD_USE_NATIVE_POLY_DECOMPOSE_32) && \
```
hanno-becker, line 189: Sorry for the dumb question, but why the name `decompose_32` and `decompose_88`?

mldsa/native/x86_64/src/poly_decompose_88_avx2.c
```diff
@@ -0,0 +1,68 @@
...
+      _mm256_load_si256(&mld_qdata.vec[MLD_AVX2_BACKEND_DATA_OFFSET_8XQ / 8]);
+  const __m256i hq = _mm256_srli_epi32(q, 1);
+  const __m256i v = _mm256_set1_epi32(11275);
+  const __m256i alpha = _mm256_set1_epi32(2 * MLDSA_GAMMA2);
+  const __m256i off = _mm256_set1_epi32(127);
+  const __m256i shift = _mm256_set1_epi32(128);
+  const __m256i max = _mm256_set1_epi32(43);
+  const __m256i zero = _mm256_setzero_si256();
+
+  for (i = 0; i < MLDSA_N / 8; i++)
```
hanno-becker, line 42: Same comment as for `cadd`: I think if we find this worthwhile from a performnace-perspective, we should add it in ASM straight away.

mldsa/native/aarch64/src/poly_decompose_32_asm.S
```diff
@@ -0,0 +1,114 @@
...
+
+#if defined(MLD_ARITH_BACKEND_AARCH64) && !defined(MLD_CONFIG_MULTILEVEL_NO_SHARED)
+
+.macro decompose32 a1, a0, input, temp
+        // Step 1: Compute ceil(a / 128) using floor((a + 127) / 128)
+        // This is the first part of computing a1 = floor(a / (2*GAMMA2))
+        // where 2*GAMMA2 = 523776. We break this into two steps:
+        // ceil(a / 128) followed by round(temp / 4092)
+        add \a1\().4s, \input\().4s, offset_127.4s
+        ushr \a1\().4s, \a1\().4s, #7
```
hanno-becker: Why don't we use a rounding shift right here?

mkannwischer: Is there a rounding shift that rounds up?

mldsa/poly.c
hanno-becker: Nevermind, got it.

mldsa/native/aarch64/src/poly_decompose_32_asm.S
hanno-becker: https://developer.arm.com/documentation/dui0801/l/A64-SIMD-Vector-Instructions/SRSHR--vector---A64- Is this not what we want?

hanno-becker: Ah, ok. Round-up... no, I don't think so.

mldsa/native/x86_64/src/poly_decompose_88_avx2.c
mkannwischer: As discussed offline, let's add this first to make faster progress and keep aarch64 and avx2 in sync, but replace it as soon as possible. I have opened https://github.com/pq-code-package/mldsa-native/issues/420.

mldsa/poly.c
```diff
@@ -174,9 +174,19 @@ void mld_poly_power2round(mld_poly *a1, mld_poly *a0, const mld_poly *a)
...
+    (MLDSA_MODE == 3 || MLDSA_MODE == 5)
+  /* TODO: proof */
+  mld_assert_bound(a->coeffs, MLDSA_N, 0, MLDSA_Q);
+  mld_poly_decompose_32_native(a1->coeffs, a0->coeffs, a->coeffs);
+#else  /* !None && MLD_USE_NATIVE_POLY_DECOMPOSE_32 && (MLDSA_MODE == 3 || \
```
hanno-becker: Something is going wrong with `autogen` here, the first conditional does not seem to be recognized

mkannwischer: Fixed. Can you take another look please?

mldsa/native/aarch64/src/poly_decompose_32_asm.S
```diff
@@ -0,0 +1,114 @@
...
+
+        // Step 2: Barrett reduction with rounding: round(temp * 1025 / 2^22)
+        // This computes: round(ceil(a/128) / 4092)
+        // Combined: a1 ≈ round(ceil(a/128) / 4092) ≈ floor(a / 523776)
```
hanno-becker: Why does the second equality hold?

mldsa/native/aarch64/src/poly_decompose_32_asm.S
```diff
@@ -0,0 +1,114 @@
...
+        // sqrdmulh computes: (2 * temp * 524800 + 2^31) >> 32
+        // which is equivalent to: (temp * 1025 + 2^21) >> 22.
+        sqrdmulh \a1\().4s, \a1\().4s, barrett_const.4s
+
+        // Step 3: Mask to valid range [0, 14] since (Q-1)/(2*GAMMA2) = 15
```
hanno-becker: The comment would suggest that 15 is excluded after the masking, but it isn't, seeing that `mask_15` is elementwise`0x0F`?

mkannwischer: Yes, the comment is wrong. the valid range is [0, 15]. will fix.

mldsa/native/aarch64/src/poly_decompose_32_asm.S
```diff
@@ -0,0 +1,111 @@
...
+ */
+#include "../../../common.h"
+
+#if defined(MLD_ARITH_BACKEND_AARCH64) && !defined(MLD_CONFIG_MULTILEVEL_NO_SHARED)
+
+.macro decompose32 a1, a0, a, temp
+        // Compute a1 = round-(a / 523776) ≈ round(a * 1074791425 /
+        // 2^49), where round-() denotes "round half down". This is
+        // exact for 0 <= a < Q. Note that half is rounded down since
+        // 1074791425 / 2^49 ≲ 1 / 523776.
```
hanno-becker, line 14: I don't think half being rounded-down is related to this estimate? This instead seems to be about how you compute the Barret constant (using round down). Your estimate says why, despite the approximation in the Barret constant, you still get the exact `round-(a/N)`. Similar to Fact 2 in https://eprint.iacr.org/2022/439.pdf.

mldsa/native/aarch64/src/poly_decompose_32_asm.S
```diff
@@ -0,0 +1,111 @@
...
+        // Compute a0' = a - a1 * 2*GAMMA2. If no wrap-around
+        // happened, this is exactly a0. Otherwise this is a' + Q -
+        // a1 * 2*GAMMA2 = a0 + Q.
+        mls \a\().4s, \a1\().4s, gamma2_2x.4s
+
+        // If a0' > (Q-1)/2, which holds iff wrap-around happened,
+        // then compute a0 = a0' - Q.
+        cmgt \temp\().4s, \a\().4s, q_half.4s
+        and  \temp\().16b, \temp\().16b, q.16b
+        sub \a0\().4s, \a\().4s, \temp\().4s
```
hanno-becker: I don't understand the relevance of `(q-1)/2` here. Can we not unconditionally compute `a0' = a - a1 * (2*GAMMA2)` first, and then: a) If a1==16, this is a - (q-1), so we need to subtract 1 more to get the right `a0`. b) Finally, mask `a1`?

That would have shorter dependency chains? I probably miss something.

Actually, for the wraparound check one wouldn't need `a1`, but could use a `cmgt` on the input `a` itself -- which further reduces dependency chains.

hanno-becker: Something like this:
```asm
```asm
// a0 aliased with a0
.macro decompose32 a1, a, temp
        sqdmulh \a1\().4s, \a\().4s, barrett_const.4s
        srshr \a1\().4s, \a1\().4s, #18
        cmgt \temp\().4s, \a\().4s, q_bound.4s
        mls \a\().4s, \a1\().4s, gamma2_2x.4s
        mla \a\().4s, \temp\().4s, \temp\().4s
        // one could also do `add \a\().4s, \a\().4s, \temp\().4s, but the mls->mla chain may be faster
        and \a1\().16b, \a1\().16b, mask_15.16b
.endm
```
where `q_bound` is `15 * 523776 + (523776 / 2) = 8118528`

mldsa/native/aarch64/src/poly_decompose_32_asm.S
jammychiou1: > I don't think half being rounded-down is related to this estimate? This instead seems to be about how you compute the Barret constant (using round down). Your estimate says why, despite the approximation in the Barret constant, you still get the exact `round-(a/N)`. Similar to Fact 2 in https://eprint.iacr.org/2022/439.pdf.

When the denominator is odd (which Fact 2 also assumed), it doesn't matter if the approximation is ≲ or ≳. However 523776 is even, so it will impact the direction where 261888 = 523776 / 2 (for example) will round. (round(261888 * 1074791425 / 2^49) = round(261888 / 523776 - epsilon) = round(0.5 - epsilon) = 0) This is what I wanted to emphasize.

mldsa/native/aarch64/src/poly_decompose_32_asm.S
jammychiou1: Yeah, this approach which the reference implementation ([ref](https://github.com/pq-crystals/dilithium/blob/444cdcc84eb36b66fe27b3a2529ee48f6d8150c2/ref/rounding.c#L51-L52) and [avx2](https://github.com/pq-crystals/dilithium/blob/444cdcc84eb36b66fe27b3a2529ee48f6d8150c2/avx2/rounding.c#L78-L82)) took is pretty weird indeed. If there's no need to follow them closely, then your approach is definitely better. I am working on a new patch that will use your approach.

jammychiou1: > Something like this:
> 
> ```asm
> // a0 aliased with a0
> .macro decompose32 a1, a, temp
>         sqdmulh \a1\().4s, \a\().4s, barrett_const.4s
>         srshr \a1\().4s, \a1\().4s, #18
>         cmgt \temp\().4s, \a\().4s, q_bound.4s
>         mls \a\().4s, \a1\().4s, gamma2_2x.4s
>         mla \a\().4s, \temp\().4s, \temp\().4s
>         // one could also do `add \a\().4s, \a\().4s, \temp\().4s, but the mls->mla chain may be faster
>         and \a1\().16b, \a1\().16b, mask_15.16b
> .endm
> ```
> 
> where `q_bound` is `15 * 523776 + (523776 / 2) = 8118528`

Thanks for the snippet! One small thing: Since we want `a0 -= 1` if `temp = -1`, the mla should be mls, right? I will try both (mls vs add) and see.

hanno-becker: Yes, you're right, it should be `mls` :+1: But upon second thought, since MULs usually have lower throughput than basic arithmetic / logical operations, the plain `add` may be better.

mldsa/native/aarch64/src/poly_decompose_32_asm.S
hanno-becker: My point was that we're approximating `a / N = (a * (R / N)) / R ~ (a * [[R/N]]) / R` where `[[R/N]]` is an approximation of `R/N`. In our context, we use rounding _down_, so (as rational numbers) `(a * [[R/N]]) / R` is always strictly smaller than `a / N`. Hence, when the latter hits an integral multiple of 1/2 -- the case where there would be ambiguity about the meaning of `round(a/N)` -- the approximation `(a * [[R/N]]) / R` is just _below_ an integral multiple of `1/2`. This would be different if `[[R/N]]` was computed using rounding up (which one would consider if it just below an integer, but here `R/N` is just _above_ an integer)

mldsa/native/aarch64/src/poly_decompose_32_asm.S
hanno-becker: @mkannwischer @jammychiou1 We should consider implementing the same strategy in C?

mldsa/native/aarch64/src/poly_decompose_32_asm.S
jammychiou1: Pretty sure we're talking about the same thing. I added this comment because I originally had some difficulty understanding why `round-(a / N)` can be computed with rounding mulhi (either `sqrdmulh` or `sqdmulh` + `srshr`) which always round half up.

If the added comment itself is confusing, should I (1) reword it (maybe mention `1074791425 = floor(2^49 / 523776)`), (2) add even more details or (3) just leave it out?

mldsa/native/aarch64/src/poly_decompose_32_asm.S
mkannwischer: Thanks! Let's consider that in a follow-up.

hanno-becker: Ack

mldsa/native/aarch64/src/poly_decompose_32_asm.S
hanno-becker: I can't come up with a good description right now, I'd suggest to leave it for now and refine later.

Issue #158, opened 2025-04-13
Port magic-constant checker from mlkem-native


hanno-becker:
mlkem-native uses a script `scripts/check-magic` to look out for magic constants in the codebase and check that they have an 'explanation' attached to them in the form of an algebraic expression explaining their origin.

This script should be ported to mldsa-native and suitable explanations added.

7908837174:
Hi, I’m Kallal Mukherjee (GitHub: 7908837174). Thanks for reviewing my pull request (#389) which aimed to address constant-time execution for polynomial norm checking — a key mitigation for timing side-channel risks (ref: #153).

I’ve already submitted my LFX mentorship application for the mldsa-native project through the LFX platform and am hopeful to contribute further with security-focused improvements, reproducible tooling, and clear onboarding.

This PR included:

Manual implementation of constant-time logic for polyvecl_chknorm and polyveck_chknorm

Loop invariant annotations for CBMC verification

DCO sign-off and reviewer-ready documentation

I completely understand the contributor policy and am happy to follow any additional steps required. Let me know if any further info would help verify my background or contributions.

Thanks again!

PR #413, opened 2025-08-06
Improve UX of cbmc scripts, lint, autogen, format (align with mlkem-native)

### LINKED ISSUE #80: Port extensions to cbmc scripts from mlkem-native
https://github.com/pq-code-package/mlkem-native/pull/905

willieyz:
This PR port the following two PR from mlkem-native to here (mldsa-native):
- https://github.com/pq-code-package/mlkem-native/pull/1108
- https://github.com/pq-code-package/mlkem-native/pull/905
- Resolves #80 

> Note: We did not port the functional changes from commit 1f40536 due to the divergence between the two project repositories. However, if `synchronize_backends`, `gen_citations_for` or other functions are introduced in the future, we will consider incorporating those changes into mldsa-native.

### DIFF
--- a/scripts/tests
+++ b/scripts/tests
@@ -16,6 +16,7 @@ import time
 import logging
 import subprocess
 import json
+import re
 
 from enum import Enum
 from functools import reduce
@@ -696,13 +697,6 @@ class Tests:
                 log = logger(f"CBMC ({i+1}/{num_proofs})", scheme, None, None)
                 log.info(f"Starting CBMC proof for {func}")
                 start = time.time()
-                if self.args.verbose is False:
-                    extra_args = {
-                        "stdout": subprocess.DEVNULL,
-                        "stderr": subprocess.DEVNULL,
-                    }
-                else:
-                    extra_args = {}
                 try:
                     p = subprocess.run(
                         [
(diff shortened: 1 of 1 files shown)


scripts/format
mkannwischer, line 1: I see that the format script has diverged from mlkem-native.
Can we consolidate that by either fixing it in this PR, or opening an issue.
I see there is already #158 for the magic constant, but there are 3 more difference:
 - check-constracts
 - shell linting
 - copyright checking

scripts/tests
mkannwischer, line 1: I do see a diff in `run_cbmc_single_step` compared to mlkem-native that looks unintentional to me. Can you consolidate that please?

mkannwischer (changes requested):
Two comments.
In general this sort of PR is an ideal candidate to identify further gaps compared to mlkem-native. Please let me know what differences you have encountered, so we can open issue for those that should be resolved.

scripts/tests
mkannwischer: While at it - can you fix this line: 
  ```
                    log.error(
                        "Could not find function {self.args.start_with}. Running all proofs"
                    )
```
                    
log is not defined.

willieyz: Hello, Matthis, thank you for your reviewing!
Both issues you mentioned been fixed as follow:
- For the first part, I aligned the error handling in mldsa with the approach used in mlkem, which involves three main updates:
  - aligned the way `capture_output` is handled.
  - aligned the handling of `subprocess.TimeoutExpired`.
  - Added an additional checking before decoding `p.stderr`.
 
- For the second part, I created a corresponding log instance for the `run_cbmc`,to solve the not defined error, 
- And I throught that the same issue exists in the mlkem and need to be fixed. If my fix at here is acceptable, I will create a separate issue and apply the same fix in the mlkem-native repository.

scripts/format
willieyz: Hello, Matthias, 
Thank you for your reviewing, open issue:
- #421
to solve these gap.

PR #414, opened 2025-08-06
Add native implementation for `poly_caddq`

### LINKED ISSUE #327: AVX2: Add native implementation of `poly_caddq`

### LINKED ISSUE #396: AArch64: Consider adding `poly_caddq` assembly
With #371 merged, there may be more potential for speed-ups using assembly.

mkannwischer:
- Resolves #396 
- Resolves #327
- Depends on #411 
- Depends on #410


mldsa/native/x86_64/src/poly_caddq_avx2.c
```diff
@@ -0,0 +1,52 @@
...
+
+/*************************************************
+ * Name:        mld_poly_caddq_avx2
+ *
+ * Description: For all coefficients of in/out polynomial add Q if
+ *              coefficient is negative.
+ *
+ * Arguments:   - int32_t *r: pointer to input/output polynomial
+ **************************************************/
+void mld_poly_caddq_avx2(int32_t *r)
```
hanno-becker, line 28: Let's not introduce new intrinsics code. New backend functions should be written in ASM.

mldsa/native/aarch64/src/poly_caddq_asm.S
```diff
@@ -0,0 +1,59 @@
...
+ * Copyright (c) The mldsa-native project authors
+ * SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+ */
+#include "../../../common.h"
+
+#if defined(MLD_ARITH_BACKEND_AARCH64) && !defined(MLD_CONFIG_MULTILEVEL_NO_SHARED)
+
+.macro caddq inout
+        sshr tmp.4s, \inout\().4s, #31
+        mls \inout\().4s, tmp.4s, q_reg.4s
```
hanno-becker: Nit: `shr` + `mla` would be a bit more intuitive?

hanno-becker (changes requested):
It makes sense to wait for benchmarks first, but if we want this, the avx2 code should be written in ASM, not intrinsics.

mkannwischer:
> It makes sense to wait for benchmarks first, but if we want this, the avx2 code should be written in ASM, not intrinsics.

In that case we will have to split this up into two PRs. I'll do AArch64, but we will need to find someone who can write AVX2 asm. Wouldn't it make more sense to get performance on-par with the pqcrystals implementation first and then rewrite it to asm later or in parallel? The integration of the existing avx2 intrinsics implementation is straightforward.

mldsa/native/aarch64/src/poly_caddq_asm.S
mkannwischer: Yes, good point. Changed.

hanno-becker (approved):

mkannwischer:
I created the following issues as follow ups: 
- #490 
- #491 
- #492 
- #493

PR #415, opened 2025-08-07
ACVP: Download testvectors transparently on the fly, Ported from https://github.com/pq-code-package/mlkem-native/pull/1119

willieyz:
This PR port the following PR from mlkem-native to here (mldsa-native):
- https://github.com/pq-code-package/mlkem-native/pull/1119

Due to differences between the ACVP test structures of `mldsa-native` and `mlkem-native`, several modifications have been made:
- Replaced the `prompt` and `expectedResult` ACVP tests with a single `internalProjection` test.
- Removed the logic related to ACVT checking, as `mldsa-native` does not currently use ACVT testing.


mkannwischer (approved):
Thanks @willieyz. LGTM!

Issue #381, opened 2025-07-24
AArch64: Relicense or rewrite ntt.s and intt.s


mkannwischer:
[ntt.s](https://github.com/pq-code-package/mldsa-native/blob/main/mldsa/native/aarch64/src/ntt.S)  and [intt.s](https://github.com/pq-code-package/mldsa-native/blob/main/mldsa/native/aarch64/src/intt.S) are are still MIT-licensed as they are inherited from SLOTHY.
That breaks our uniform licensing.

We should either complete the relicensing in #236 (requiring constent by Arm) or we will have to reimplement the NTTs from scratch.

PR #424, opened 2025-08-11
Add copyright checking and shell lint to `lint`

### LINKED ISSUE #423: lint: Add shell script linting and copyright checking
Originally posted by @mkannwischer in https://github.com/pq-code-package/mldsa-native/pull/413#pullrequestreview-3096006399

- During the https://github.com/pq-code-package/mldsa-native/pull/413, we identified the following gaps compared to mlkem-native:
- add shell linting part in lint.
- add copyright checking part in lint.

willieyz:
- Resolves #423 
- This PR add the copyright checking and shell lint to `lint`, also align the color format and syntax with mlkem-native


hanno-becker (dismissed):
Thanks @willieyz! Something is not yet working correctly, because in fact not all source files have the `Apache-2.0 OR MIT OR ISC` license -- see e.g. https://github.com/pq-code-package/mldsa-native/blob/main/mldsa/native/aarch64/src/ntt.S. Somehow the script does not seem to catch this.

scripts/lint
```diff
@@ -71,52 +142,64 @@ check-eol-dry-run()
...
   for file in $(git ls-files -- "*.[chsS]" "*.py" ":/!proofs/cbmc/*.py"); do
     # Ignore symlinks
     if [[ ! -L $file && $(grep "Copyright (c) The mldsa-native project authors" $file | wc -l) == 0 ]]; then
-      echo "::error file=$file,line=${line:-1},title=Missing copyright header error::$file is missing copyright header"
+      gh_error "$file" "${line:-1}" "Missing copyright header error" "$file is missing copyright header"
+      success=false
+    fi
+  done
+  # For source files in dev/* and mldsa/*, we enforce `Apache-2.0 OR ISC OR MIT`
+  for file in $(git ls-files -- "*.[chsSi]" | grep "^dev/\|^mlkem/"); do
```
hanno-becker: This needs to be `mldsa`, not `mlkem`.

willieyz: Hello, @hanno-becker,
Thank you for your reviewing, and apologies for my careless mistake...
After correcting it to use `mldsa` instead of `mlkem`, the script correctly catch the `SPDX + Copyright headers` error,
see following link:
https://github.com/pq-code-package/mldsa-native/actions/runs/17578845391/job/49930280691?pr=424
I have a question, for these two file(`intt.S`, `ntt.S`), 
Should I update them to use Apache-2.0 OR ISC OR MIT as the script requires, instead of only MIT(same as `mlkem-native`), or should I modify the script so that it passes if only one of the licenses is present?

mkannwischer: You cannot just change licenses of code without asking the copyright owners - so that is not an option. 

Let's temporarily whitelist these two files and add a TODO in the script it needs to be removed once licenses have been sorted out as a part of #381.

scripts/lint
```diff
@@ -71,52 +158,69 @@ check-eol-dry-run()
...
+      success=false
+    fi
+  done
+  # For source files in dev/* and mldsa/*, we enforce `Apache-2.0 OR ISC OR MIT`
+  for file in $(git ls-files -- "*.[chsSi]" | grep "^dev/\|^mldsa/"); do
+    # TODO: Temporarily whitelist these files until PR #381 is resolved.
+    #       Once licenses are properly sorted out, remove this conditional, is_spdx_whitelisted() and SPDX_WHITELIST.
+    if is_spdx_whitelisted "$file"; then
+      continue
+    fi
```
mkannwischer: No need to make this overly complicated.
```suggestion
        # Temporarily exclude AArch64 (i)NTT pending license resolution (see issue #381)
        if [[ "$file" == "mldsa/native/aarch64/src/ntt.S" ]] || 
           [[ "$file" == "mldsa/native/aarch64/src/intt.S" ]]; then
            continue
        fi
```

willieyz: Thanks for your reviewing!
You are right, I had update it with your suggestion, thanks!

mkannwischer (changes requested):
@willieyz, please rebase this on top of main.

mkannwischer (approved):
Thanks @willieyz. LGTM now.

PR #436, opened 2025-09-03
Refactor squeeze/absorb to use mld_keccakf1600 primitives

bremoran:
Squeeze & Absorb have used


mldsa/fips202/keccakf1600.c
mkannwischer, line 1: Changes in this file seem unrelated to the PR description? Can we remove those?
Also, the code introduced here is dead code - so we should not add this until we actually have a backend that needs a native extract/xor.

mkannwischer (changes requested):
Thanks @bremoran. I support the refactoring of fips202.c.
The changes in keccakf1600.c seem not needed for the time being.

The CBMC proofs need to be adjusted (basically you need to add the newly called function to the `USE_FUNCTION_CONTRACT=` make variable in the respective failing proofs). I can take care of this after addressed my comment above.

mkannwischer (approved):
I am happy with these changes now. Before we merge the commit history needs to be cleaned up though.

@hanno-becker, could you also have a brief look - particularly the CBMC proof changes.

mldsa/fips202/fips202.c
```diff
@@ -266,13 +190,15 @@ static void keccak_absorb_once(uint64_t s[MLD_KECCAK_LANES],
...
-  requires(r <= sizeof(uint64_t) * MLD_KECCAK_LANES)
+  requires(r < sizeof(uint64_t) * MLD_KECCAK_LANES)
+  requires((r / 8) >= 1)
   requires(memory_no_alias(s, sizeof(uint64_t) * MLD_KECCAK_LANES))
   requires(memory_no_alias(in, inlen))
   assigns(memory_slice(s, sizeof(uint64_t) * MLD_KECCAK_LANES)))
 {
   unsigned int i;
 
+  /* i = keccak_absorb(s, 0, r, in, inlen); */
```
hanno-becker: Remove?

mldsa/fips202/fips202.c
```diff
@@ -174,8 +103,9 @@ __contract__(
   assigns(memory_slice(s, sizeof(uint64_t) * MLD_KECCAK_LANES))
 )
 {
-  s[pos / 8] ^= (uint64_t)p << 8 * (pos % 8);
-  s[r / 8 - 1] ^= 1ULL << 63;
```
hanno-becker, line 178: This was anyway endianness-specific? We should have caught this long ago.

hanno-becker: No, my fault.

mldsa/fips202/fips202.c
```diff
@@ -218,33 +148,27 @@ __contract__(
    * assigns() clauses. */
   size_t bytes_to_go = outlen;
 
-  while (bytes_to_go > 0)
+  if (pos == r)
+  {
+    mld_keccakf1600_permute(s);
+    pos = 0;
+  }
```
hanno-becker: This is a functional change, isn't it? We now permute when running with `pos == r` and `outlen == 0`, while previously, this would have been a no-op.

bremoran: You're correct. I will fix.

bremoran: There's also the point that the tests didn't catch this => it needs a test as well.

mldsa/fips202/fips202.c
```diff
@@ -199,15 +120,15 @@ static unsigned int keccak_squeeze(uint8_t *out, size_t outlen,
                                    unsigned int pos, unsigned int r)
 __contract__(
   requires((r == SHAKE128_RATE && pos <= SHAKE128_RATE) ||
-    (r == SHAKE256_RATE && pos <= SHAKE256_RATE))
+           (r == SHAKE256_RATE && pos <= SHAKE256_RATE) ||
```
hanno-becker, line 123: Why do we restrict `r` in the first place?

mldsa/fips202/fips202.c
hanno-becker: This is not a correctness issue, but one of consistency I think. Previously, it seems that `pos == r` was allowed and would lead to permutation and `pos = 0` reset only upon the next squeeze. With the change -- and it is still the case after your latest change -- the squeeze logic instead permutes and resets to `pos = 0` immediately when the block is full. Either is OK, but I think it should be consistent with `finalize` which does not permute and reset.

If possible, I would like to keep this uniform and in line with mlkem-native. Can you have a think?

hanno-becker (changes requested):
I support the change per-se, and the speedup is great, thank you @bremoran.

There is one minor issue of consistency, though, with regards to when to permute and reset `pos = 0` during squeezing. Previously, we would do this lazily only when we need to squeeze more. Now, we keep `pos = r` at finalize, but `squeeze` eagerly permutes and resets to `pos = 0`. If we can uphold the previous approach (in line with mlkem-native) while still riping the performance benefits, I'd prefer that.

mldsa/fips202/fips202.c
```diff
@@ -218,33 +139,27 @@ __contract__(
    * assigns() clauses. */
   size_t bytes_to_go = outlen;
 
-  while (bytes_to_go > 0)
+  if ((outlen > 0) && (pos >= r))
+  {
+    mld_keccakf1600_permute(s);
+    pos = 0;
+  }
```
hanno-becker: Would this case not be handled correctly by the first loop iteration already? (`xor_bytes` being a no-op)

mldsa/fips202/fips202.c
```diff
@@ -218,33 +139,27 @@ __contract__(
...
+
+  while (pos + bytes_to_go >= r)
```
hanno-becker: Perhaps the previous behavior can be retained simply by using `>` here.

mldsa/fips202/fips202.c
bremoran: Yes, I will issue a PR to mlkem-native as well, since it is also using an assignment loop instead of an accelerated memset.

bremoran:
@hanno-becker hopefully this change is more what you had in mind. The cbmc proofs do not pass, but I'm not sure why.

PR #437, opened 2025-09-04
Split files into level-specific and level-generic

### LINKED ISSUE #154: Split files into level-specific and level-generic
If we want to support a multi-level build as in mlkem-native, files need to be split into level-specific and level-generic functionality. In a multi-level build, the level-generic files will then be compiled only once.

jakemas:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/154

Following mlkem-native, I used `poly_k.c` as the name of the level-dependent files. While ML-DSA uses two parameters `K` and `L`, I think the consistency with mlkem is a good balance.

I can build addition elements into this PR, or we can split into a series.


hanno-becker (commented):
What was the logic behind the splitting? I was expecting functions in `poly_k.c` to be about `polyveck` or `polyvecl`, not `poly`, but that's not the case.

hanno-becker (changes requested):
What was the logic behind the splitting? I was expecting functions in poly_k.c to be about polyveck or polyvecl, not poly, but that's not the case.

PR #450, opened 2025-09-10
Avoid calling keccak_absorb with partial lanes

bremoran:


mkannwischer:
@bremoran, sorry for the long wait for the review on this. Could you please rebase this on top of the changes in main, so we can benchmark and review it?

mkannwischer:
@bremoran, that was not quite what I meant by rebasing.
I applied the changes required to make this work myself in https://github.com/pq-code-package/mldsa-native/pull/450/commits/a8d2d6a8b0efde1e37923bb4d8373a1645d12b6c.

rod-chapman:
I see one proof failure in mld_H. Let me take a look...

hanno-becker (changes requested):
Thanks @bremoran! I can definitely see this being useful for 32-bit platforms.

A few requests:
* I don't think this needs an API extension: Instead, the buffering of state prior to XOR'ing should be an implementation detail (add a buffer for the incomplete lane) of the existing absorb API.
* We should have documentation and CBMC proofs for new functionality.
* The new logic belongs to FIPS-202.

Could you adjust the PR accordingly?

mkannwischer:
> Thanks @bremoran! I can definitely see this being useful for 32-bit platforms.
> 
> A few requests:
> 
> * I don't think this needs an API extension: Instead, the buffering of state prior to XOR'ing should be an implementation detail (add a buffer for the incomplete lane) of the existing absorb API.
> * We should have documentation and CBMC proofs for new functionality.
> * The new logic belongs to FIPS-202.
> 
> Could you adjust the PR accordingly?

I agree. Marking this as draft for now. @bremoran, please mark it as ready when you have updated the PR.
Let me know if you need help with adjusting the CBMC proofs.

Issue #455, opened 2025-09-15
CBMC: Make buffer bounds explicit


hanno-becker:
(Sibling of https://github.com/pq-code-package/mlkem-native/issues/1189)

CBMC has an implicit internal bound on the maximum size of objects (~2^53 in our setup). In consequence, if e.g.  `memory_no_alias(buf, sz)` is assumed, it is implicitly assumed that `sz` does not exceed this object size bound.

While CBMC's bound should be practically sufficient, it is confusing to the user that the bound is not made explicit in the contract. For example, in In particular, this applies to https://github.com/pq-code-package/mldsa-native/blob/main/mldsa/fips202/fips202.c#L113-126 the addition `inlen + pos` cannot overflow because of the implicit buffer bound, but this is not visible from the in-source annotations.

CBMC offers the option --malloc-may-fail --malloc-fail-assert to instead assert upon use of memory_no_alias, the provided size does not exceed the maximum object size. This forces us to explicitly state upper bounds on object length (where they are not static anyway).

**Task:** The Makefile.common should be modified to always apply --malloc-may-fail --malloc-fail-assert, and the CBMC contracts adjusted if necessary. In particular, `keccak_absorb` and its call-site(s) will then need explicit message bounds in the contracts, and those should be mentioned in the top-level API documentation. 

Note that even though the above parameters are documented for cbmc, it seems they need to be applied to goto-instrument, so best to amend both CBMCFLAGS and GOTO_INSTRUMENT.

PR #456, opened 2025-09-15
FIPS 202: avoid potential overflow in addition

nmouha:
To avoid a potential overflow in ```pos + inlen```, I'm proposing to rewrite:

```
while (pos + inlen >= r)
```

as

```
while (inlen >= r - pos)
```

Note that there cannot be an underflow in the subtraction, as ```pos <= r``` is a precondition.

### DIFF
--- a/mldsa/fips202/fips202.c
+++ b/mldsa/fips202/fips202.c
@@ -123,7 +123,7 @@ __contract__(
 {
   unsigned int i;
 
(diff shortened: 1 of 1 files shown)


hanno-becker (commented):
Thank you @nmouha for noting this and proposing a fix.

The change looks like an clear improvement to me, independently of whether we also address #455.

I have opened #458 to run the full CI.

Issue #439, opened 2025-09-08
nix: Introduce separate shells for cross compilation


mkannwischer:
We should port https://github.com/pq-code-package/mlkem-native/pull/1104

PR #457, opened 2025-09-15
Nix: Introduce separate shells for cross compilation

willieyz:
- Resolves: #439 
- Previously, we used a single `ci-cross` nix shell for all cross-compilation tool chains. This shell takes a very long time to build and is therefore inconvenient to use for cross-compilation on a specific target.
- This commit extends the nix flake by architecture specific ci cross shells:
  - ci-cross-x86_64
  - ci-cross-aarch64
  - ci-cross-riscv64
  - ci-cross-riscv32
  - ci-cross-ppc64le
  - ci-cross-aarch64_be
- It also adjusts CI to use those shells.
- This commit is port from mlkem-native: https://github.com/pq-code-package/mlkem-native/pull/1104


mkannwischer (approved):
LGTM. Thanks @willieyz!

PR #463, opened 2025-09-17
Add magic-constant checker ported from mlkem-native

willieyz:
- This PR resolves: #158 
- This PR is rebased on the following branch because they are related `lint` change:
  - # 451
- Currently, this commit is a work in progress. It only ports the magic-constants checker from `mlkem-native` to `mldsa-native`. Since the two projects do not share the same magic-constant list, we will need to fix them one by one in the following work.


mldsa/api.h
```diff
@@ -8,9 +8,12 @@
 #include <stddef.h>
 #include <stdint.h>
 
+/* Sizes of cryptographic material, per parameter set */
```
mkannwischer: Why add this comment? Remove.

willieyz: Hello Matthias, thank you for your review!
The reason I originally added it was that I noticed a similar description in mlkem_native.h of mlkem-native for the following parameter definitions.
But for now, I have removed this one like you suggestion. 

https://github.com/pq-code-package/mlkem-native/blob/91f19272797e9960ff1e3e41292ea0ac765a6dbe/mlkem/mlkem_native.h#L72-L77

mkannwischer: ah okay. There is another issue for aligning these two files - so let's handle that separately.

willieyz: got it, thanks~

PR #464, opened 2025-09-18
Add config param set and namespace prefix

### LINKED ISSUE #446: Config: Add MLD_CONFIG_PARAMETER_SET
We should follow mlkem-native in the configuration options for the parameter set and namespace prefix.

jakemas:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/446

Added in minimal framework for multi level build to support new config param mlkem-approach


test/break_pct_config.h
```diff
@@ -7,10 +7,53 @@
...
+ * Description: Specifies the parameter set for ML-DSA
+ *              - MLD_CONFIG_PARAMETER_SET=44 corresponds to ML-DSA-44
+ *              - MLD_CONFIG_PARAMETER_SET=65 corresponds to ML-DSA-65
+ *              - MLD_CONFIG_PARAMETER_SET=87 corresponds to ML-DSA-87
+ *
+ *              This can also be set using CFLAGS.
+ *
+ *****************************************************************************/
+#ifndef MLD_CONFIG_PARAMETER_SET
+/* Map legacy MLDSA_MODE to new parameter set for backward compatibility */
```
hanno-becker, line 22: We can break backwards compatibility here, so let's choose `MLD_CONFIG_PARAMETER_SET` as the single source of truth and not allow any other configuration.

hanno-becker: Can `MLDSA_MODE` be removed entirely?

hanno-becker:
> The change appears to affect run time of ML-DSA-44 CBMC proof to 28m:

@jakemas It's already like this on `main`, so not related to your PR.

hanno-becker (changes requested):
@jakemas Can you clean up the history and add short yet expressive commit messages?

Also, I don't think we need to worry about backwards compatibility at this point. If you introduce `MLD_CONFIG_PARAMETER_SET` as the new source of truth for the parameter set, that should supersede all others. Can `MLDSA_MODE` be removed?

test/break_pct_config.h
mkannwischer: Agree - MLDSA_MODE should be removed entirely.

mkannwischer:
Marking this as draft. Please mark it as ready for review when you think this should be merged.

PR #469, opened 2025-09-22
CBMC: Increase CBMC_OBJECT_BITS to improve proof runtime

mkannwischer:
This improves CBMC proof run-time for ML-DSA-44 from 25 minutes to 5 minutes, for ML-DSA-65 from 8 minutes to 6 minutes, and for ML-DSA-87 from 8 minutes to 7 minutes.


hanno-becker (approved):
I'm surprised that CBMC gets faster with a _bigger_ value of `CBMC_OBJECT_BITS`.

PR #477, opened 2025-09-24
CBMC: Introduce explicit upper bounds

willieyz:
- Resolves: #455 
- This PR is similar with (https://github.com/pq-code-package/mlkem-native/pull/1191), introduce the explicit upper bounds on the maximum size of objects (~2^53).

- This PR add the explicit upper bounds for following:
  - `crypto_sign`
  - `crypto_sign_open`
  - `crypto_sign_signature`


mldsa/sign.h
```diff
@@ -102,6 +102,8 @@ int crypto_sign_signature_internal(uint8_t *sig, size_t *siglen,
                                    const uint8_t rnd[MLDSA_RNDBYTES],
                                    const uint8_t *sk, int externalmu)
 __contract__(
+  requires(mlen <= MLD_MAX_BUFFER_SIZE)
+  requires(prelen <= MLD_MAX_BUFFER_SIZE)
   requires(memory_no_alias(sig, CRYPTO_BYTES))
   requires(memory_no_alias(siglen, sizeof(size_t)))
   requires(memory_no_alias(m, mlen))
```
rod-chapman, line 112: A question possibly unrelated to this PR... if crypto_sign_signature_internal() is an internal function, why is it declared here in sign.h, rather than being local and static in sign.c?

mldsa/sign.h
```diff
@@ -238,6 +243,9 @@ int crypto_sign_verify_internal(const uint8_t *sig, size_t siglen,
                                 const uint8_t *pre, size_t prelen,
                                 const uint8_t *pk, int externalmu)
 __contract__(
```
rod-chapman, line 252: Same question. If this function is "internal", why is it declared in sign.h?

rod-chapman (commented):
2 comments. Otherwise good. All 140 proofs reproduce OK on an r7g.16xlarge instance with MLDSA_MODE=5 for me. Tests all OK. Lint OK.

mldsa/sign.h
mkannwischer: They are called internal in FIPS204, but they need to be part of the API because they are needed  in both AWS-LC and liboqs (e.g. for ACVP testing).
I would prefer the _derand name as in ML-KEM, but that would be deviating from the spec.

mkannwischer: See https://github.com/pq-code-package/mldsa-native/issues/354

rod-chapman: Understood

mldsa/cbmc.h
```diff
@@ -60,6 +61,18 @@
...
+/* Maximum supported buffer size
+ *
+ * Larger buffers may be supported, but due to internal modeling constraints
+ * in CBMC, the proofs of memory- and type-safety won't be able to run.
+ *
+ * If you find yourself in need for a buffer size larger than this,
+ * please contact the maintainers, so we can prioritize work to relax
+ * this somewhat artificial bound.
+ */
+#define MLD_MAX_BUFFER_SIZE (SIZE_MAX >> 12)
```
rod-chapman, line 73: One obvious comment - on a 32-bit system, MLD_MAX_BUFFER_SIZE will be 2^20 == 1 Megabyte. Is that going to be a problem for any users?

hanno-becker: Yes. We will need to think about how to document or (better) address the limitation.

mkannwischer (approved):

PR #479, opened 2025-09-25
Makefile: Remove use of .WAIT primitive to support make <4.4

### LINKED ISSUE #427: Incompatible .WAIT syntax in Makefile for Make < 4.4
### Description 
The current Makefile uses .WAIT syntax in benchmark targets, which is incompatible with GNU Make versions prior to 4.4. This prevents building these targets on systems with older but still commonly used Make versions.

Steps to Verify

Ensure you have Make 4.3 or earlier installed
Run make --version to confirm version < 4.4
Attempt to build benchmark targets:

`make run_bench CYCLES=NO`
 Expected Behaviour:
`make: *** No rule to make target '.WAIT', needed by 'run_bench'. Stop.`

Image

### Root Cause

The current MakeFile uses this syntax
```
# Use .WAIT to prevent parallel execution when -j is passed
run_bench: \
    run_bench_44 .WAIT\
    run_bench_65 .WAIT\
    run_bench_87
```

GNU Make versions prior to 4.4 treat .WAIT as a regular target name rather than a special target, causing the build to fail when it can't find a rule to make .WAIT.

### Affected Code

Lines 129-133: run_bench target
Lines 150-154: run_bench_components target

Image

### Solutions:

1. Add Version Detection: Check Make version and use different syntax accordingly

```
# Detect Make version
MAKE_VERSION := $(shell $(MAKE) --version | head -n1 | sed 's/.*Make \([0-9]\+\.[0-9]\+\).*/\1/')
MAKE_MAJOR := $(shell echo $(MAKE_VERSION) | cut -d. -f1)
MAKE_MINOR := $(shell echo $(MAKE_VERSION) | cut -d. -f2)
```

```
# Check if Make >= 4.4
MAKE_SUPPORTS_WAIT := $(shell [ $(MAKE_MAJOR) -gt 4 ] || ([ $(MAKE_MAJOR) -eq 4 ] && [ $(MAKE_MINOR) -ge 4 ]) && echo yes)

ifeq ($(MAKE_SUPPORTS_WAIT),yes)
# Use .WAIT syntax for Make 4.4+
run_bench: \
    run_bench_44 .WAIT\
    run_bench_65 .WAIT\
    run_bench_87
else
# Use fallback for older Make versions
.NOTPARALLEL: run_bench
run_bench: run_bench_44 run_bench_65 run_bench_87
endif
```

2. Document Minimum Version: Update documentation to require Make 4.4+

I'd recommend implementing Solution: backward compatibility with fallback  combined with documentation

hanno-becker:
Previously, the Makefile used .WAIT to order the execution of the benchmarks, breaking compatibility with versions of make prior to 4.4.

This commit fixes this by serializing the benchmark execution explicitly, without relying on .WAIT.

Fixes #427

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@
 	build test all \
 	clean quickcheck check-defined-CYCLES \
 	size_44 size_65 size_87 size \
-	run_size_44 run_size_65 run_size_87 run_size 
+	run_size_44 run_size_65 run_size_87 run_size
 
 SHELL := /bin/bash
 .DEFAULT_GOAL := build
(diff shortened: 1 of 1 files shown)


mkannwischer (approved):
LGTM

Issue #40, opened 2025-02-22
ACVP: Consider adding "internal" test cases


mkannwischer:
Currently the ACVP tests only cover `signatureInterface=external`. This should be extended to cover the internal test cases as well.

PR #496, opened 2025-09-28
Change API of signature_internal and verify_internal to match FIPS204

mkannwischer:
According to FIPS204, the internal functions take a formated message as input
defined as
BytesToBits(IntegerToBytes(0, 1) | IntegerToBytes(|ctxlen|, 1) | ctx) | msg

However, our current API takes ctx and message separately. This makes it
impossible to pass the ACVP tests which only contain one field for
the formates message (even worse: those tests apparently do not follow the
encoding above making it impossible to parse it and split it into ctx and msg.

This commit changes the APIs to match FIPS204, and adds the ACVP tests for
the internal functions. It also adjusts the CBMC proofs accordingly.

- Resolves https://github.com/pq-code-package/mldsa-native/issues/40
- Depends on #495


mkannwischer:
@jakemas, @hanno-becker, @rod-chapman: I need some feed back on whether or not this change is acceptable for the AWS-LC integration. I don't really like this change, but I don't see another way to pass ACVP tests.

mkannwischer:
If we decide to not change the previous API, a workaround would be to pass an empty pre and a formatted message as m. That would solve the problem for the ACVP tests too. Maybe that is a better option for us.

test/acvp_mldsa.c
```diff
@@ -108,6 +116,52 @@ static int decode_hex(const char *prefix, unsigned char *out, size_t out_len,
   return 1;
 }
 
+
+static int decode_int(const char *prefix, int *out, const char *str)
```
hanno-becker: Nit: The naming is perhaps confusing because a) it doesn't mention the prefix in the function name, and b) when I hear 'prefix decode' I think of something like `0x` or `0b` prefixes.

Perhaps one should say `decode_keyed_int` here or something.

test/acvp_mldsa.c
```diff
@@ -263,6 +354,54 @@ int main(int argc, char *argv[])
...
+      unsigned char sk[CRYPTO_SECRETKEYBYTES];
+      int externalMu;
+      size_t mlen;
+
+      /* Parse message */
+      if (argc == 0)
+      {
+        goto siggen_internal_usage;
+      }
+      mlen = (strlen(*argv) - strlen("message=")) / 2;
```
hanno-becker, line 371: Nit: This can underflow

hanno-becker (approved):
Thank you @mkannwischer for the explanation and consideration.

I agree with your assessment. Following the FIPS204 internal API to the letter would require dynamic allocation, which we have avoided so far and should not introduce for this point alone. Also, I note that one can argue that this is an implementation detail as we represent a logical string as a concatenation of two separate arguments (this would be different if, e.g. the prefix preparation was done inside `xxx_internal` too).

Finally, we're still working towards our first release, so everything can still change if I learn more going forward.

test/acvp_mldsa.c
mkannwischer: That should be fine. Then decode_hex will fail because mlen does not message the length of the decoded message.

test/acvp_mldsa.c
mkannwischer: Adjusted.

PR #498, opened 2025-09-28
Add support for PreHash ML-DSA 

mkannwischer:
This commit adds two functions crypto_sign_signature_pre_hash
and crypto_sign_verify_pre_hash implementing the pre-hashing mode of ML-DSA.
Instead of receiving a message, they receive a pre-hashed message.
Details can be found in Algorithm 4 and 5 in FIPS204.

The message to signed is formatted as
0x01 || ctxlen (1 byte) || ctx || oid || ph
where ph is the pre-hash and oid is the object identifier of the used hash
algorithm.

The ACVP client is adjusted to support the pre-hashing test cases. Note that
the ACVP testvectors only contain the message, not the pre-hash. I opted for
computing the hash in the ACVP Python client as for that we do not have to
add implementations for the 12 hash functions.

CBMC proofs for the new functions are added.

- Resolves https://github.com/pq-code-package/mldsa-native/issues/39
- Depends on #496


mkannwischer:
Something that is not yet clear to me is if the pre hashing step should be included in mldsa-native or if it should be performed by the consumer.

If one strictly follows the NIST algorithms one should include it.

The reason for not including it was that the use case of PreHash ML-DSA is that the message is too large for the signer/verifier to process it internally.
But of course it could be that only one of signer/verifier cannot handle large message - for implementing the other party it can then be useful to have the regular signing API and perform prehashing internally.

I am not sure what we want here. The prospect of including 12 hash functions is not nice.

mkannwischer:
> Something that is not yet clear to me is if the pre hashing step should be included in mldsa-native or if it should be performed by the consumer.
> 
> If one strictly follows the NIST algorithms one should include it.
> 
> The reason for not including it was that the use case of PreHash ML-DSA is that the message is too large for the signer/verifier to process it internally. But of course it could be that only one of signer/verifier cannot handle large message - for implementing the other party it can then be useful to have the regular signing API and perform prehashing internally.
> 
> I am not sure what we want here. The prospect of including 12 hash functions is not nice.

I looked into how other crypto libraries are implementing this and it indeed seems to include the hashing internally. This PR will have to change and I will have to add a bunch of hash implementations. Probably going to take SHA-2 from https://github.com/pq-code-package/slhdsa-c.

jakemas:
> I looked into how other crypto libraries are implementing this and it indeed seems to include the hashing internally. 
NIST provide some more details on this https://csrc.nist.gov/csrc/media/Projects/post-quantum-cryptography/documents/faq/fips204-sec6-03192025.pdf

They do mention HashML-DSA: 
>For example, as shown in the diagram below, module A may implement ML-DSA.Sign(sk, M, ctx) or
HashML-DSA.Sign(sk, M, ctx, PH). Module A’s implementation of either function may call MLDSA.Sign_mu(sk, mu), implemented on module B, which may then generate a random value rnd
and call ML-DSA.Sign_internal_mu(sk, mu, rnd). To perform a signature, module A would construct
M’, compute tr from pk, compute mu from tr and M’, then call module B via ML-DSA.Sign_mu(sk,
mu) to obtain the signature. Module A would be the module that computes mu, and module B
would be the signing module. In this case sk may not be the private key but a reference to it if it is
already available on module B. If module B only contains a single private key, sk may be omitted
from the interface.

mkannwischer:
@jakemas @hanno-becker @rod-chapman
This is how it looks if we want to support all 12 hash functions for PreHash ML-DSA.
If we want to go down the path of having a local SHA-2 implementation, we will have to proof memory-safety of it. I tried this briefly, but I got stuck in proofing the compression function (it spins forever). It may be quite some work.
I currently don't see another way to pass all the ACVP testvectors.

Any better ideas?

mkannwischer:
Marking this as ready for review because I really need some feedback to move forward with this.

mkannwischer:
After discussion with @hanno-becker - we will go back to the approach of having the pre-hash passed from externally. We will only add the internal pre-hashing for SHAKE256 - this way we don't have to include all the hash functions.
I put the current version supporting all versions here: https://github.com/pq-code-package/mldsa-native/tree/acvp-prehash-all for furture reference

mldsa/sign.c
```diff
@@ -862,3 +862,242 @@ int crypto_sign_open(uint8_t *m, size_t *mlen, const uint8_t *sm, size_t smlen,
...
+ *              to oid and checks that the given hash length matches
+ *              the given hash function.
+ *
+ * Arguments:   - uint8_t oid[11]: pointer to output oid
+ *              - mld_hash_alg_t hashAlg: hash algorithm enumeration
+ *              - size_t len: Hash length to be checked
+ *
+ * Returns 0 if hash algorithm is known and the hash length matches
+ * and -1 otherwise.
+ **************************************************/
```
hanno-becker: Not a blocker, but I find it surprising that this function does two seemingly disconnected things: Find-and-compare the output length of a hash function, and provide an OID. It'd seem cleaner to me to split this in two functions.

mldsa/sign.c
```diff
@@ -862,3 +862,242 @@ int crypto_sign_open(uint8_t *m, size_t *mlen, const uint8_t *sm, size_t smlen,
...
+                                    0x65, 0x03, 0x04, 0x02, 0x0A};
+  const uint8_t shake_128_oid[11] = {0x06, 0x09, 0x60, 0x86, 0x48, 0x01,
+                                     0x65, 0x03, 0x04, 0x02, 0x0B};
+  const uint8_t shake_256_oid[11] = {0x06, 0x09, 0x60, 0x86, 0x48, 0x01,
+                                     0x65, 0x03, 0x04, 0x02, 0x0C};
+
+  switch (hashAlg)
+  {
+    case MLD_SHA2_224:
+      if (len == 28)
```
hanno-becker: We should introduce `#define`d constants here, or otherwise rewrite the comparisons as `len == 224 / 8` etc. to make it more obvious where this number comes from.

mldsa/sign.c
```diff
@@ -862,3 +862,242 @@ int crypto_sign_open(uint8_t *m, size_t *mlen, const uint8_t *sm, size_t smlen,
...
+  }
+  fmsg[0] = 1;
+  fmsg[1] = ctxlen;
+  if (ctx != NULL && ctxlen != 0)
+  {
+    mld_memcpy(fmsg + 2, ctx, ctxlen);
+  }
+  if (sha_oid(fmsg + 2 + ctxlen, hashAlg, phlen))
+  {
+    *siglen = 0;
```
hanno-becker, line 889: It would be good hygiene to not write anything to `fmsg` if the function fails; or, to fall through to the cleanup phase where `fmsg` is zeroized.

mldsa/sign.c
```diff
@@ -862,3 +862,242 @@ int crypto_sign_open(uint8_t *m, size_t *mlen, const uint8_t *sm, size_t smlen,
...
+  int result;
+
+  if (ctxlen > 255)
+  {
+    *siglen = 0;
+    return -1;
+  }
+  fmsg[0] = 1;
+  fmsg[1] = ctxlen;
+  if (ctx != NULL && ctxlen != 0)
```
hanno-becker: Shorten to `ctxlen != 0`?

mldsa/sign.c
```diff
@@ -862,3 +862,242 @@ int crypto_sign_open(uint8_t *m, size_t *mlen, const uint8_t *sm, size_t smlen,
...
+  {
+    mld_memcpy(fmsg + 2, ctx, ctxlen);
+  }
+  if (sha_oid(fmsg + 2 + ctxlen, hashAlg, phlen))
+  {
+    return -1;
+  }
+  mld_memcpy(fmsg + 2 + ctxlen + 11, ph, phlen);
```
hanno-becker: The logic for preparing the formatted message should be hoisted out into a helper which can be used both by sign and verify.

mldsa/sign.h
```diff
@@ -383,4 +410,176 @@ __contract__(
...
 );
 
+/*************************************************
+ * Name:        crypto_sign_signature_pre_hash_internal
+ *
+ * Description: FIPS 204: Algorithm 4 HashML-DSA.Sign.
+ *              Computes signature with pre-hashed message.
+ *
+ * Arguments:   - uint8_t *sig: pointer to output signature (of length
+ *                              CRYPTO_BYTES)
```
hanno-becker, line 420: Probably not specific to this PR, but why don't we use `uint8_t sig[CRYPTO_BYTES]` below?

mldsa/sign.h
```diff
@@ -383,4 +410,176 @@ __contract__(
...
+ *
+ * The supported hash functions are: "SHA2-224", "SHA2-256", "SHA2-384",
+ *                                   "SHA2-512", "SHA2-512/224", "SHA2-512/256",
+ *                                   "SHA3-224", "SHA3-256", "SHA3-384",
+ *                                   "SHA3-512", "SHAKE-128", "SHAKE-256"
+ *
+ * Warning: This is an unstable API that may change in the future. If you need
+ * a stable API use crypto_sign_signature_pre_hash_shake256.
+ *
+ * Returns 0 (success) or -1 (context string too long OR nonce exhaustion)
```
hanno-becker: Or invalid `phlen`

mldsa/sign.h
```diff
@@ -383,4 +410,176 @@ __contract__(
...
+  requires(memory_no_alias(siglen, sizeof(size_t)))
+  requires(memory_no_alias(ph, phlen))
+  requires(ctxlen == 0 || memory_no_alias(ctx, ctxlen))
+  requires(memory_no_alias(rnd, MLDSA_RNDBYTES))
+  requires(memory_no_alias(sk, CRYPTO_SECRETKEYBYTES))
+  requires(hashAlg >= MLD_SHA2_224 && hashAlg <= MLD_SHAKE_256)
+  assigns(memory_slice(sig, CRYPTO_BYTES))
+  assigns(object_whole(siglen))
+  ensures((return_value == 0 && *siglen == CRYPTO_BYTES) ||
+          (return_value == -1 && *siglen == 0))
```
hanno-becker: General remark: For top-level API, the preconditions should be as weak as possible, to give the user the strongest possible safety guarantees even in the face of invalid input. Of course, one can't get past assuming memory validity, but something like `requires(hashAlg >= MLD_SHA2_224 && hashAlg <= MLD_SHAKE_256)` could be avoided and instead be replaced by a runtime check.

mldsa/sign.h
```diff
@@ -383,4 +410,176 @@ __contract__(
...
+                                         mld_hash_alg_t hashAlg)
+__contract__(
+  requires(phlen <= 64)
+  requires(ctxlen <= MLD_MAX_BUFFER_SIZE - 77)
+  requires(siglen <= MLD_MAX_BUFFER_SIZE)
+  requires(memory_no_alias(sig, siglen))
+  requires(memory_no_alias(ph, phlen))
+  requires(ctxlen == 0 || memory_no_alias(ctx, ctxlen))
+  requires(memory_no_alias(pk, CRYPTO_PUBLICKEYBYTES))
+  requires(hashAlg >= MLD_SHA2_224 && hashAlg <= MLD_SHAKE_256)
```
hanno-becker: Similar comment as above

hanno-becker (changes requested):
Thanks a lot @mkannwischer! A few smaller comments on code sharing and the preconditions on the global API.

mldsa/sign.c
mkannwischer: Yes, good point. I split it into two functions and also moved it to a separate file called prehash.c/h to not pollute the sign.c too much.

mldsa/sign.c
mkannwischer: changed to 224/8 - I hope that is obvious enough.

mldsa/sign.c
mkannwischer: done - added `mld_format_pre_hash_message`

mldsa/sign.h
mkannwischer: Agreed. I opened https://github.com/pq-code-package/mldsa-native/issues/554 for this.

mldsa/sign.h
mkannwischer: Yes, updated.

mldsa/sign.h
mkannwischer: Yes, this makes sense. I removed the pre-condition on hashAlg and changed the pre-condition for phlen to be <= MLD_MAX_BUFFER_SIZE.

This is actually a very good point for discussion. For example, we have 
```c
requires(ctxlen <= MLD_MAX_BUFFER_SIZE)
```

This isn't an actual pre-condition. You may very well call this function with any ctxlen and it will just fail the run-time check immediately. However, if we remove this precondition, `memory_no_alias(ctx, ctxlen)` fails. 
Any ideas?

mldsa/sign.h
mkannwischer: fixed.

mldsa/sign.c
mkannwischer: Done.

mldsa/sign.c
mkannwischer: Resolved by moving the formatting of the message to a standalone function.

mldsa/prehash.c
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * Description: Returns the OID of a given SHA-2/SHA-3 hash function.
+ *
+ * Arguments:   - uint8_t oid[11]: pointer to output oid
+ *              - mld_hash_alg_t hashAlg: hash algorithm enumeration
+ *
+ **************************************************/
+static void mld_get_hash_oid(uint8_t oid[11], mld_hash_alg_t hashAlg)
+{
+  const uint8_t sha2_224_oid[11] = {0x06, 0x09, 0x60, 0x86, 0x48, 0x01,
```
hanno-becker: Nit: A fair amount of repetition. One could consider something like
```c
#define MLD_HASH_OID_LEN 11
void mld_get_hash_oid2(uint8_t oid[11], mld_hash_alg_t hashAlg)
{
  static const struct {
    mld_hash_alg_t alg;
    uint8_t oid[MLD_HASH_OID_LEN];
  } oid_map[] = {
    {MLD_SHA2_224, {0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x04}},
    {MLD_SHA2_256, {0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x01}},
    {MLD_SHA2_384, {0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x02}},
    {MLD_SHA2_512, {0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x03}},
    {MLD_SHA2_512_224, {0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x05}},
    {MLD_SHA2_512_256, {0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x06}},
    {MLD_SHA3_224, {0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x07}},
    {MLD_SHA3_256, {0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x08}},
    {MLD_SHA3_384, {0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x09}},
    {MLD_SHA3_512, {0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x0A}},
    {MLD_SHAKE_128, {0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x0B}},
    {MLD_SHAKE_256, {0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x0C}}
  };

  for (size_t i = 0; i < sizeof(oid_map) / sizeof(oid_map[0]); i++) {
    if (oid_map[i].alg == hashAlg) {
      mld_memcpy(oid, oid_map[i].oid, MLD_HASH_OID_LEN);
      return;
    }
  }
}
```
instead, but I'm OK as it is for now.

mkannwischer: Changed to your suggestion.

PR #501, opened 2025-09-30
Add native implementation of poly_chknorm

### LINKED ISSUE #401: AVX2: Consider adding `poly_chknorm` assembly
When #392 gets merged merged, there may be more potential for speed-ups using assembly.

### LINKED ISSUE #402: AArch64: Consider adding `poly_chknorm` assembly
When #392 gets merged merged, there may be more potential for speed-ups using assembly.

jammychiou1:
- Resolves #401
- Resolves #402


mldsa/native/api.h
```diff
@@ -254,4 +254,20 @@ static MLD_INLINE void mld_poly_use_hint_88_native(int32_t *b, const int32_t *a,
                                                    const int32_t *h);
 #endif /* MLD_USE_NATIVE_POLY_USE_HINT_88 */
 
+#if defined(MLD_USE_NATIVE_POLY_CHKNORM)
+/*************************************************
+ * Name:        mld_poly_chknorm_native
+ *
+ * Description: Check infinity norm of polynomial against given bound.
+ *              Assumes input coefficients were reduced by mld_reduce32().
```
mkannwischer, line 262: Unless I am missing something here this documentation is wrong - this code (just like the C code for it) only makes sense if inputs are canonically reduced. What must be happening here is that we are lucky and for the possible inputs to reduce32() it always return canonical results. I will open https://github.com/pq-code-package/mldsa-native/issues/504 to follow up on this.

mkannwischer:
I opened issues for follow-up verification/SLOTHY work for this function:
 - #505 
 - #506 
 - #507 
 - #508

mldsa/poly.c
```diff
@@ -304,10 +304,17 @@ void mld_poly_use_hint(mld_poly *b, const mld_poly *a, const mld_poly *h)
...
  * coefficient itself must remain secret).
  * We instead perform everything in constant-time.
+ * Also it is sufficient to check that it is smaller than
+ * MLDSA_Q - REDUCE32_RANGE_MAX > (MLDSA_Q - 1) / 8).
  */
 MLD_INTERNAL_API
 uint32_t mld_poly_chknorm(const mld_poly *a, int32_t B)
 {
+#if defined(MLD_USE_NATIVE_POLY_CHKNORM)
+  /* TODO: proof */
```
rod-chapman, line 330: Do we need to add a proof dir for "poly_chknorm_native" below proofs/cbmc to exercise this?

mldsa/poly.c
```diff
@@ -320,6 +327,16 @@ uint32_t mld_poly_chknorm(const mld_poly *a, int32_t B)
     invariant((t == 0) == array_abs_bound(a->coeffs, 0, i, B))
   )
   {
+    /*
+     * Since we know that a is in [-REDUCE32_RANGE_MAX+1, REDUCE32_RANGE_MAX-1],
```
rod-chapman: What does the [A, B] notation mean? Inclusive bounds? If so, I don't think you need the +1 on the lower bound?

mldsa/poly.c
```diff
@@ -320,6 +327,16 @@ uint32_t mld_poly_chknorm(const mld_poly *a, int32_t B)
...
+     * and B <= MLDSA_Q - REDUCE32_RANGE_MAX, to check if a mod+- MLDSA_Q is
+     * in [-B+1, B-1], it suffices to check if a is in [-B+1, B-1].
+     * In other words: If a is not in [-B+1, B-1], a +- k*MLDSA_Q for k >= 1 is
+     * also not in [-B+1, B-1].
+     */
+    mld_assert(a->coeffs[i] < B || a->coeffs[i] - MLDSA_Q <= -B);
```
rod-chapman, line 354: It might be worth adding a comment here that you're trying to write
`
a ==> b
`
but need to write
`
!a || b
`
to make it executable, right?

rod-chapman (commented):
See comments in individual files so far. I have _not_ reviewed either asm implementation in any detail.

mldsa/poly.c
mkannwischer: Yes, we have an issue for that: https://github.com/pq-code-package/mldsa-native/issues/350

The proofs itself are straightforward, but what needs to happen before that is that for each function we need to add comments to the asm documenting the bound propagation and derive the suitable contract that matches alll backends. For poly_chknorm this is trivial, but for the NTT needs a bit of thinking.

We left comments in the code for all other native functions as well so we do not forget it.

mldsa/poly.c
mkannwischer: Yes, you are right. I eliminated the [ ] notation and used explicit <= and < to make this both correct and easier to follow.

mldsa/poly.c
mkannwischer: Yes, the CBMC assertions also work with `==>`, but it would work for the run-time `assert`s in the DEBUG builds. I added a comment.

mldsa/poly.c
```diff
@@ -320,6 +327,19 @@ uint32_t mld_poly_chknorm(const mld_poly *a, int32_t B)
...
+    /*
+     * Since we know that -REDUCE32_RANGE_MAX <= a < REDUCE32_RANGE_MAX,
+     * and B <= MLDSA_Q - REDUCE32_RANGE_MAX, to check if
+     * -B < (a mod± MLDSA_Q) < B, it suffices to check if -B < a < B.
+     * In other words: If -B < a < B, then for k >= 1, we also have
+     * -B < (a ± k*MLDSA_Q) < B.
+     *
+     * We prove this to be true using the following CBMC assertions.
+     * a ==> b expressed as !a || b to also allow run-time assertion.
+     */
```
hanno-becker, line 353: This comment should be moved into a `Specification:` section as it's something one needs to know when validating that the spec in the code is equivalent to the spec in FIPS-204.

I would also emphasize the main point:
```
The definition of this FIPS-204 requires signed canonical reduction prior 
to applying the bounds check. However, `-B < (a mod± MLDSA_Q) < B`
is equivalent to `-B < a < B` under the assumption that 
`B <= MLDSA_Q - REDUCE32_RANGE_MAX` (cf. the assertion in below).
Hence, the present spec and implementation are correct without reduction.
```

mldsa/poly.h
```diff
@@ -307,14 +307,14 @@ __contract__(
  * Arguments:   - const mld_poly *a: pointer to polynomial
  *              - int32_t B: norm bound
  *
- * Returns 0 if norm is strictly smaller than B <= (MLDSA_Q-1)/8 and 0xFFFFFFFF
- * otherwise.
+ * Returns 0 if norm is strictly smaller than
+ * B <= (MLDSA_Q - REDUCE32_RANGE_MAX) and 0xFFFFFFFF otherwise.
```
hanno-becker, line 311: Nit
```suggestion
 * Returns 0 if the infinity norm is strictly smaller than B, and 0xFFFFFFFF otherwise.
 *
 * B must not be larger than MLDSA_Q - REDUCE32_RANGE_MAX.
```
And then also add the comment below regarding (un)necessity of modular reduction in the infinity norm.

mldsa/poly.c
mkannwischer: Okay. I moved it to the header and adjusted it as suggested.

mldsa/poly.h
mkannwischer: changed.

rod-chapman:
I am happy with the C code and updates to explanatory comments.
CT tests are still failing by the look of it.
I have not reviewed the assembly language.

mkannwischer:
> I am happy with the C code and updates to explanatory comments. CT tests are still failing by the look of it. I have not reviewed the assembly language.

I aborted them earlier to speed-up CI for other PRs. Restarted them now, but everything was fine before my comment changes, so it should still be fine.

hanno-becker (approved):

PR #511, opened 2025-10-01
Add native implementations for Keccak on Armv8.1-m

bremoran:
Also add tests for Keccak implementations on Armv8.1-m


mldsa/fips202/keccakf1600.c
```diff
@@ -40,7 +40,10 @@ void mld_keccakf1600_extract_bytes(uint64_t *state, unsigned char *data,
                                    unsigned offset, unsigned length)
 {
   unsigned i;
-#if defined(MLD_SYS_LITTLE_ENDIAN)
+#if defined(MLD_USE_FIPS202_X1_NATIVE)
+  (void) i;
```
mkannwischer: remove this and move the unsigned i inside of the other block instead.

Makefile
```diff
@@ -2,14 +2,14 @@
 # Copyright (c) The mldsa-native project authors
 # SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
 
-.PHONY: func kat acvp stack \
+.PHONY: func kat acvp stack keccak \
```
mkannwischer: Can we use the same naming as in mlkem-native and call this unit: https://github.com/pq-code-package/mlkem-native/blob/9ae2223e835c1abcc8d6857dd3b7f8ce00a05216/Makefile#L84? 

If we can keep the diff to mlkem-native minimal that would be appreciated.

mkannwischer (changes requested):
Thanks @bremoran! We'll need to discuss this with the others. Here are some initial comments.

I don't see this test being exercised in the MVE CI so far. That definitely should be changed.

We should align the test names with mlkem-native. 

My biggest concern right now is uniform licensing. At least everything in `mldsa/` needs to be `Apache-2.0 OR ISC OR MIT`. For `test/` we may be slightly more flexible, but just having `Apache-2.0 OR ISC OR MIT` for everything makes it less confusing. For code that is public domain/CC0 relicensing should be easy.

Makefile
bremoran: Yes, I will attempt to set this up more like the unit test framework in mlkem-native.

mkannwischer (changes requested):
Thanks @bremoran! I'm very sorry for the long silence on this PR. 
Could you please split this up into 3 smaller PRs: 
(1) Add benchmarking for Armv8-M
(2) Add the unit tests for Keccak
(3) Add the native MVE Keccak

PR #523, opened 2025-10-02
Add MLD_CONFIG_PARAM_SET

jakemas:


mkannwischer (changes requested):
Thanks @jakemas! Let's get rid of `mode` entirely. Rest looks good to me.

.github/workflows/cbmc.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       kattest: false
       acvptest: false
       cbmc: true
-      cbmc_mldsa_mode: 2
+      cbmc_mldsa_mode: 44
```
mkannwischer: Let's rename all occurences of mode to paramater_set, so there can be no confusion?

jakemas: Thank you! Okay, I hit those last ones!

mkannwischer: Thanks!

mkannwischer:
There was still a problem with the cbmc proofs. I took the liberty to fix that: https://github.com/pq-code-package/mldsa-native/compare/95079508e99a49ce7fe190b192f3b3d8929b687a..2e6e16578b72cecc37b86cd92493821da3e0750a

Issue #527, opened 2025-10-04
Remove `MLD_NAMESPACETOP`


mkannwischer:
We inherited `MLD_NAMESPACETOP` from the reference implementation, but it does not serve much purpose anymore if I see correctly. Let's remove it. See https://github.com/pq-code-package/mldsa-native/pull/526.

PR #532, opened 2025-10-04
Add MLD_CONFIG_MULTILEVEL_WITH_SHARED and NO_SHARED

jakemas:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/154
- Resolves https://github.com/pq-code-package/mldsa-native/issues/527
- Resolves https://github.com/pq-code-package/mldsa-native/issues/534

TODO/FOLLOWUP:
- complete MLB and add examples/multilevel_build like mlkem-native's https://github.com/pq-code-package/mlkem-native/tree/main/examples/multilevel_build (maybe 2 part this PR to help with size/staleness)


jakemas:
Considering splitting in 2 parts. So far this PR lays down a good foundation for the new config. Hanno suggested first populating `examples/` with the basic case (_ala_ mlkem-native) so we have some infrastructure to test plumbing and configuration during the multi-level-build build-out. So I've pulled https://github.com/pq-code-package/mldsa-native/pull/533 to do exactly that. Happy to hear thoughts on how much to add to this PR before a merge.

mldsa/config.h
```diff
@@ -43,13 +43,83 @@
...
+ *              need only a single parameter set, keep this unset.
+ *
+ *              If this is set, all MLD_CONFIG_PARAMETER_SET-independent
+ *              code will be included in the build, including code needed only
+ *              for other parameter sets.
+ *
+ *              Example: mld_shake256_init is shared across all parameter sets.
+ *              If this option is set for a build with MLD_CONFIG_PARAMETER_SET
+ *              == 65, it would be included even though it's also needed for
+ *              levels 44 and 87.
```
mkannwischer: This is a bad example as `mld_shake256_init` is also needed for 65.
A better example would be `mld_poly_uniform_eta_4x` - we will need to versions of that: one for eta=2 and one for eta=4. 65 only need the eta=4 version, but the shared build would include both. But i am not sure if you have implemented that yet.

mldsa/config.h
```diff
@@ -43,13 +43,83 @@
...
+ *              be #undef'ined or not at the end of the SCU file. This is needed
+ *              in multilevel builds.
+ *
+ *              See examples/multilevel_build_mldsa_native for an example.
+ *
+ *              This can also be set using CFLAGS.
+ *
+ *****************************************************************************/
+/* #define MLD_CONFIG_MONOBUILD_KEEP_SHARED_HEADERS */
+
```
mkannwischer: I don't think it makes sense to include this one yet before we actually have a SCU build.

mldsa/poly.h
mkannwischer, line 1: We should not have level specific functions in poly.c/h. Do you plan to resolve this in a follow-up?

mkannwischer (changes requested):
I don't see `MLD_CONFIG_MULTILEVEL_WITH_SHARED` / `MLD_CONFIG_MULTILEVEL_NO_SHARED` having any effect yet. 
For example `ntt.c` should be guarded - `poly.c` should also be guarded, but that is more work. 
I'm fine with merging this in smaller steps, but ntt.c definitely need to be guarded and a follow-up issue created for guarding `poly.c`.

mldsa/poly.h
jakemas: Ok, I will implement NTT split first, then get some input from you. The poly split you have in mind is similar to the split suggested in https://github.com/pq-code-package/mldsa-native/pull/437 correct? a `poly_kl.c` file for level specific

mldsa/config.h
jakemas: Ok -- removed!

mldsa/config.h
jakemas: aha yup, silly mistake! Thank you -- fixed

mldsa/poly.h
jakemas: I've implemented what I think you mean (and based on mlkem-native) in [455881b](https://github.com/pq-code-package/mldsa-native/pull/532/commits/455881bb81207f6b424d27e8f83a8c90fa0c4ec7)

mldsa/poly_kl.c
```diff
@@ -0,0 +1,790 @@
...
+    if (t_valid) /* t0 < 9 */
+    {
+      a[ctr++] = 4 - (int32_t)t0;
+    }
+    t_valid = t1 < 9; /* t1 < 9 */
+    MLD_CT_TESTING_DECLASSIFY(&t_valid, sizeof(int));
+    if (t_valid && ctr < target)
+    {
+      a[ctr++] = 4 - (int32_t)t1;
+    }
```
mkannwischer, line 310: The downside of this approach is that in a multi-level build you will have 3 copies of this function (44, 65, 87), while actually you would only need two (eta=2 and eta=4). It would be better to split up this function and move them back to poly.c. We can do that in a folllow-up. 
Same applies to other functions in this file.

mldsa/polyvec.c
```diff
@@ -16,6 +16,9 @@
 #include <string.h>
 
 #include "common.h"
+
+#if !defined(MLD_CONFIG_MULTILEVEL_NO_SHARED)
```
mkannwischer: I don't see how this makes sense? All the functions in this file are dependent on the K/L parameters, so in multi-level build you cannot get away with building this file only once.
This guard should be removed.

mldsa/config.h
```diff
@@ -43,13 +43,69 @@
...
+ *              If this is set, all MLD_CONFIG_PARAMETER_SET-independent
+ *              code will be included in the build, including code needed only
+ *              for other parameter sets.
+ *
+ *              Example: mld_poly_uniform_eta_4x needs two versions - one for
+ *              eta=2 (used by parameter sets 44 and 87) and one for eta=4
+ *              (used by parameter set 65). If this option is set for a build
+ *              with MLD_CONFIG_PARAMETER_SET == 65, both versions would be
+ *              included even though parameter set 65 only needs the eta=4
+ *              version.
```
mkannwischer: This is not what is currently implemented, is it?

mkannwischer (changes requested):
Thanks @jakemas! Getting there.
polyvec.c should not be guarded by `MLD_CONFIG_MULTILEVEL_NO_SHARED` - we actually need to build that one 3 times.
The comment in config.h still does not make sense.

Please create a follow-up issue to clean up poly_kl.c - that leads to a lot of duplicate code right now.

mldsa/config.h
jakemas: ok addressed again

mldsa/polyvec.c
jakemas: done

mldsa/config.h
mkannwischer: It still does not make sense because there simply is no example for this yet in the code. I have opened https://github.com/pq-code-package/mldsa-native/issues/536 as a follow-up. There we can adjust it.

mkannwischer (approved):
It remains to be seen if this PR actually works, but the changes look good to me. Let's get this merged and move forward in small steps. 

I created a follow-up issue https://github.com/pq-code-package/mldsa-native/issues/536.

Issue #173, opened 2025-04-18
Add simple example for the use of mldsa-native as code package


hanno-becker:
Following https://github.com/pq-code-package/mlkem-native/tree/main/examples/mlkem_native_as_code_package, add a minimal example showing how to use mldsa-native as a code package.

PR #533, opened 2025-10-06
Add examples/basic

jakemas:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/173


examples/basic/main.c
```diff
@@ -0,0 +1,1021 @@
...
+      {
+        printf(", ");
+        if ((i + 1) % 12 == 0)
+        {
+          printf("\n    ");
+        }
+      }
+    }
+  }
+  printf("\n};\n");
```
mkannwischer: This pollutes the CI output somehow. Can we maybe only print the first 32 or 64 bytes or? Also no need to print this as a C array.

examples/basic/README.md
```diff
@@ -0,0 +1,40 @@
...
+## Usage
+
+Build this example with `make build`, run with `make run`.
+
+## What this example demonstrates
+
+This basic example shows how to use the ML-DSA (Module-Lattice-Based Digital Signature Algorithm) for:
+
+1. **Key Generation**: Generate a public/private key pair
+2. **Digital Signing**: Sign a message with a private key and optional context
```
mkannwischer:
```suggestion
2. **Signing**: Sign a message with a private key and optional context
```

examples/basic/main.c
```diff
@@ -0,0 +1,1021 @@
...
+   * Normally, you would want to seed a PRNG with trustworthy entropy here. */
+  randombytes_reset();
+
+  printf("ML-DSA-%d Basic Example\n", MLD_CONFIG_PARAMETER_SET);
+  printf("======================\n\n");
+
+  /* Initialize test message and context */
+  memcpy(m, test_msg, msglen);
+  memcpy(ctx, "test_context_123", ctxlen);
+
```
mkannwischer: Why do we need to copy this around?

examples/basic/main.c
```diff
@@ -0,0 +1,1021 @@
...
+  const char *test_msg =
+      "This is a test message for ML-DSA digital signature algorithm!";
+  size_t msglen = strlen(test_msg);
+
+
+  /* The PCT modifies the PRNG state, so the KAT tests don't work.
+   * We run KAT tests only for disabled PCT. */
+#if !defined(MLD_CONFIG_KEYGEN_PCT)
+#if MLD_CONFIG_PARAMETER_SET == 44
+  const uint8_t expected_signature[] = {
```
mkannwischer: These long arrays are arther distracting given that the main purpose of the examples are to demonstrate how mldsa-native can be used? Should we move them to a separate file that is included? Or should we compare a hash of te expected signature.

mkannwischer (changes requested):
Thanks @jakemas! Looking fairly good. A couple of nits.

examples/basic/main.c
jakemas: Thanks, now removed. I'll compare to the hash of the signatures, as you suggest

examples/basic/README.md
jakemas: thanks!

examples/basic/main.c
jakemas: Thank you, simplified this now.

examples/basic/main.c
jakemas: Good point, I did think the same, but thought it may be nice for users to see their full signatures etc. but I agree with you, and have just chosen to compare to the hash now

examples/basic/main.c
```diff
@@ -0,0 +1,155 @@
...
+#include <string.h>
+
+/* Import public mldsa-native API
+ *
+ * This requires specifying the parameter set and namespace prefix
+ * used for the build.
+ */
+#include "../../mldsa/common.h"
+#include "../../mldsa/fips202/fips202.h"
+#include "../../mldsa/sign.h"
```
mkannwischer: A consumer should not be including these headers. There should be a mldsa_native.h and that should be all that is included. I think we can live with including `sign.h` for now until https://github.com/pq-code-package/mldsa-native/issues/354 is resolved. 

common.h is probably not needed. 
fips202.h can definitely not be included. These are not public APIs.

jakemas: I needed fips202.h to hash the signature... what do you suggest?

mkannwischer: The only thing I can think of is adding another SHA-3 implementation, but that is ugly. Sorry, using the hash was not a good idea - my bad. Maybe we switch back to comparing the actual signature, but include it from another file?

Other suggestions are welcome.

jakemas: I like keeping the full signature, as I imagine a customer use-case being swaping out the example message and getting to generate full signatures on their own messages. I'll keep the full signature check and include the extra file.

mkannwischer (approved):
Thanks for the adjustments @jakemas. LGTM!

PR #541, opened 2025-10-17
Port `simpasm` from mlkem-native to mldsa-native

### LINKED ISSUE #471: Port simpasm script from mlkem-native
See https://github.com/pq-code-package/mlkem-native/blob/main/scripts/simpasm
See https://github.com/pq-code-package/mlkem-native/pull/717

This will be a prerequisite for #425.

willieyz:
- Resolves #471 
- Reference from https://github.com/pq-code-package/mlkem-native/pull/717
- Reference from https://github.com/pq-code-package/mlkem-native/blob/main/scripts/simpasm
- The purpose of this PR is to port the `simpasm` script from `mlkem-native` to `mldsa-naive`. Currently, there are some differences between the two repositories, so the `simpasm` script and the related CI jobs need to be adjusted. All `aarch64` and `x86_64` assembly code has been simplified on their respective systems.
- We have also modified some assembly code in `mldsa-native` to make it compatible with our `simpasm` scripts. The detailed description is as follows:
  - Since the CFI have not been ported to `mldsa-native` yet, we have temporarily removed the `--cfify` argument in the `autogen` scripts and added a comment:  
   `# TODO: Temporarily remove the "--cfify"; add back when the CFI script is implemented.`  
   Once the CFI scripts are ported (see issue #425), this will be restored.

  - For the assembly file `poly_caddq_asm.S`, there is an `.balign 16` directive. We have changed this `.balign` to match our standard setting in `simpasm` (see [autogen_header](https://github.com/pq-code-package/mldsa-native/blob/093e624c6eeb9e3c2ed8121f34cb26c6501e7e90/scripts/simpasm#L259-L269)), where `.balign` is set to `4`, and simplified it in a separate commit.

  - For the assembly files `rej_uniform_eta2_asm.S` and `rej_uniform_eta4_asm.S`, there are two consecutive labels:  
    ```
    rej_uniform_eta2_loop8_end:
    rej_uniform_eta2_memory_copy:
    ```
    ```
    rej_uniform_eta4_loop8_end:
    rej_uniform_eta4_memory_copy:
    ```
    The `llvm-objdump` discards one of the consecutive labels, which causes the `autogen` script to fail. We removed the end label since it is not used anywhere, and also simplified this two in a separate commit.


.github/workflows/base.yml
```diff
@@ -218,6 +218,33 @@ jobs:
...
           make lib
+  simpasm:
+    strategy:
+      fail-fast: false
+      matrix:
+        backend:
+           - arg: '--aarch64-clean'
+             name: Clean
+           - arg: ''
+             name: Optimized
```
mkannwischer: What is the purpose of this? We don't have an optimized/clean seperation so far?

dev/aarch64_clean/meta.h
mkannwischer, line 1: Please change the second commit so that git can track this as a file move so that we keep history for these files. I thought I had mentioned this before.

dev/aarch64_clean/src/poly_caddq_asm.S
mkannwischer, line 1: I don't understand why the third commit is removing files from `dev/aarch64_clean/src`

mkannwischer (changes requested):
Thanks @willieyz! I think the changes look okay - I still need to check the autogen script in detail.

The commit structure, however, is currently a large mess.
Most importantly, please make sure that git can track the move from src/ to dev/. That should be done in a separate commit. Then make all the necessary changes to the assembly in dev/ in a separate commit each. Then finally autogenerate the assemby in a single commit.

.github/workflows/base.yml
willieyz: Yes, you are right about this, I originally thought that we would have optimized/clean eventually, so I added this.
Now I remove this and add a TODO comment instead, thanks for your review!

dev/aarch64_clean/meta.h
willieyz: Hello, @mkannwischer, sorry about this mistake, 
Now I re-organized all file structure, do it by move instead of copy, this is my fault, 
Thank you for your review!

dev/aarch64_clean/src/poly_caddq_asm.S
willieyz: This one also fixed after I re-organized the structure from copy to move,
thanks you for your review!

.github/workflows/ci.yml
```diff
@@ -461,7 +461,7 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
       - uses: ./.github/actions/setup-shell
         with:
-          nix-shell: 'ci'
+          nix-shell: 'ci-cross' # Need cross-compiler for ASM simplification
```
mkannwischer, line 464: Did you forget to copy the `nix-cache: 'true'` here? Or was that intentionaly omited?

scripts/autogen
```diff
@@ -1244,6 +1244,251 @@ class BibliographyEntry:
...
+def update_via_remove(filename, dry_run=False):
+    if dry_run is True:
+        print(
+            f"Autogenerated file {filename} needs removing. Have you called scripts/autogen?",
+            file=sys.stderr,
+        )
+        exit(1)
+
+    # Remove the file
+    os.remove(filename)
```
mkannwischer: Can we stick to the same order of functions as in mlkem-native, to minimize the diff. 
In mlkem-native `update_via_copy` and `update_via_remove` come much later in the file.

scripts/autogen
```diff
@@ -1244,6 +1244,251 @@ class BibliographyEntry:
...
+            raise Exception(f"Could not find cross toolchain {cross_prefix}")
+    else:
+        cross_prefix = None
+
+    with tempfile.NamedTemporaryFile(suffix=".S") as tmp:
+        try:
+            # Determine architecture from filename
+            arch = "aarch64" if "aarch64" in infile_full else "x86_64"
+
+            # TODO: Temporary remvoe the "--cfify", add back when CFI script added.
```
mkannwischer, line 1292: Can you just leave it in there but comment it out? That will make it easier to re-add it in the correct place later.

.github/workflows/base.yml
```diff
@@ -218,6 +218,28 @@ jobs:
       - name: make lib
         run: |
           make lib
+  simpasm:
+    strategy:
+      fail-fast: false
+      matrix: # TODO: add backend option after we have optimized/clean seperation
```
mkannwischer: Can you please keep it in the file but comment it out in that case?

scripts/autogen
```diff
@@ -1489,11 +1737,29 @@ def _main():
     gen_avx2_zeta_file(args.dry_run)
     gen_avx2_rej_uniform_table(args.dry_run)
     high_level_status("Generated zeta and lookup tables")
+
+    synchronize_backends(
+        dry_run=args.dry_run,
+        clean=args.aarch64_clean,
+        force_cross=args.force_cross,
+        no_simplify=args.no_simplify,
+    )
```
mkannwischer, line 1747: Why are the arguments ordered differently here from mlkem-native? Please try to minimize the diff.

scripts/autogen
```diff
@@ -1489,11 +1737,29 @@ def _main():
...
+
+    high_level_status("Synchronized backends")
```
mkannwischer, line 1748: There is another diff here because of spacing issues.

.github/workflows/base.yml
mkannwischer:
```yml
      matrix: 
        backend:
           - arg: '--aarch64-clean'
             name: Clean
        # TODO: add backend option after we have optimized/clean seperation
        #   - arg: ''
        #     name: Optimized
```

Then you can also change the default for `--aarch64-clean` back to False.

mkannwischer (changes requested):
Thanks @willieyz! The commit structure is now much better.
Here are a couple of more comments primarily to align things with mlkem-native - otherwise maintainance of this is going to be unmanagable over time.

scripts/autogen
willieyz: Yes, I had fixed this, thanks for your review.

.github/workflows/base.yml
willieyz: Yes, I’ve added back the backend option and commented it out.
Regarding the default value for `--aarch64-clean`, I think we should keep it false, since we only have one version called aarch64_clean for now?
I had test with the default value of `--aarch64-clean` to true, the script would fail since we don’t have the `aarch64_opt` folder.

.github/workflows/ci.yml
willieyz: Hello Matthias,
In the previous push, I intentionally skipped this option because I thought I shouldn’t enable nix-cache when running several different tests, as I assumed each test should use a separate Nix setup. 
However, after you pointed this out, I realized that with this option, the CI can save a lot of time (from 2 hours to 2 minutes) for the check_autogenerated_files step. Also, for auto-generated files, there is no difference whether we use nix-cache or not. 
So yes, I should definitely add it.

Issue #543, opened 2025-10-21
Add `-Wconversion` and `-Wsign-conversion` warnings (and resolve findings)


mkannwischer:
- https://github.com/pq-code-package/mlkem-native/pull/1230
- https://github.com/pq-code-package/mlkem-native/pull/1238

PR #552, opened 2025-10-24
Resolve -Wconversion and -Wsign-conversion warnings

mkannwischer:
This commit enables additional warnings (-Wconversion, -Wsign-conversions) requiring more explicit casts.
All of those casts should already be safe as checked by CBMC.

- Resolves https://github.com/pq-code-package/mldsa-native/issues/543

### DIFF
--- a/test/baremetal/platform/m55-an547/platform.mk
+++ b/test/baremetal/platform/m55-an547/platform.mk
@@ -58,4 +58,7 @@ LDFLAGS += \
 
 # Extra sources to be included in test binaries
 EXTRA_SOURCES = $(wildcard $(M55_AN547_PATH)/*.c)
+# The CMSIS files fail compilation if conversion warnings are enabled
+EXTRA_SOURCES_CFLAGS = -Wno-conversion -Wno-sign-conversion
+
 EXEC_WRAPPER := $(realpath $(PLATFORM_PATH)/exec_wrapper.py)
--- a/test/mk/components.mk
+++ b/test/mk/components.mk
@@ -70,3 +70,11 @@ $(foreach scheme,mldsa44 mldsa65 mldsa87, \
 $(ALL_TESTS:%=$(MLDSA44_DIR)/bin/%44): $(call MAKE_OBJS, $(MLDSA44_DIR), $(wildcard test/notrandombytes/*.c) $(EXTRA_SOURCES))
 $(ALL_TESTS:%=$(MLDSA65_DIR)/bin/%65): $(call MAKE_OBJS, $(MLDSA65_DIR), $(wildcard test/notrandombytes/*.c) $(EXTRA_SOURCES))
 $(ALL_TESTS:%=$(MLDSA87_DIR)/bin/%87): $(call MAKE_OBJS, $(MLDSA87_DIR), $(wildcard test/notrandombytes/*.c) $(EXTRA_SOURCES))
+
+# Apply EXTRA_CFLAGS to EXTRA_SOURCES object files
+ifneq ($(EXTRA_SOURCES),)
+$(call MAKE_OBJS, $(MLDSA44_DIR), $(EXTRA_SOURCES)): CFLAGS += $(EXTRA_SOURCES_CFLAGS)
+$(call MAKE_OBJS, $(MLDSA65_DIR), $(EXTRA_SOURCES)): CFLAGS += $(EXTRA_SOURCES_CFLAGS)
+$(call MAKE_OBJS, $(MLDSA87_DIR), $(EXTRA_SOURCES)): CFLAGS += $(EXTRA_SOURCES_CFLAGS)
+endif
+


mldsa/fips202/fips202.c
```diff
@@ -94,9 +94,9 @@ __contract__(
     mld_keccakf1600_permute(s);
     pos = 0;
   }
-  mld_keccakf1600_xor_bytes(s, in, pos, inlen);
+  mld_keccakf1600_xor_bytes(s, in, pos, (unsigned)inlen);
```
hanno-becker, line 98: Similar to `// Safety: ...` comments in Rust, any not entirely obvious conversion should have a type-safety comment. Here, one would need to argue that while `inlen` could be larger than `unsigned` initially, by the time we get here it's `< r`.

mldsa/fips202/fips202x4.c
```diff
@@ -55,17 +55,17 @@ __contract__(
 
   if (inlen > 0)
   {
-    mld_keccakf1600x4_xor_bytes(s, in0, in1, in2, in3, 0, inlen);
+    mld_keccakf1600x4_xor_bytes(s, in0, in1, in2, in3, 0, (unsigned)inlen);
```
hanno-becker, line 60: Similar to above, though a single comment may be enough refering to all those cases.

mldsa/native/aarch64/meta.h
```diff
@@ -77,7 +77,7 @@ static MLD_INLINE int mld_rej_uniform_eta2_native(int32_t *r, unsigned len,
   MLD_CT_TESTING_DECLASSIFY(buf, buflen);
   outlen =
       (int)mld_rej_uniform_eta2_asm(r, buf, buflen, mld_rej_uniform_eta_table);
-  MLD_CT_TESTING_SECRET(r, sizeof(int32_t) * outlen);
+  MLD_CT_TESTING_SECRET(r, sizeof(int32_t) * (unsigned)outlen);
```
hanno-becker: Can this be addressed by declaring `outlen` as `unsigned` instead and casting it to `int` in the end? When that happens, a comment should be added why it is safe (we should have such for L79 already, where the conversion currently happens).

mldsa/native/x86_64/src/rej_uniform_avx2.c
```diff
@@ -111,7 +111,7 @@ unsigned int mld_rej_uniform_avx2(
 
     if (t < MLDSA_Q)
     {
-      r[ctr++] = t;
+      r[ctr++] = (int32_t)t;
```
hanno-becker, line 115: Borderline because it's obvious, but: Consider adding `// Cast is safe because t < MLDSA_Q`

mldsa/poly_kl.c
```diff
@@ -78,7 +78,7 @@ unsigned int mld_poly_make_hint(mld_poly *h, const mld_poly *a0,
   )
   {
     const unsigned int hint_bit = mld_make_hint(a0->coeffs[i], a1->coeffs[i]);
```
hanno-becker, line 80: Why don't we change the return type of `mld_make_hint` to suit the usage?

mldsa/poly_kl.c
```diff
@@ -663,38 +664,38 @@ void mld_polyz_pack(uint8_t *r, const mld_poly *a)
...
-    t[0] = MLDSA_GAMMA1 - a->coeffs[2 * i + 0];
-    t[1] = MLDSA_GAMMA1 - a->coeffs[2 * i + 1];
-
-    r[5 * i + 0] = (t[0]) & 0xFF;
-    r[5 * i + 1] = (t[0] >> 8) & 0xFF;
-    r[5 * i + 2] = (t[0] >> 16) & 0xFF;
-    r[5 * i + 2] |= (t[1] << 4) & 0xFF;
-    r[5 * i + 3] = (t[1] >> 4) & 0xFF;
-    r[5 * i + 4] = (t[1] >> 12) & 0xFF;
+    t[0] = (uint32_t)(MLDSA_GAMMA1 - a->coeffs[2 * i + 0]);
```
hanno-becker, line 696: Can you add a safety comment?

mldsa/polyvec.c
```diff
@@ -157,20 +157,24 @@ void mld_polyvecl_uniform_gamma1(mld_polyvecl *v,
                                  const uint8_t seed[MLDSA_CRHBYTES],
                                  uint16_t nonce)
 {
-  nonce = MLDSA_L * nonce;
+  nonce = (uint16_t)(MLDSA_L * nonce);
```
hanno-becker, line 162: This, too, would benefit from a safety comment explaining why no truncation will happen.

mldsa/polyvec.c
```diff
@@ -157,20 +157,24 @@ void mld_polyvecl_uniform_gamma1(mld_polyvecl *v,
...
 #if MLDSA_L == 4
   mld_poly_uniform_gamma1_4x(&v->vec[0], &v->vec[1], &v->vec[2], &v->vec[3],
-                             seed, nonce, nonce + 1, nonce + 2, nonce + 3);
+                             seed, nonce, (uint16_t)(nonce + 1),
+                             (uint16_t)(nonce + 2), (uint16_t)(nonce + 3));
```
hanno-becker, line 167: Hmm... this is quite ugly. If the nonces in `mld_poly_uniform_gamma1_4x` are always consecutive, why are we passing them all rather than just the first?

mldsa/sign.c
```diff
@@ -587,7 +587,7 @@ int crypto_sign_signature(uint8_t *sig, size_t *siglen, const uint8_t *m,
 
   /* Prepare pre = (0, ctxlen, ctx) */
   pre[0] = 0;
-  pre[1] = ctxlen;
+  pre[1] = (uint8_t)ctxlen;
```
hanno-becker, line 592: Add `/* Cast is safe by virtue of the above check. */`.

test/acvp_mldsa.c
```diff
@@ -198,7 +198,7 @@ static void acvp_mldsa_sigGen_AFT(const unsigned char *message, size_t mlen,
   /* TODO: shouldn't this be moved to be in the internal function? */
   unsigned char pre[MAX_CTX_LENGTH + 2];
   pre[0] = 0;
-  pre[1] = ctxlen;
+  pre[1] = (uint8_t)ctxlen;
```
hanno-becker, line 258: Perhaps make `ctxlen` a `uint8_t` here?

hanno-becker (changes requested):
Thank you @mkannwischer for embarking on this! There are a few cases where I think we should document why the cast is safe, or try to avoid it altogether. Otherwise, we may as well leave the warnings off.

mldsa/fips202/fips202.c
mkannwischer: Comments added.

mldsa/fips202/fips202x4.c
mkannwischer: Comment added.

mldsa/native/aarch64/meta.h
mkannwischer: Changed and comment. This is safe because we sample at most 256 coefficients.

mldsa/native/x86_64/src/rej_uniform_avx2.c
mkannwischer: Okay.

mldsa/poly_kl.c
mkannwischer: Conceptionally the return value is a bit (0/1)
From it we need to compute two things: 
1) A polynomial (which has int32_t coeffients) with 0/1 coefficients
2) The number of hint bits that are 1 (s). 

The latter is fundamentally unsigned and it makes sense to keep it that way.
So you will have to cast one way or the other. Casting when you convert it to a polynomial makes sense to me.

Keeping this unchanged.

mldsa/poly_kl.c
mkannwischer: Done.

mldsa/polyvec.c
mkannwischer: Yes, added a comment referencing the NONCE_UB comment in sign.c

mldsa/polyvec.c
mkannwischer: Okay, changed to only taking a single nonce.

mkannwischer: In that case we should also change `mld_poly_uniform_eta_4x` so that the two functions do not diverge. I pushed that as a separate commit. I'm not fully convinced this is is an improved. The explicit nonce values for all 4 polynomials make it much easier to understand what is going on.

WDYT?

mldsa/sign.c
mkannwischer: Added.

test/acvp_mldsa.c
mkannwischer: I'd rather not. I added an explicit check - that will make it easier to support test vectors that have an invalid ctxlen (should NIST decide to add those add some point).

mldsa/poly.c
```diff
@@ -474,35 +477,35 @@ void mld_polyt0_pack(uint8_t *r, const mld_poly *a)
...
-    r[13 * i + 11] |= (t[7] << 3) & 0xFF;
-    r[13 * i + 12] = (t[7] >> 5) & 0xFF;
+    t[0] = (uint32_t)((1 << (MLDSA_D - 1)) - a->coeffs[8 * i + 0]);
+    t[1] = (uint32_t)((1 << (MLDSA_D - 1)) - a->coeffs[8 * i + 1]);
+    t[2] = (uint32_t)((1 << (MLDSA_D - 1)) - a->coeffs[8 * i + 2]);
+    t[3] = (uint32_t)((1 << (MLDSA_D - 1)) - a->coeffs[8 * i + 3]);
+    t[4] = (uint32_t)((1 << (MLDSA_D - 1)) - a->coeffs[8 * i + 4]);
+    t[5] = (uint32_t)((1 << (MLDSA_D - 1)) - a->coeffs[8 * i + 5]);
+    t[6] = (uint32_t)((1 << (MLDSA_D - 1)) - a->coeffs[8 * i + 6]);
+    t[7] = (uint32_t)((1 << (MLDSA_D - 1)) - a->coeffs[8 * i + 7]);
```
hanno-becker, line 489: Remind me why this is OK?

mkannwischer: It's a pre-condition of the function

```c
  mld_assert_bound(a->coeffs, MLDSA_N, -(1 << (MLDSA_D - 1)) + 1,
                   (1 << (MLDSA_D - 1)) + 1);
```

mkannwischer: In other words: power2round outputs values for a0 that are <= 2^{MLDSA_D-1}

mkannwischer: Probably worthwhile to add a comment about that.

hanno-becker: That would be great, yes :+1:

mkannwischer: Added.

mldsa/poly_kl.c
hanno-becker: Ack

mldsa/poly_kl.c
```diff
@@ -339,10 +338,11 @@ void mld_poly_uniform_eta_4x(mld_poly *r0, mld_poly *r1, mld_poly *r2,
   mld_memcpy(extseed[1], seed, MLDSA_CRHBYTES);
   mld_memcpy(extseed[2], seed, MLDSA_CRHBYTES);
   mld_memcpy(extseed[3], seed, MLDSA_CRHBYTES);
-  extseed[0][MLDSA_CRHBYTES] = nonce0;
-  extseed[1][MLDSA_CRHBYTES] = nonce1;
-  extseed[2][MLDSA_CRHBYTES] = nonce2;
-  extseed[3][MLDSA_CRHBYTES] = nonce3;
+  /* Safety: Truncation is safe because nonce it at most 11 */
```
hanno-becker: This comment is a bit confusing since the contract is weaker and only assumed UINT8_MAX - 3.

mldsa/polyvec.c
hanno-becker: Thanks!

test/acvp_mldsa.c
hanno-becker: Ack

mldsa/polyvec.c
hanno-becker: On second thought, I think you are right. Thank you for making this easy to revert through separate commits.

hanno-becker: Reverted.

Issue #528, opened 2025-10-04
Add bounds reasoning comments to AVX2 backend


mkannwischer:
The AVX2 backend is directly pulled in from the pqcrystals repository. 
We should add documenting comments along the lines of what [was done in mlkem-native](https://github.com/pq-code-package/mlkem-native/blob/18b4bd798288f94ebb2cc6ed5e96f90b77d71f18/dev/x86_64/src/intt.S#L191), in particular regarding coefficient bounds.
This can later help with writing a contract for the native functions and proving them in HOL-Light.

PR #560, opened 2025-10-26
Add bounds reasoning comments to AVX2 [I]NTT and basemul

jammychiou1:
- Towards resolving #528. See follow-up PR #629

### DIFF
--- a/.github/workflows/base.yml
+++ b/.github/workflows/base.yml
@@ -188,7 +188,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        mingw-version: [5.4.0, 11.2.0, 12.2.0, 13.2.0, 14.2.0, 15.2.0]
+        # TODO: re-enable 5.4.0 or other old version
+        mingw-version: [11.2.0, 12.2.0, 13.2.0, 14.2.0, 15.2.0]
     name: Quickcheck (Mingw-w64 ${{ matrix.mingw-version }})
     runs-on: windows-latest
     steps:


dev/x86_64/src/intt.S
```diff
@@ -43,6 +43,12 @@ vpsrlq		$32,%ymm\r0,%ymm\r0
 vpblendd	$0xAA,%ymm\r1,%ymm\r0,%ymm\r3
 .endm
 
+/*
+ * Compute l + h, montmul(h - l, zh) then store the results back to l, h
+ * respectively.
+ *
+ * The general abs bound of Montgomery multiplication is 3q/4.
```
mkannwischer: As discussed offline: Please include reasoing for the 3q/4 bound --- you actually mean Montgomery multiplication by a constant, not general Montgomery multiplication.

dev/x86_64/src/ntt.S
```diff
@@ -44,6 +44,16 @@ vpsrlq		$32,%ymm\r0,%ymm\r0
 vpblendd	$0xAA,%ymm\r1,%ymm\r0,%ymm\r3
 .endm
 
+/*
+ * Compute l + montmul(h, zh), l - montmul(h, zh) then store the results back to
+ * l, h respectively.
+ *
+ * Although the general abs bound of Montgomery multiplication is 3q/4, we use
+ * the more convenient bound q here.
```
mkannwischer: Rephrase and reference intt.S for the 3q/4.

mkannwischer (changes requested):
Thanks @jammychiou1. I checked the bounds tracking for NTT, iNTT, basemul and everything makes sense to me. Please include some reasoning for the 3q/4 bounds.
It would also be great if you can extend https://github.com/pq-code-package/mlkem-native/blob/main/test/test_bounds.py to demonstrate the 3q/4 bound.

Let's move the decompose changes to a separate follow-up PR, please.

dev/x86_64/src/intt.S
jammychiou1: Fixed

dev/x86_64/src/ntt.S
jammychiou1: Fixed

mkannwischer (changes requested):
Thanks @jammychiou1. I am now happy with the changes. Great work! 
Could you clean up the commit history please? Please also update the PR description - this does not yet resolve #528 - we will continue with the other functions in #629.

@hanno-becker could also also review this, please?

mkannwischer (approved):
LGTM now. Thanks @jammychiou1.

dev/x86_64/src/intt.S
```diff
@@ -256,6 +305,31 @@ vmovshdup	%ymm7,%ymm7
...
+ *
+ * In Section 2.2 of https://eprint.iacr.org/2023/1962, they showed a bound that
+ * works for any variable input a, as long as the constant b is signed
+ * canonical:
+ *
+ *   |montmul(a, b)| <= (|a| (q/2) + (R/2) q) / R = (q/2) (1 + |a|/R).
+ *
+ * Therefore, even if we know nothing about a except that it fits inside
+ * int32_t (thus |a| <= R/2), we still have |montmul(a, b)| <= 3q/4. This can be
+ * strengthened to |montmul_pos(a, b)| <= floor(3q/4) < ceil(3q/4) since LHS is
```
hanno-becker: Why is `montmul_pos` mentioned here?

dev/x86_64/src/intt.S
```diff
@@ -256,6 +305,31 @@ vmovshdup	%ymm7,%ymm7
...
+ *
+ * See test/test_bounds.py for more empirical evidence (and some minor technical
+ * details).
+ *
+ * TODO: Use proper citation. Currently, citations within asm can cause linter
+ *       to complain about unused citation, because comments are not preserved
+ *       after simpasm.
+ */
+
+/* 4, 5, 6, 7: abs bound < ceil(3q/4) */
```
hanno-becker: There are a lot of occurrences of `ceil(3q/4)` in the documentation. To reduce the amount of repetition, and seeing that we currently only rely on the output bound of Q, does it make sense to introduce a name for this here and argue only once what the actual value is?

```
In the following, we call MONTMUL_BOUND any absolute bound on the result of a signed Montgomery multiplication of a 16-bit signed integer with a signed canonical integer (in the range -Q/2...Q/2).
...
```

dev/x86_64/src/ntt.S
```diff
@@ -56,16 +67,30 @@ vpmuldq		%ymm0,%ymm13,%ymm13
...
+                                  *   + mulhi(q, mullo(h, zl)) */
 
 vmovshdup	%ymm13,%ymm13
-vpblendd	$0xAA,%ymm14,%ymm13,%ymm13
+vpblendd	$0xAA,%ymm14,%ymm13,%ymm13 /* mulhi(q, mullo(h, zl)) */
 
-vpaddd		%ymm13,%ymm12,%ymm\h
-vpsubd		%ymm13,%ymm\l,%ymm\l
+/* Finish the delayed task mentioned above */
+vpaddd		%ymm13,%ymm12,%ymm\h /* l - montmul(h, zh) */
```
hanno-becker: Note that we don't need to argue about the interim additions in L81 and L84. It doesn't matter if those overflow, associativity holds unconditionally and it's sufficient to do the bounds reasoning on the values in L92 and L93.

(that's also the approach used by the HOL-Light proof)

dev/x86_64/src/pointwise.S
```diff
@@ -92,6 +94,11 @@ _looptop1:
         vpsrlq  ymm2, ymm2, 32
         vpsrlq  ymm4, ymm4, 32
         vmovshdup ymm6, ymm6
+        /*
+         * All coefficients are Montgomery-reduced. This results in the bound
+         *
+         * All: abs bound <= "input abs bound"/R + q/2 < (qR/2)/R + q/2 = q
```
hanno-becker: Can we either spell this out in more detail, or refer to where this bounds reasoning for the Montgomery reduction has happened before?

dev/x86_64/src/intt.S
jammychiou1: Oops! Copypasta. Thanks for the catch!

dev/x86_64/src/ntt.S
jammychiou1: Umm, it can't really overflow here, can it? I mean, the bound for `|montmul(h, zh)|` literally comes from triangle inequality `|mulhi(h, zh) - mulhi(q, mullo(h, zl))| <= |mulhi(h, zh)| + |mulhi(q, mullo(h, zl))|`.

(If a similar trick is done with Barrett multiplication instead, then yes, this is important to note.)

I guess I can still remark that overflow doesn't matter, for those unfamiliar/uninterested to how bounds on montmul is proven.

dev/x86_64/src/pointwise.S
jammychiou1: The new patch refers to mld_montgomery_reduce() in mldsa/src/reduce.h, specifically `MONTGOMERY_REDUCE_STRONG_DOMAIN_MAX`.

I did notice that the bound is also left unexplained there. Think we should address it in #602.

dev/x86_64/src/intt.S
jammychiou1: Addressed in the new patch.

dev/x86_64/src/ntt.S
jammychiou1: Addressed in the new patch.

hanno-becker (approved):

PR #568, opened 2025-10-29
autogen: fix incorrect `dry_run` handling in `gen_bib_file()`

willieyz:
- During PR #567, we notice that the CI will not rising error if there are missing reference in `BIBLIOGRAPHY.md`
- The problem was that the `dry_run` parameter for `gen_bib_file()` was always set to False, so the dry-run mode was never enabled during execution.
- This PR do following:
  - Fixes the issue where the dry_run argument was not working in the `gen_bib_file()` function. As a result, the CI now rising error correctly when reference missing update in `BIBLIOGRAPHY.md`.
  - Adds the missing reference to ensure the bibliography is consistent.


mkannwischer:
The more interesting question is why our CI did not catch this before. We should fix the CI at the same time.

PR #569, opened 2025-10-29
HOL-Light: Allow interactive use of hol.sh in the hol_light shell

willieyz:
- Ported from: https://github.com/pq-code-package/mlkem-native/pull/1244
- Previously, while compiling and running the HOL-Light proofs in `nix develop .#hol_light` worked, trying to start the interactive shell `hol.sh` for proof development would fail. This required proof developers to install `HOL-Light` and `s2n-bignum` locally, both inconvenient and error proone because one would need to keep that local version and the nix version in sync.
- This commit adjusts the nix flake to setup HOL-Light in such a way that hol.sh can be used interactively inside the `hol_light` shell. Now, when you enter the shell via `nix develop .#hol_light`, the interactive shell `hol.sh` is in the PATH, and within it, both `s2n-bignum` and `proofs/hol_light/arm` are added to HOL-Light's own search path (`load_path`).
  Testing for this PR with typing following:
```
   nix develop .#hol_light
   hol.sh
```

- Note that mldsa-native currently does not have any HOL-Light proofs. Therefore, we have not ported the HOL-Light interactive shell CI tests to mldsa-native.

### DIFF
--- a/nix/hol_light/default.nix
+++ b/nix/hol_light/default.nix
@@ -7,13 +7,14 @@ hol_light.overrideAttrs (old: {
   setupHook = writeText "setup-hook.sh" ''
     export HOLDIR="$1/lib/hol_light"
     export HOLLIGHT_DIR="$1/lib/hol_light"
+    export PATH="$1/lib/hol_light:$PATH"
   '';
-  version = "unstable-2024-12-22";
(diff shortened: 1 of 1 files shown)


mkannwischer (approved):
LGTM. Thanks @willieyz!

PR #537, opened 2025-10-08
Add mldsa-native.h API header file

jakemas:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/354

This PR adds mldsa_native.h which is a prerequisite for the monobuilds.

PR #587, opened 2025-11-01
Move sources to mldsa/src

mkannwischer:
To align with mlkem-native and to enable the monobuild, we move the sources to mldsa/src/.
- Depends on #537


hanno-becker (changes requested):
There's a bunch of config files still mentioning `mldsa/fips202` and/or `mldsa/config.h`

hanno-becker (approved):

PR #595, opened 2025-11-03
Use array when passing keys/signature between functions

### LINKED ISSUE #554: Consider using array notation with explicit sizes in function signatures for fixed-size buffers
Currently, many functions in the API use pointer notation with size information only in comments:

```c
  int crypto_sign_signature(uint8_t *sig, size_t *siglen,
                           const uint8_t *m, size_t mlen,
                           const uint8_t *ctx, size_t ctxlen,
                           const uint8_t *sk);
```

  where the comment states uint8_t *sig: pointer to output signature (of length CRYPTO_BYTES) and uint8_t *sk: pointer to bit-packed secret key.

  We could make buffer size requirements more explicit in the function signature itself:
```c
  int crypto_sign_signature(uint8_t sig[CRYPTO_BYTES], 
                           size_t *siglen,
                           const uint8_t *m, size_t mlen,
                           const uint8_t *ctx, size_t ctxlen,
                           const uint8_t sk[CRYPTO_SECRETKEYBYTES]);
```

jammychiou1:
- Resolves #554

Replace uint8_t pointers in function signatures with arrays whenever the size of the buffer is known beforehand.

Note that some `uint8_t *sig` actually contains message whose size is undetermined. For those cases, we still use pointers.

The documentation comments are also updated (and cleaned up slightly) to reflect this change.

### DIFF
--- a/mldsa/mldsa_native.h
+++ b/mldsa/mldsa_native.h
@@ -156,12 +156,13 @@
  *              When MLD_CONFIG_KEYGEN_PCT is set, performs a Pairwise
  *              Consistency Test (PCT) as required by FIPS 140-3 IG.
  *
- * Arguments:   - uint8_t *pk:   pointer to output public key (allocated
- *                               array of MLDSA_PUBLICKEYBYTES bytes)
- *              - uint8_t *sk:   pointer to output private key (allocated
- *                               array of MLDSA_SECRETKEYBYTES bytes)
- *              - uint8_t *seed: pointer to input random seed (MLDSA_SEEDBYTES
- *                               bytes)
+ * Arguments:
+ *     - uint8_t pk[MLDSA_PUBLICKEYBYTES(MLD_CONFIG_API_PARAMETER_SET)]:
+ *           output public key
+ *     - uint8_t sk[MLDSA_SECRETKEYBYTES(MLD_CONFIG_API_PARAMETER_SET)]:
+ *           output private key
+ *     - const uint8_t seed[MLDSA_SEEDBYTES]:
+ *           input random seed
  *
  * Returns 0 (success) or -1 (PCT failure)
  *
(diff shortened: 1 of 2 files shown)


mldsa/mldsa_native.h
```diff
@@ -448,9 +448,11 @@ typedef enum
  **************************************************/
 MLD_API_MUST_CHECK_RETURN_VALUE
 int MLD_API_NAMESPACE(signature_pre_hash_internal)(
-    uint8_t *sig, size_t *siglen, const uint8_t *ph, size_t phlen,
-    const uint8_t *ctx, size_t ctxlen, const uint8_t rnd[MLDSA_RNDBYTES],
-    const uint8_t *sk, mld_hash_alg_t hashAlg);
+    uint8_t sig[MLDSA_BYTES(MLD_CONFIG_API_PARAMETER_SET)], size_t *siglen,
```
mkannwischer, line 480: If there is an explicit length parameter we should stick to pointer notation instead of array notation. 
Please change that back to a pointer. I see that this is also wrong in other places - please fix them too.

mldsa/mldsa_native.h
```diff
@@ -505,9 +508,10 @@ int MLD_API_NAMESPACE(verify_pre_hash_internal)(
  **************************************************/
 MLD_API_MUST_CHECK_RETURN_VALUE
 int MLD_API_NAMESPACE(signature_pre_hash_shake256)(
-    uint8_t *sig, size_t *siglen, const uint8_t *m, size_t mlen,
-    const uint8_t *ctx, size_t ctxlen, const uint8_t rnd[MLDSA_RNDBYTES],
-    const uint8_t *sk);
+    uint8_t sig[MLDSA_BYTES(MLD_CONFIG_API_PARAMETER_SET)], size_t *siglen,
```
mkannwischer, line 545: same here.

mkannwischer (changes requested):
Thanks @jammychiou1! 

Can you please also adjust the documentation comments to use array notation instead of points?

Also some sig buffers that may have a different length (as we are using an explicit siglen parameter) should be changes back to array notation.

mldsa/mldsa_native.h
jammychiou1: I thought `siglen` in this function is an output parameter, used to return "number of byte that was written to `sig` in this function" to the caller, and will only be 0 or `CRYPTO_BYTES`.

If that's the case, the array notation should be fine? Am I missing something?

mkannwischer: You are right. Sorry about that. no change needed here.

mldsa/mldsa_native.h
mkannwischer: same - sorry.

Issue #591, opened 2025-11-02
HashML-DSA: Replace enum by #define


mkannwischer:
Currently we are using an enum to specify which pre-hashing hash function to be used: 
https://github.com/pq-code-package/mldsa-native/blob/1d7ed8d83f0afd3d1796d35ca980a0df56443e91/mldsa/sign.h#L495-L499
https://github.com/pq-code-package/mldsa-native/blob/1d7ed8d83f0afd3d1796d35ca980a0df56443e91/mldsa/sign.h#L48-L62

This leads to problems when introducing a separate `mldsa_native.h` specifying the public API that can be included multiple times as it introduces clashes. #537 works around this guarding it by an `#ifndef` to only define it once.
Futhermore, there is a (theoretical) concern about the type of the enum being implementation-defined in C90: https://github.com/pq-code-package/mldsa-native/pull/537#discussion_r2484226645.

We should eliminate the enum and instead use #defines to define numerical values for each possible value.

PR #605, opened 2025-11-04
HashML-DSA: Replace enum by define to pick pre-hash function

mkannwischer:
Currently we have an enum for the 12 different pre-hash functions for HashML-DSA. This leads to problems in multi-level builds as we must only define the enum once. Currently we work around this by guarding the enum definition with a pre-processor conditional.
However, there is also a (theoretical) concern about the type of the enum being implementation-defined in C90:
https://github.com/pq-code-package/mldsa-native/pull/537#discussion_r2484226645

It seems cleaner to not use an enum here, but instead use #defines avoiding all the above problems.
This commit implements that change.

We also eliminate the camel case hashAlg - that was inconsistent with the remaining code base from the start.

- Resolves https://github.com/pq-code-package/mldsa-native/issues/591


mldsa/mldsa_native.h
```diff
@@ -422,27 +422,20 @@ int MLD_API_NAMESPACE(open)(
...
+#define MLD_PREHASH_SHA2_384 3
+#define MLD_PREHASH_SHA2_512 4
+#define MLD_PREHASH_SHA2_512_224 5
+#define MLD_PREHASH_SHA2_512_256 6
+#define MLD_PREHASH_SHA3_224 7
+#define MLD_PREHASH_SHA3_256 8
+#define MLD_PREHASH_SHA3_384 9
+#define MLD_PREHASH_SHA3_512 10
+#define MLD_PREHASH_SHAKE_128 11
+#define MLD_PREHASH_SHAKE_256 12
```
hanno-becker, line 438: This is now re-defined 3 times in a multi-level build, but for syntactically equal `#define`s, that's OK, and the situation is already the same for e.g. the various size constants.

PR #607, opened 2025-11-04
Add Runtime dispatch based on custom CPU capabilities function

### LINKED ISSUE #609: Port run-time dispatch from mlkem-native
https://github.com/pq-code-package/mlkem-native/pull/1152

willieyz:
- Resolves #609 
- This PR ported: https://github.com/pq-code-package/mlkem-native/pull/1152

This PR made following changes:
  - Port runtime dispatch mechanism for AArch64, x86 native, and FIPS202 APIs.
  - Port configuration for static ON/OFF, AVX2 and AA64PFR1_EL1, as well as the corresponding config-variations tests in CI.

CBMC-related changes, including adding the CBMC proof for sys_check_capability, will be added in a separate PR (tracked by issue #723).

Additional changes discovered during PR progress, related with `mld_poly_chknorm`:
  - The return value of `mld_poly_chknorm` has been adjusted to properly signal fallback, instead of directly using 0 or -1. 
  This distinction allows the runtime to detect when to fallback to the C implementation.
    - 0 → all coefficients are within the bound 
    - 1 → at least one coefficient exceeds the bound 
    - -1 (MLD_NATIVE_FUNC_FALLBACK) → the platform does not have the required capabilities to run the native function
  - Add a declassification of `(ret == FALLBACK)` to make this code pass the constant time tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -513,32 +513,28 @@ jobs:
             ec2_volume_size: 20
             compile_mode: native
             opt: all
-            # TODO: This config do not exist for now, will be available after PR #607 ("Add Runtime Dispatch") is merged.
-            # config_variations: 'native-cap-CPUID_AVX2'
+            config_variations: 'native-cap-CPUID_AVX2'
           - name: Intel Xeon 4th gen (t3)
             ec2_instance_type: t3.small
             ec2_ami: ubuntu-latest (x86_64)
             ec2_volume_size: 20
             compile_mode: native
             opt: all
-            # TODO: This config do not exist for now, will be available after PR #607 ("Add Runtime Dispatch") is merged.
-            # config_variations: 'native-cap-CPUID_AVX2'
+            config_variations: 'native-cap-CPUID_AVX2'
           - name: Graviton2 (c6g.medium)
             ec2_instance_type: c6g.medium
             ec2_ami: ubuntu-latest (aarch64)
             ec2_volume_size: 20
             compile_mode: native
             opt: all
-            # TODO: This config do not exist for now, will be available after PR #607 ("Add Runtime Dispatch") is merged.
-            # config_variations: 'native-cap-ON native-cap-OFF native-cap-ID_AA64PFR1_EL1'
+            config_variations: 'native-cap-ON native-cap-OFF native-cap-ID_AA64PFR1_EL1'
           - name: Graviton3 (c7g.medium)
             ec2_instance_type: c7g.medium
             ec2_ami: ubuntu-latest (aarch64)
             ec2_volume_size: 20
             compile_mode: native
             opt: all
-            # TODO: This config do not exist for now, will be available after PR #607 ("Add Runtime Dispatch") is merged.
-            # config_variations: 'native-cap-ID_AA64PFR1_EL1'
+            config_variations: 'native-cap-ID_AA64PFR1_EL1'
     name: Platform tests  (${{ matrix.target.name }})
     permissions:
       contents: 'read'


mldsa/src/poly_kl.c
```diff
@@ -37,19 +37,38 @@
...
+#if defined(MLD_USE_NATIVE_POLY_DECOMPOSE_88) && MLD_CONFIG_PARAMETER_SET == 44
+  {
+    int ret;
+    /* TODO: proof */
+    ret = mld_poly_decompose_88_native(a1->coeffs, a0->coeffs, a->coeffs);
+    if (ret == MLD_NATIVE_FUNC_SUCCESS)
+    {
+      mld_assert_abs_bound(a0->coeffs, MLDSA_N, MLDSA_GAMMA2 + 1);
+      mld_assert_bound(a1->coeffs, MLDSA_N, 0,
+                       (MLDSA_Q - 1) / (2 * MLDSA_GAMMA2));
```
hanno-becker, line 51: General comment: Those repeated post-conditions are not ideal. Seeing that they are the same for native and C code, one may want to follow a logic
```c
try_native();
if (fallback) {
    do_C();
}
check_bounds();
```
Can be a follow up

dev/fips202/aarch64/x4_v8a_scalar.h
```diff
@@ -13,11 +13,17 @@
...
 #if !defined(__ASSEMBLER__)
+#include "../api.h"
 #include "src/fips202_native_aarch64.h"
-static MLD_INLINE void mld_keccak_f1600_x4_native(uint64_t *state)
+static MLD_INLINE int mld_keccak_f1600_x4_native(uint64_t *state)
 {
+  if (!mld_sys_check_capability(MLD_SYS_CAP_SHA3))
+  {
+    return MLD_NATIVE_FUNC_FALLBACK;
+  }
```
mkannwischer: This is wrong. This function does not need SHA-3 instructions.

mldsa/src/poly.c
```diff
@@ -589,19 +589,37 @@ void mld_polyt0_unpack(mld_poly *r, const uint8_t *a)
...
+     * -1 (MLD_NATIVE_FUNC_FALLBACK) if the platform does not have the
+     * required capabilities to run the native function.
+     */
+    ret = mld_poly_chknorm_native(a->coeffs, B);
+
+    success = (ret != MLD_NATIVE_FUNC_FALLBACK);
+    /* Constant-time: It would be fine to leak the return value of chknorm
+     * entirely (as it is fine to leak if any coefficient exceeded the bound or
+     * not). However, it is cleaner to perform declassification in sign.c.
+     * Hence, here we only declassify if a the native function returned
```
hanno-becker: Typo

hanno-becker: This should be complemented by explicit documentation in the native API that the decision of whether to fallback or not must never depend on the input data itself (only system capabilities and, where present, length information). This applies to all backend functions.

hanno-becker (changes requested):
Thanks a lot @willieyz for all your work!

Overall, it looks good, but there are a few points we may want to refine:
1. As @mkannwischer mentions, some SHA3 guards in the AArch64 FIPS202 are unnecessary.
2. We need documentation in the native backend that the decision whether to fall back or not must not depend on the input data. This is part of the contract between frontend and backend.
3. There is a lot of duplicated post-conditions now which we may want to remove.

.github/actions/config-variations/action.yml
mkannwischer, line 1: Please align this file with mlkem-native.

mkannwischer:
> 3. There is a lot of duplicated post-conditions now which we may want to remove.

The next step is going to be to hoist out the C versions to separate functions to unblock the backend unit tests.
Then you will definitely need the asserations twice - the position of those can be argued about, but I'd want it to align with mlkem-native.

dev/fips202/aarch64/x4_v8a_scalar.h
willieyz: Thank you for your review!
Fixed.

mldsa/src/poly.c
willieyz: Thank you for your review!
fixed.

willieyz: I had add the documentation at `mldsa/src/native/api.h`,
Thank you for your review!

.github/actions/config-variations/action.yml
willieyz: I had aligned this and squash to corresponding commit,
Thank you for your review!

mldsa/src/native/api.h
```diff
@@ -31,6 +39,12 @@
  * To add a function to a backend, define MLD_USE_NATIVE_XXX and
  * implement `static inline xxx(...)` in the profile header.
  *
+ * IMPORTANT: Backend implementations must ensure that the decision of whether
+ * to fallback (return MLD_NATIVE_FUNC_FALLBACK) or not must never depend on
+ * the input data itself. Fallback decisions may only depend on system
+ * capabilities (e.g., CPU features) and, where present, length information.
+ * This requirement applies to all backend functions to maintain constant-time
+ * properties.
```
mkannwischer: CAn you move this further up to the text about `MLD_NATIVE_FUNC_FALLBACK`, i.e., after line 30. 

Also please duplicate this text into the fips202 api.h in the same location

mldsa/src/sys.h
```diff
@@ -203,4 +211,37 @@
...
+{
+#if defined(MLD_SYS_X86_64)
+  MLD_SYS_CAP_AVX2
+#elif defined(MLD_SYS_AARCH64)
+  MLD_SYS_CAP_SHA3
+#else
+  /* C90 does not allow empty enums, so use a dummy value
+   * for architectures other than AArch64 and x86_64. */
+  MLD_SYS_CAP_DUMMY
+#endif
```
mkannwischer: This breaks CBMC proofs on x86 - see https://github.com/pq-code-package/mlkem-native/issues/1318. 
Could you please change this to include my changes from https://github.com/pq-code-package/mlkem-native/pull/1319?

willieyz: Hello, Matthias, I had included this changes.
Many thanks for your reviewing!

mldsa/src/native/api.h
willieyz: Understood, I had shift it up to the text right before `MLD_NATIVE_FUNC_FALLBACK`. thanks!

hanno-becker (approved):
Thank you @willieyz, this is great. LGTM

PR #558, opened 2025-10-26
Add `MLD_CONFIG_SERIAL_FIPS202_ONLY` option

mkannwischer:
This PR ports the following two PRs from mlkem-native:
 - https://github.com/pq-code-package/mlkem-native/pull/1237
 - https://github.com/pq-code-package/mlkem-native/pull/1231

Unfortunately, the changes to mldsa-native are more invasive, as we have to touch 3 functions `mld_polyvecl_uniform_gamma1`, `mld_sample_s1_s2`, and `mld_polyvec_matrix_expand).
See the individual commit messages for details.

 - Resolves #542

PR #611, opened 2025-11-05
CBMC: Improve proof stability

### LINKED ISSUE #547: CBMC proof performance regression
#468 caused a performance regression for the ML-DSA-87 CBMC proofs from 8 minutes to 20 minutes.
It's odd because #468 does not modify any code that should be relevant to CBMC proofs.
We should investigate and get the proof performance back to what it was before.

mkannwischer:
This hoists @rod-chapman's work to fix and stabilize CBMC proofs in #558.
The changes are independent of the changes in #558, so we should review and merge them separately.

- Resolves #547

### DIFF
--- a/proofs/cbmc/pointwise_sum_of_products/Makefile
+++ b/proofs/cbmc/pointwise_sum_of_products/Makefile
@@ -26,7 +26,7 @@ USE_DYNAMIC_FRAMES=1
 
 # Disable any setting of EXTERNAL_SAT_SOLVER, and choose SMT backend instead
 EXTERNAL_SAT_SOLVER=
-CBMCFLAGS=--smt2 --slice-formula
+CBMCFLAGS=--smt2 --slice-formula --no-array-field-sensitivity
 
 FUNCTION_NAME = pointwise_sum_of_products
 


mldsa/src/polyvec.c
```diff
@@ -105,27 +100,31 @@ void mld_polyvec_matrix_expand(mld_polyvecl mat[MLDSA_K],
...
    */
+#ifndef CBMC
   for (i = 0; i < MLDSA_K; i++)
   {
     for (j = 0; j < MLDSA_L; j++)
     {
       mld_poly_permute_bitrev_to_custom(mat[i].vec[j].coeffs);
     }
   }
+#endif /* !CBMC */
```
mkannwischer: I think this is a very bad idea. (1) It introduces a proof gap; (2). It's ugly.

mkannwischer:
Proof runtime in CI improves significantly:
 - ML-DSA-44: 9m -> 6m
 - ML-DSA-65: 8m -> 7m
 - ML-DSA-87: 11m -> 9m 
 
 
 More importantly: It makes proofs in #558 feasible.

mldsa/src/polyvec.c
rod-chapman: I agree! Unrolling the loop leads to instability, so I will see if I can do it with an invariant.

rod-chapman: or factor it out into a distinct function.

mkannwischer: Factoring it out would also align it with mlkem-native - I'm in favour of that! Thanks.

rod-chapman: Data:
Proof times for the 3 parameter sets as-is on r7g: 25s, 2m38s, 4m respectively.
With the `#ifndef CBMC` guard removed, it's 30s, 100s, and "gave up after 30 minutes".

Refactoring into a single function as-per mlkem-native looks like the best option.

hanno-becker: Agree, this is a show-stopper. We must not alter code depending on whether it is under verification or not. This is a big soundness gun and can cost us a lot of trust. This may be a vehicle for debugging and prototyping, but nothing else. Please, if we do this in the future, can be always mark PRs as DO-NOT-MERGE or otherwise make it very clear that the PR is _not_ ready?

rod-chapman: Done - introduced local mld_matrix_permute_bitrev_to_custom() function, and isolated native-only functionality in there.

PR #614, opened 2025-11-05
CI: Do not use custom AMI for CBMC runners

mkannwischer:
This PR ports part of https://github.com/pq-code-package/mlkem-native/pull/1250

Currently we are using custom AMIs for the CBMC tests. The only
reason for using custom AMI is that require a disk of larger than
the default 8 GB.
This commit switches to default AMIs and instead uses the ec2_volume_size
argument of the ec2-github-runner action.

### DIFF
--- a/.github/workflows/cbmc.yml
+++ b/.github/workflows/cbmc.yml
@@ -18,8 +18,8 @@ jobs:
     with:
       name: CBMC (ML-DSA-44)
       ec2_instance_type: c7g.4xlarge
-      ec2_ami: ubuntu-latest (custom AMI)
-      ec2_ami_id: ami-0d7f502261b31b27f # aarch64, ubuntu-latest, 64g
+      ec2_ami: ubuntu-latest (aarch64)
+      ec2_volume_size: 20
       compile_mode: native
       opt: no_opt
       lint: false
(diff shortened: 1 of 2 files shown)


.github/workflows/cbmc.yml
```diff
@@ -17,9 +17,9 @@ jobs:
     uses: ./.github/workflows/ci_ec2_reusable.yml
     with:
       name: CBMC (ML-DSA-44)
-      ec2_instance_type: c7g.4xlarge
-      ec2_ami: ubuntu-latest (custom AMI)
-      ec2_ami_id: ami-0d7f502261b31b27f # aarch64, ubuntu-latest, 64g
+      ec2_instance_type: r8g.xlarge
```
rod-chapman: r8g.xlarge has only 4 CPU cores and 32GB RAM. Is that really enough compared to the previous c7g.4xlarge (16 cores and 32GB RAM)?

mkannwischer:
Proof times are increased 6 -> 10mins; 6->11mins; 8->12mins when switching from c7g.4xlarge -> r8g.xlarge. This is still a net saving in proof cost with current EC2 pricing: https://aws.amazon.com/ec2/pricing/on-demand/. 

Let me try r8g.2xlarge.

Issue #612, opened 2025-11-05
Port: autogen/simpasm: Add support for Intel syntax


mkannwischer:
https://github.com/pq-code-package/mlkem-native/pull/1275

PR #616, opened 2025-11-05
autogen/simpasm: Add support for Intel syntax

willieyz:
- Resolves: #612 
- This PR is ported from: https://github.com/pq-code-package/mlkem-native/pull/1275
- This PR extends `scripts/simpasm` to support the emission of x86_64 assembly in Intel syntax via `--x86-64-syntax` intel, also the CI is extended to exercise this on an x86_64 runner and run the full tests afterwards.


mkannwischer (approved):
Thanks @willieyz. Tested this on my x86 machine and it works as it should.

PR #618, opened 2025-11-06
Consistently use mld_memcpy and mld_memset where possible

mkannwischer:
There have been a handful of leftover loops doing memcpy and memset manually. This commit replaces the loops with mld_memcpy and mld_memset.

There are two notable cases that cannot be replaced: the copying of sm <-> m
in crypto_sign() and crypto_sign_open() as those buffers may overlap and,
hence, would require memmove.
As we currently don't rely on memmove anywhere else, I am hesistent to
introduce this stdlib dependency here.

### DIFF
--- a/mldsa/src/sign.c
+++ b/mldsa/src/sign.c
@@ -594,7 +594,6 @@ int crypto_sign_signature(uint8_t sig[CRYPTO_BYTES], size_t *siglen,
                           size_t ctxlen,
                           const uint8_t sk[CRYPTO_SECRETKEYBYTES])
 {
-  size_t i;
   uint8_t pre[257];
   uint8_t rnd[MLDSA_RNDBYTES];
   int result;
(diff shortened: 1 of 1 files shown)


mldsa/src/fips202/keccakf1600.c
```diff
@@ -39,15 +39,14 @@
...
 {
-  unsigned i;
 #if defined(MLD_SYS_LITTLE_ENDIAN)
   uint8_t *state_ptr = (uint8_t *)state + offset;
-  for (i = 0; i < length; i++)
-  __loop__(invariant(i <= length))
+  if (length > 0)
   {
-    data[i] = state_ptr[i];
+    mld_memcpy(data, state_ptr, length);
```
hanno-becker: We have an explicit loop in mlkem-native. Is it worth calling into `memcpy` for this?

mkannwischer: I thought that we want to replace all copying operations with mld_memcpy? What's a reason to keep the loop?

mkannwischer: Removed from this PR for now.

Issue #598, opened 2025-11-03
Example: `basic_deterministic`


mkannwischer:
We should add an example showcasing deterministic ML-DSA.
We can use that opportunity to also test #597.

Issue #597, opened 2025-11-03
Config: Add `MLD_CONFIG_NO_RANDOMIZED_API` option


mkannwischer:
We should add a `MLD_CONFIG_NO_RANDOMIZED_API` configuration option allowing to build mldsa-native without providing a randombytes() implementation. This can be excercised in a deterministic example.

PR #619, opened 2025-11-06
Config: Add `MLD_CONFIG_NO_RANDOMIZED_API`

mkannwischer:
This adds the MLD_CONFIG_NO_RANDOMIZED_API configuration option to the config
file aligning with mlkem-native.

MLD_CONFIG_NO_RANDOMIZED_API allows to exclude all top-level APIs requiring
randombytes() from a mldsa-native build. This way consumers only requiring
the deterministic internal APIs do not have to provide a randomytes
implementation.
The excluded APIs  are currently crypto_sign_keypair, crypto_sign,
crypto_sign_signature, and crypto_sign_signature_extmu.

Currently, this option is incompatible with the MLD_CONFIG_KEYGEN_PCT
configuration option at that relies on crypto_sign_signature.
A corresponding error is added in case both configuration options are set.

The `basic_deterministic` example is added to excercise the configuration option. 

- Resolves https://github.com/pq-code-package/mldsa-native/issues/597
- Resolves https://github.com/pq-code-package/mldsa-native/issues/598

### DIFF
--- a/scripts/tests
+++ b/scripts/tests
@@ -11,12 +11,12 @@ See the command line interface for more information."""
 import platform
 import argparse
 import os
+import re
 import sys
 import time
 import logging
 import subprocess
 import json
-import re
 
 from enum import Enum
 from functools import reduce
@@ -582,6 +582,12 @@ class Tests:
             l = TEST_TYPES.examples()
         else:
             l = list(map(TEST_TYPES.from_string, self.args.l))
+
+        # Filter out excluded examples
+        if hasattr(self.args, "exclude_example") and self.args.exclude_example:
+            excluded = [TEST_TYPES.from_string(ex) for ex in self.args.exclude_example]
+            l = [e for e in l if e not in excluded]
+
         for e in l:
             self._compile_schemes(e, None)
             self._run_scheme(e, None, None)
@@ -648,6 +654,8 @@ class Tests:
         self.check_fail()
 
     def stack(self):
+        """Stack usage analysis"""
+
         def _stack(opt):
             self._compile_schemes(TEST_TYPES.STACK, opt)
             if self.args.run:
@@ -769,6 +777,7 @@ class Tests:
                         )
                         exit(1)
                     continue
+
                 end = time.time()
                 dur = int(end - start)
                 if p.returncode != 0:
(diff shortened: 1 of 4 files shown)


scripts/tests
```diff
@@ -582,6 +582,12 @@ class Tests:
             l = TEST_TYPES.examples()
         else:
             l = list(map(TEST_TYPES.from_string, self.args.l))
+
+        # Filter out excluded examples
+        if hasattr(self.args, "exclude_example") and self.args.exclude_example:
```
hanno-becker, line 587: This `hasattr(self.args, "exclude_example")` seems unnecessary as there's a `default`. I think one can just drop the conditional.

mldsa/src/sign.c
```diff
@@ -230,6 +230,7 @@ int crypto_sign_keypair_internal(uint8_t pk[CRYPTO_PUBLICKEYBYTES],
   return 0;
 }
 
+#if !defined(MLD_CONFIG_NO_RANDOMIZED_API)
```
hanno-becker, line 233: Shouldn't we modify `sign.h` as well?

I'm not sure we can/should do something about `mldsa_native.h`. We don't have access to the config in there. In mlkem-native, I think it's a (harmless) bug that `mlkem_native.h` uses `MLK_CONFIG_NO_RANDOMIZED_API`.

hanno-becker: Ah... mlkem-native also does not alter `kem.h`. We should leave it or change it in both places.

scripts/tests
mkannwischer: This was copied from mlkem-native. I prefer to keep it aligned and it does not bother me enough that I want to change it in mlkem-native.

mldsa/src/sign.c
mkannwischer: guarding it in mldsa_native.h would provide the most value, but I can't think of a way how we can do that other than adding a `MLD_CONFIG_API_NO_RANDOMIZED_API` flag and that does not seem appealing at all as it seems unlikely that a consumer correctly sets  `MLD_CONFIG_API_NO_RANDOMIZED_API` and then tried to use one of the randomized APIs... 

As for sign.h, I had aligned it with mlkem-native not guarding it. We could guard it but the benefit is marginal as it's mostly used internally.

PR #629, opened 2025-11-06
Update AVX2 decompose to use a more explainable (and very slightly faster) approach, along with bounds reasoning comments.

### LINKED ISSUE #528: Add bounds reasoning comments to AVX2 backend

jammychiou1:
- Resolves #528
- Separated out from #560

This approach was suggested by Hanno Becker in https://github.com/pq-code-package/mldsa-native/pull/411#discussion_r2371088154, when we implemented the same function in AArch64.

The speedup is barely noticeable, less than 50 cycles per call on my laptop.

### DIFF
--- a/dev/x86_64/src/poly_decompose_32_avx2.c
+++ b/dev/x86_64/src/poly_decompose_32_avx2.c
@@ -54,9 +54,9 @@ void mld_poly_decompose_32_avx2(__m256i *a1, __m256i *a0, const __m256i *a)
 
     /* check-magic: 4092 == intdiv(2 * intdiv(MLDSA_Q - 1, 32), 128) */
     /*
-     * The goal is to compute f1 = round-(f / (2*GAMMA2)), which can be computed
-     * alternatively as round-(f / (128B)) = round-(ceil(f / 128) / B) where
-     * B = 2*GAMMA2 / 128 = 4092. Here round-() denotes "round half down".
+     * Compute f1 = round-(f / (2*GAMMA2)) as round-(f / (128B)) =
+     * round-(ceil(f / 128) / B) where B = 2*GAMMA2 / 128 = 4092. See
+     * mld_decompose() in mldsa/src/rounding.h for more details.
      *
      * range: 0 <= f <= Q-1 = 32*GAMMA2 = 16*128*B
      */
--- a/dev/x86_64/src/poly_decompose_88_avx2.c
+++ b/dev/x86_64/src/poly_decompose_88_avx2.c
@@ -55,9 +55,9 @@ void mld_poly_decompose_88_avx2(__m256i *a1, __m256i *a0, const __m256i *a)
 
     /* check-magic: 1488 == intdiv(2 * intdiv(MLDSA_Q - 1, 88), 128) */
     /*
-     * The goal is to compute f1 = round-(f / (2*GAMMA2)), which can be computed
-     * alternatively as round-(f / (128B)) = round-(ceil(f / 128) / B) where
-     * B = 2*GAMMA2 / 128 = 1488. Here round-() denotes "round half down".
+     * Compute f1 = round-(f / (2*GAMMA2)) as round-(f / (128B)) =
+     * round-(ceil(f / 128) / B) where B = 2*GAMMA2 / 128 = 1488. See
+     * mld_decompose() in mldsa/src/rounding.h for more details.
      *
      * range: 0 <= f <= Q-1 = 88*GAMMA2 = 44*128*B
      */
--- a/mldsa/src/rounding.h
+++ b/mldsa/src/rounding.h
@@ -93,20 +93,46 @@ __contract__(
   ensures((*a1 * 2 * MLDSA_GAMMA2 + *a0 - a) % MLDSA_Q == 0)
 )
 {
+  /*
+   * The goal is to compute f1 = round-(f / (2*GAMMA2)), which can be computed
+   * alternatively as round-(f / (128B)) = round-(ceil(f / 128) / B) where
+   * B = 2*GAMMA2 / 128. Here round-() denotes "round half down".
+   *
+   * The equality round-(f / (128B)) = round-(ceil(f / 128) / B) can deduced
+   * as follows. Since changing f to align-up(f, 128) can move f onto but not
+   * across a rounding boundary for division by 128*B (note that we need B to be
+   * even for this to work), and round- rounds down on the boundary, we have
+   *
+   *   round-(f / (128B)) = round-(align-up(f, 128) / (128B))
+   *                      = round-((align-up(f, 128) / 128) / B)
+   *                      = round-(ceil(f / 128) / B).
+   */
   *a1 = (a + 127) >> 7;
   /* We know a >= 0 and a < MLDSA_Q, so... */
   /* check-magic: 65472 == round((MLDSA_Q-1)/128) */
   mld_assert(*a1 >= 0 && *a1 <= 65472);
 
 #if MLD_CONFIG_PARAMETER_SET == 44
-  /* check-magic: 11275 == round((2**24*128) / ((MLDSA_Q - 1) / 44)) */
+  /* check-magic: 1488 == 2 * intdiv(intdiv(MLDSA_Q - 1, 88), 128) */
+  /* check-magic: 11275 == floor(2**24 / 1488) */
+  /*
+   * Compute f1 = round-(f1' / B) ≈ round(f1' * 11275 / 2^24). This is exact
+   * for 0 <= f1' < 2^16. Note that half is rounded down since 11275 / 2^24 ≲
+   * 1 / 1488.
+   */
   *a1 = (*a1 * 11275 + (1 << 23)) >> 24;
   mld_assert(*a1 >= 0 && *a1 <= 44);
 
   *a1 = mld_ct_sel_int32(0, *a1, mld_ct_cmask_neg_i32(43 - *a1));
   mld_assert(*a1 >= 0 && *a1 <= 43);
 #else /* MLD_CONFIG_PARAMETER_SET == 44 */
-  /* check-magic: 1025 == round((2**22*128) / ((MLDSA_Q - 1) / 16)) */
+  /* check-magic: 4092 == 2 * intdiv(intdiv(MLDSA_Q - 1, 32), 128) */
+  /* check-magic: 1025 == floor(2**22 / 4092) */
+  /*
+   * Compute f1 = round-(f1' / B) ≈ round(f1' * 1025 / 2^22). This is exact
+   * for 0 <= f1' < 2^16. Note that half is rounded down since 1025 / 2^22 ≲
+   * 1 / 4092.
+   */
   *a1 = (*a1 * 1025 + (1 << 21)) >> 22;
   mld_assert(*a1 >= 0 && *a1 <= 16);
 
--- a/mldsa/src/native/x86_64/src/poly_decompose_32_avx2.c
+++ b/mldsa/src/native/x86_64/src/poly_decompose_32_avx2.c
@@ -54,9 +54,9 @@ void mld_poly_decompose_32_avx2(__m256i *a1, __m256i *a0, const __m256i *a)
 
     /* check-magic: 4092 == intdiv(2 * intdiv(MLDSA_Q - 1, 32), 128) */
     /*
-     * The goal is to compute f1 = round-(f / (2*GAMMA2)), which can be computed
-     * alternatively as round-(f / (128B)) = round-(ceil(f / 128) / B) where
-     * B = 2*GAMMA2 / 128 = 4092. Here round-() denotes "round half down".
+     * Compute f1 = round-(f / (2*GAMMA2)) as round-(f / (128B)) =
+     * round-(ceil(f / 128) / B) where B = 2*GAMMA2 / 128 = 4092. See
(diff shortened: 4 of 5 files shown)


dev/x86_64/src/poly_decompose_32_avx2.c
```diff
@@ -36,30 +36,75 @@
...
+     * for 0 <= f1' < 2^16. Note that half is rounded down since 1025 / 2^22 ≲
+     * 1 / 4092.
+     *
+     * The odd-index 16-bit lanes are still all 0 after this. As such, despite
+     * that the following steps use 32-bit lanes, the value of f1 is unaffected.
+     */
+    /* check-magic: on */
     f1 = _mm256_mulhi_epu16(f1, v);
     f1 = _mm256_mulhrs_epi16(f1, shift);
-    f1 = _mm256_and_si256(f1, mask);
```
mkannwischer, line 57: Please add a `Reference:` comment here on how it differs from the pq-crystals implementation

dev/x86_64/src/poly_decompose_32_avx2.c
```diff
@@ -36,30 +36,75 @@
...
     f1 = _mm256_srli_epi32(f1, 7);
+    /*
+     * range: 0 <= f1' <= (Q-1)/128 = 16B
+     *
+     * Also, f1' <= (Q-1)/128 = 2^16 - 2^6 < 2^16 ensures that the odd-index
+     * 16-bit lanes are all 0, so no bits will be dropped in the input of the
+     * _mm256_mulhi_epu16() below.
+     */
+
+    /* check-magic: off */
```
mkannwischer: These check-magic comments are distracting as you already explain the constants well. Please add the magic constants you need to the whitelist in the check-magic script. 
@hanno-becker, maybe we want to ignore magic constants in comments altogether?

dev/x86_64/src/poly_decompose_32_avx2.c
jammychiou1: Done

dev/x86_64/src/poly_decompose_32_avx2.c
jammychiou1: Should be done now. I'm not entirely sure how the the check-magic script works - if the way I whitelisted it is not ideal, please let me know. Thanks!

mkannwischer (approved):
Thanks @jammychiou1. I am good with the changes now.

WDYT @hanno-becker?

dev/x86_64/src/poly_decompose_32_avx2.c
```diff
@@ -36,30 +36,75 @@
 void mld_poly_decompose_32_avx2(__m256i *a1, __m256i *a0, const __m256i *a)
 {
   unsigned int i;
-  __m256i f, f0, f1;
```
hanno-becker, line 39: The top of the file says
```
/*
 * This file is derived from the public domain
 * AVX2 Dilithium implementation @[REF_AVX2].
 */
```
one should remark what changes have been made.

dev/x86_64/src/poly_decompose_32_avx2.c
```diff
@@ -36,30 +36,75 @@
...
-  const __m256i mask = _mm256_set1_epi32(15);
 
   for (i = 0; i < MLDSA_N / 8; i++)
   {
     f = _mm256_load_si256(&a[i]);
+
+    /* check-magic: 4092 == 2 * ((MLDSA_Q-1) // 32) // 128 */
+    /*
+     * The goal is to compute f1 = round-(f / (2*GAMMA2)), which can be computed
+     * alternatively as round-(f / (128B)) = round-(ceil(f / 128) / B) where
```
hanno-becker: A brief comment on why that equality holds would be helpful. It should at the same time be added to the C code, which follows the same approach I think.

My understanding is that rounding up to a multiple of 128 can move you to a half-boundary for the division by 2*GAMMA2, but since we round-half-down, that does not change the result.

dev/x86_64/src/poly_decompose_32_avx2.c
```diff
@@ -36,30 +36,75 @@
...
+     */
+
+    /* check-magic: off */
+    /*
+     * Compute f1 = round-(f1' / B) ≈ round(f1' * 1025 / 2^22). This is exact
+     * for 0 <= f1' < 2^16. Note that half is rounded down since 1025 / 2^22 ≲
+     * 1 / 4092.
+     *
+     * The odd-index 16-bit lanes are still all 0 after this. As such, despite
+     * that the following steps use 32-bit lanes, the value of f1 is unaffected.
```
hanno-becker: This is a bit confusing. The value of `f1` _is_ affected by the steps below?

dev/x86_64/src/poly_decompose_32_avx2.c
hanno-becker: Agree the comments are distracting. However, why do you need to turn check-magic off? Didn't you explain both 4092 and 1025 above?

(Nit on terminology: Let's use "allow list" and "deny/block list" moving forward)

dev/x86_64/src/poly_decompose_32_avx2.c
```diff
@@ -36,30 +36,75 @@
...
+     * Also, f1' <= (Q-1)/128 = 2^16 - 2^6 < 2^16 ensures that the odd-index
+     * 16-bit lanes are all 0, so no bits will be dropped in the input of the
+     * _mm256_mulhi_epu16() below.
+     */
+
+    /* check-magic: off */
+    /*
+     * Compute f1 = round-(f1' / B) ≈ round(f1' * 1025 / 2^22). This is exact
+     * for 0 <= f1' < 2^16. Note that half is rounded down since 1025 / 2^22 ≲
+     * 1 / 4092.
```
hanno-becker, line 78: Same quibble as in the AArch64 version, I don't think the comments are sufficient to understand what's going on. Can we create an issue to work out a better explanation for both AArch64 and AVX2? Should not block this PR.

dev/x86_64/src/poly_decompose_32_avx2.c
```diff
@@ -36,30 +36,75 @@
...
+     *
+     * The odd-index 16-bit lanes are still all 0 after this. As such, despite
+     * that the following steps use 32-bit lanes, the value of f1 is unaffected.
+     */
+    /* check-magic: on */
     f1 = _mm256_mulhi_epu16(f1, v);
     f1 = _mm256_mulhrs_epi16(f1, shift);
```
hanno-becker, line 95: This could benefit from a comment. We're (ab)using a high-multiply as a rounding right shift here. One should note the peculiarity that this is rounding division by 2^15, not 2^16. 

Also, one could leave a note or open an issue investigating if this is really better than just doing an add+shift in 16-bit arithmetic.

dev/x86_64/src/poly_decompose_88_avx2.c
```diff
@@ -38,31 +38,74 @@ void mld_poly_decompose_88_avx2(__m256i *a1, __m256i *a0, const __m256i *a)
...
+     * Compute f1 = round-(f1' / B) ≈ round(f1' * 11275 / 2^24). This is exact
+     * for 0 <= f1' < 2^16. Note that half is rounded down since 11275 / 2^24 ≲
+     * 1 / 1488.
+     *
+     * The odd-index 16-bit lanes are still all 0 after this. As such, despite
+     * that the following steps use 32-bit lanes, the value of f1 is unaffected.
+     */
+    /* check-magic: on */
     f1 = _mm256_mulhi_epu16(f1, v);
     f1 = _mm256_mulhrs_epi16(f1, shift);
```
hanno-becker, line 96: Maybe mention that the intermediate value for `f1` is < 2^15, so epi15 is an unsigned multiplication?

dev/x86_64/src/poly_decompose_88_avx2.c
```diff
@@ -38,31 +38,74 @@ void mld_poly_decompose_88_avx2(__m256i *a1, __m256i *a0, const __m256i *a)
...
     f1 = _mm256_srli_epi32(f1, 7);
+    /*
+     * range: 0 <= f1' <= (Q-1)/128 = 44B
+     *
+     * Also, f1' <= (Q-1)/128 = 2^16 - 2^6 < 2^16 ensures that the odd-index
+     * 16-bit lanes are all 0, so no bits will be dropped in the input of the
+     * _mm256_mulhi_epu16() below.
+     */
+
+    /* check-magic: off */
```
hanno-becker: Same Q as above: Why is this needed? Didn't you explain the relevant constants with check-magic annotations already?

hanno-becker (changes requested):
Overall LGTM and really helps understanding the code, thank you @jammychiou1. A few smaller change requests, see comments.

dev/x86_64/src/poly_decompose_32_avx2.c
jammychiou1: It's due to how the check-magic script works currently: Only one "explained magic" is kept at any time, so the later `check-magic: 4092 == ...` overwrites the earlier `check-magic: 1025 == ...`.

Maybe we should look into fixing it in the future ...

dev/x86_64/src/poly_decompose_32_avx2.c
jammychiou1: I see what you mean. I'll reword it. Thanks!

dev/x86_64/src/poly_decompose_32_avx2.c
jammychiou1: I opened #654.

dev/x86_64/src/poly_decompose_32_avx2.c
jammychiou1: Good point. I also opened #655.

dev/x86_64/src/poly_decompose_88_avx2.c
jammychiou1: Yes, that's indeed worth pointing out. Thanks!

dev/x86_64/src/poly_decompose_32_avx2.c
jammychiou1: Will fix

dev/x86_64/src/poly_decompose_88_avx2.c
jammychiou1: Answered above.

dev/x86_64/src/poly_decompose_32_avx2.c
jammychiou1: I just thought of a (hopefully intuitive) way to explain this. If we accept that round(f / (128B)) = round(floor(f/128) / B) for any even B, then the desired equality is just its "mirrored image w.r.t. f=0". What do you think? It this perhaps too abstract/handwavy?

jammychiou1: The spelled out proof would look like this:
```
round-(f / (128B))
 = -round(-f / (128B))
 = -round(floor((-f)/128) / B)
 = -round(-ceil(f/128) / B) 
 = round-(ceil(f/128) / B)
```

hanno-becker: Was thinking of something like this:
```
Since changing f to align-up(f, 128) can move f onto but not across
a rounding boundary for division by 128*B, and round- rounds down
on the boundary, we have

round-(f / (128B)) = round-(align-up(f, 128) / (128B))
                   = round-((align-up(f, 128) / 128) / B)
                   = round-(ceil(f / 128) / B)
```

jammychiou1: Thank you for the suggestion! I've included them in the new patch.

jammychiou1:
Thank you @hanno-becker for your suggestions (and your wonderful work on #659)!

Please let me know if there's more things to change. If not, I'll clean up the commit history to prepare for the merge into main.

mldsa/src/rounding.h
```diff
@@ -93,20 +93,46 @@ __contract__(
...
 )
 {
+  /*
+   * The goal is to compute f1 = round-(f / (2*GAMMA2)), which can be computed
+   * alternatively as round-(f / (128B)) = round-(ceil(f / 128) / B) where
+   * B = 2*GAMMA2 / 128. Here round-() denotes "round half down".
+   *
+   * The equality round-(f / (128B)) = round-(ceil(f / 128) / B) can deduced
+   * as follows. Since changing f to align-up(f, 128) can move f onto but not
+   * across a rounding boundary for division by 128*B, and round- rounds down
```
hanno-becker: It may be worth noting that we need `B` to be even here (because the rounding boundaries for 128*B are the odd multiples of 128*B/2=64*B; which is a multiple of 128 iff B is even). I think you had this before, apologies for not including it in this explanation to begin with.

jammychiou1: Ah yes. I wanted to add it, but forgot to do it in the end. Thanks for the catch!

jammychiou1: Fixed

Issue #601, opened 2025-11-04
Port compile-time feature detection


mkannwischer:
https://github.com/pq-code-package/mlkem-native/pull/1146

PR #632, opened 2025-11-06
Makefile: track environment variable changes

L-series:
This PR adds the functionality in which buils will now also be retriggered by changes to environment vairables.

### DIFF
--- a/test/mk/config.mk
+++ b/test/mk/config.mk
@@ -80,10 +80,18 @@ endif
 # Include retained variables #
 ##############################
 
+RETAINED_VARS := CROSS_PREFIX CYCLES OPT AUTO
+
+# Capture values of environment variables before setting defaults,
+# this ensures we can detect when they change in order to trigger a rebuild.
+define CAPTURE_VAR
+$(1)_FROM_ENV := $$($(1))
+endef
+$(foreach var,$(RETAINED_VARS),$(eval $(call CAPTURE_VAR,$(var))))
+
 AUTO ?= 1
 CYCLES ?=
 OPT ?= 1
-RETAINED_VARS := CROSS_PREFIX CYCLES OPT AUTO
 
 ifeq ($(AUTO),1)
 include test/mk/auto.mk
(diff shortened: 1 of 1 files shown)


L-series:
Ahh! I've pushed an older version of the commit. Let me fix this now.

test/mk/config.mk
```diff
@@ -80,10 +80,19 @@ endif
...
+$(1)_FROM_ENV := $$($(1))
+endef
+$(foreach var,$(RETAINED_VARS),$(eval $(call CAPTURE_VAR,$(var))))
+
 AUTO ?= 1
 CYCLES ?=
 OPT ?= 1
-RETAINED_VARS := CROSS_PREFIX CYCLES OPT AUTO
 
 ifeq ($(AUTO),1)
```
hanno-becker, line 96: Note: We have a divergence here between mldsa-native and mlkem-native. In mlkem-native, auto.mk is unconditionally included from Makefile after config.mk. And within auto.mk, we check for AUTO.

As far as I see (but I'd prefer another check, @mkannwischer @L-series), things work as intended even with the include at this point, but we should align this between the repositories.

mkannwischer: - I think this will be consolidated when we port https://github.com/pq-code-package/mlkem-native/pull/1146

- There is an issue for that #601

L-series: Would it make sense to unify the the two config.mk as of now in preparation for the addition of compiler/host feature detection? It doesn't seem to me like any functionalities would break.  Please let me know if I am missing something.

mkannwischer: I think it's easier to port the entire PR.
Unless there is a problem in this PR, let's merge this one first and then do the port in a follow-up PR. Do you want to give that a try @L-series?

mkannwischer: If you would like to contribute more, we should add you as a contributor so you can do PRs from a local branches, so we don't have to use the hacky workaround #633. 

(Context: Some part of the CI only works for PRs from local branches as it needs to authenticate to AWS EC2 to spin up machines).

L-series: Gotcha, that's what I figured was happening with the few failing tests. I'd love to continue contributing so that sounds great. I'll assign myself the ticket https://github.com/pq-code-package/mldsa-native/issues/601 and start working on it shortly.

mkannwischer: Cool, thanks! I opened https://github.com/pq-code-package/tsc/pull/190. Once that is merged you can assign yourself. 

Could you point me to where you saw failing tests? The tests should be skipped not fail - if they fail that's likely a mistake in our CI workflows.

L-series: That was a typo on my end. You're correct it should have read skipped not failed!

PR #635, opened 2025-11-07
Native FIPS202: Remove X2 API (align with mlkem-native)

willieyz:
- This PR is ported from https://github.com/pq-code-package/mlkem-native/pull/1152
  - The purpose of this change is to keep mldsa-native aligned with mlkem-native.
  - In mlkem-native, initially 3 Keccak APIs were added: x1, x2, x4. The reasoning for this initial API design was that on some platforms x2 is optimal and using x4 provides no additional benefit. However, it later turned out that the additional flexibility only provides marginal performance improvements in ML-KEM. Ultimately, only the x1 and x4 were used in the ML-KEM code base, so the x2 API is not useful and can safely be removed. The same applies to ML-DSA.

This PR change following: 
- Change MLD_USE_FIPS202_X2_NATIVE to MLD_USE_FIPS202_X4_NATIVE in x2_v84a.h
- Implement mld_keccak_f1600_x4_native using two x2 native calls
- Remove x2-specific fallback logic in keccakf1600x4_permute
- Clean up x2 function declaration from API header

- Resolves https://github.com/pq-code-package/mldsa-native/issues/638


mkannwischer (approved):
Thanks @willieyz. LGTM.
I compared the files to mlkem-native, and it perfectly aligns.

PR #640, opened 2025-11-07
HOL-Light: Add HOL-Light proof framework and x86 AVX2 NTT proof

jakemas:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/338
- Resolves https://github.com/pq-code-package/mldsa-native/issues/646
- Resolves https://github.com/pq-code-package/mldsa-native/issues/645
- Resolves https://github.com/pq-code-package/mldsa-native/issues/647

This PR includes:

- An update to s2n-bignum version to support proofs
- Addition of `list_proofs.sh`, `build-proof.sh`, `dump_bytecode.ml` utilities
- Addition of `mldsa-specs.ml` and `mldsa-utils.ml` to support HOL-Light proofs
- Addition of first HOL-Light proof for ML-DSA x86 AVX2 NTT in `mldsa_ntt.ml`
- Addition of `hol_light` to `scripts/tests` to wrap and run proofs
- READMEs for usage descriptions
- Autogen updated to use `dump_bytecode.ml` by the argument `--update-hol-light-bytecode` and `--dry-run` to check/update
- Autogen HOL-Light assembly files `gen_hol_light_asm_file` and `gen_hol_light_asm`
- 3 New CI jobs for HOL-light proofs

See the added [proofs/hol_light/x86_64/README.md](https://github.com/pq-code-package/mldsa-native/blob/fd1787ad8c412b8f651d01433689d43d83f47c4e/proofs/hol_light/x86_64/README.md) for details. 

Running `make -C proofs/hol_light/x86_64` produces an `mldsa_ntt.correct` file that ends in the following:
```
Running time: 18147.000000 sec, Start unixtime: 1762547981.000000, End unixtime: 1762566128.000000
```
(~5hour on my test EC2 m5.2xlarge, Successful in 249m in CI).

#### Discussion
There is an issue where the x86 bytecode generated on clang ARM Mac is different to that when compiled on a gcc Linux. The clang compiler performs some optimization of instructions by switching the source operands of `VPADDD` etc. to get a shorter encoding. This means the `define_assert_from_elf` check will fail when compiled on clang ARM Mac, as the proof is set with the machine code as compiled on gcc Linux (I do have proofs for both). We will have the same issue in mlkem-native when the x86 proofs arrive. So looking for opinions on (1) being okay with the x86 proofs not working on clang ARM Mac environments (2) fix the selective swapping of source operands to `VPADDD` to "optimise" the instructions down to the shorter size on gcc Linux, so both compile consistently (3) adding both proof types... likely all but (1) is out of scope of this PR -- just keeping notes.

### DIFF
--- a/.github/workflows/hol_light.yml
+++ b/.github/workflows/hol_light.yml
@@ -1,0 +1,114 @@
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+
+name: HOL-Light
+permissions:
+  contents: read
+on:
+  push:
+    branches: ["main"]
+    paths:
+     - '.github/workflows/hol_light.yml'
+     - 'proofs/hol_light/x86_64/Makefile'
+     - 'proofs/hol_light/x86_64/**/*.S'
+     - 'proofs/hol_light/x86_64/**/*.ml'
+     - 'nix/hol_light/*'
+     - 'nix/s2n_bignum/*'
+  pull_request:
+    branches: ["main"]
+    paths:
+      - '.github/workflows/hol_light.yml'
+      - 'proofs/hol_light/x86_64/Makefile'
+      - 'proofs/hol_light/x86_64/**/*.S'
+      - 'proofs/hol_light/x86_64/**/*.ml'
+      - 'nix/hol_light/*'
+      - 'nix/s2n_bignum/*'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  # The proofs also check that the byte code is up to date,
+  # but we use this as a fast path to not even start the proofs
+  # if the byte code needs updating.
+  hol_light_bytecode:
+    name: HOL-Light bytecode check
+    runs-on: pqcp-x64
+    if: github.repository_owner == 'pq-code-package' && !github.event.pull_request.head.repo.fork
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          fetch-depth: 0
+      - uses: ./.github/actions/setup-shell
+        with:
+          gh_token: ${{ secrets.GITHUB_TOKEN }}
+          nix-shell: 'hol_light'
+          script: |
+            autogen --update-hol-light-bytecode --dry-run
+  hol_light_interactive:
+    name: HOL-Light interactive shell test
+    runs-on: pqcp-x64
+    needs: [ hol_light_bytecode ]
+    if: github.repository_owner == 'pq-code-package' && !github.event.pull_request.head.repo.fork
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          fetch-depth: 0
+      - uses: ./.github/actions/setup-shell
+        with:
+          gh_token: ${{ secrets.GITHUB_TOKEN }}
+          nix-shell: 'hol_light'
+          script: |
+            # Load base infrastructure and specs to validate HOL-Light environment
+            # When we have smaller/faster proofs, we can run them here instead:
+            # make -C proofs/hol_light/x86_64 mldsa/mldsa_ntt.o
+            # echo 'needs "proofs/mldsa_ntt.ml";;' | hol.sh
+            echo 'needs "x86/proofs/base.ml";; needs "proofs/mldsa_specs.ml";; #quit;;' | hol.sh
+  hol_light_proofs:
+    needs: [ hol_light_bytecode ]
+    strategy:
+      fail-fast: false
+      matrix:
+        proof:
+          # Dependencies on {name}.{S,ml} are implicit
+          - name: mldsa_ntt
+            needs: ["mldsa_specs.ml", "mldsa_utils.ml"]
+    name: HOL Light proof for ${{ matrix.proof.name }}.S
+    runs-on: pqcp-x64
+    if: github.repository_owner == 'pq-code-package' && !github.event.pull_request.head.repo.fork
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          fetch-depth: 0
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@24d32ffd492484c1d75e0c0b894501ddb9d30d62 # v47.0.0
+      - name: Check if dependencies changed
+        id: check_run
+        shell: bash
+        run: |
+          run_needed=0
+          changed_files="${{ steps.changed-files.outputs.all_changed_files }}"
+          dependencies="${{ join(matrix.proof.needs, ' ') }} ${{ format('{0}.S {0}.ml', matrix.proof.name) }}"
+          for changed in $changed_files; do
+            for needs in $dependencies; do
+               if [[ "$changed" == *"$needs" ]]; then
+                 run_needed=1
+               fi
+            done
+          done
+
+          # Always re-run upon change to nix files for HOL-Light
+          if [[ "$changed_files" == *"nix/"* ]] || [[ "$changed_files" == *"hol_light.yml"* ]] || [[ "$changed_files" == *"flake"* ]] || [[ "$changed_files" == *"proofs/hol_light/x86_64/Makefile"* ]]; then
+            run_needed=1
+          fi
+
+          echo "run_needed=${run_needed}" >> $GITHUB_OUTPUT
+      - uses: ./.github/actions/setup-shell
+        if: |
+          steps.check_run.outputs.run_needed == '1'
+        with:
+          gh_token: ${{ secrets.GITHUB_TOKEN }}
+          nix-shell: 'hol_light'
+          script: |
+            tests hol_light -p ${{ matrix.proof.name }} --verbose
--- a/proofs/README.md
+++ b/proofs/README.md
@@ -7,3 +7,7 @@ This directory contains material related to the formal verification of the sourc
 ## C verification: CBMC
 
 We use the [C Bounded Model Checker (CBMC)](https://github.com/diffblue/cbmc) to show the absence of various classes of undefined behaviour in the mldsa-native C source, including out of bounds memory accesses and integer overflows. See [proofs/cbmc](cbmc), or the [proof_guide](https://github.com/pq-code-package/mlkem-native/blob/main/proofs/cbmc/proof_guide.md).
+
+## Assembly verification: HOL-Light
+
+We use the [HOL-Light](https://github.com/jrh13/hol-light) interactive theorem prover alongside the verification infrastructure from [s2n-bignum](https://github.com/awslabs/s2n-bignum) to show the functional correctness of highly optimized assembly routines in mlkem-native at the object-code level. See [proofs/hol_light/x86_64](hol_light/x86_64).
--- a/proofs/hol_light/x86_64/README.md
+++ b/proofs/hol_light/x86_64/README.md
@@ -1,0 +1,108 @@
+[//]: # (SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT-0)
+
+# HOL Light functional correctness proofs
+
+This directory contains functional correctness proofs for some of the x86_64 assembly routines
+used in mldsa-native. The proofs were largely developed by John Harrison and Jake Massimo,
+and are written in the [HOL Light](https://hol-light.github.io/) theorem
+prover, utilizing the assembly verification infrastructure from [s2n-bignum](https://github.com/awslabs/s2n-bignum).
+
+Each function is proved in a separate `.ml` file in [proofs/](proofs). Each file
+contains the byte code being verified, as well as the specification that is being
+proved.
+
+## Primer
+
+Proofs are 'post-hoc' in the sense that HOL-Light/s2n-bignum operate on the final object code. In particular, the means by which the code was generated need not be trusted.
+
+Specifications are essentially [Hoare triples](https://en.wikipedia.org/wiki/Hoare_logic), with the noteworthy difference that the program is implicit as the content of memory at the PC; which is asserted to
+be the code under verification as part of the precondition. For example, the following is the specification of the `mldsa_ntt` function:
+
+```ocaml
+ (* For all (abbreviated by `!` in HOL):
+    - a: Polynomial coefficients pointer
+    - zetas: NTT constants pointer
+    - x: Original polynomial coefficients
+    - pc: Current value of Program Counter (PC)
+    - stackpointer: Stack pointer
+    - returnaddress: Return address on the stack *)
+`!a zetas x pc stackpointer returnaddress.
+    (* Alignment and non-overlapping requirements *)
+    aligned 32 a /\
+    aligned 32 zetas /\
+    nonoverlapping (word pc,LENGTH mldsa_ntt_mc) (a, 1024) /\
+    nonoverlapping (word pc,LENGTH mldsa_ntt_mc) (zetas, 2496) /\
+    nonoverlapping (a, 1024) (zetas, 2496) /\
+    nonoverlapping (stackpointer,8) (a, 1024) /\
+    nonoverlapping (stackpointer,8) (zetas, 2496)
+    ==> ensures x86
+      (* Precondition *)
+      (\s. (* The memory at the current PC is the byte-code of mldsa_ntt() *)
+        bytes_loaded s (word pc) mldsa_ntt_mc /\
+        read RIP s = word pc /\
+        read RSP s = stackpointer /\
+        (* The return address is on the stack *)
+        read (memory :> bytes64 stackpointer) s = returnaddress /\
+        (* Arguments are passed via C calling convention *)
+        C_ARGUMENTS [a; zetas] s /\
+        (* NTT constants are properly loaded *)
+        wordlist_from_memory(zetas,624) s = MAP (iword: int -> 32 word) mldsa_complete_qdata /\
+        (* Input bounds checking *)
+        (!i. i < 256 ==> abs(ival(x i)) <= &8380416) /\
+        (* Give a name to the memory contents at the source pointer *)
+        !i. i < 256
+            ==> read(memory :> bytes32(word_add a (word(4 * i)))) s = x i)
+      (* Postcondition: Eventually we reach a state where ... *)
+      (\s.
+        (* The PC is the return address *)
+        read RIP s = returnaddress /\
+        (* Stack pointer is adjusted *)
+        read RSP s = word_add stackpointer (word 8) /\
+        (* The integers represented by the final memory contents
+         * are congruent to the ML-DSA forward NTT transformation
+         * of the original coefficients, modulo 8380417, with proper bounds *)
+        !i. i < 256
+            ==> let zi = read(memory :> bytes32(word_add a (word(4 * i)))) s in
+                (ival zi == mldsa_forward_ntt (ival o x) i) (mod &8380417) /\
+                abs(ival zi) <= &42035261)
+      (* Footprint: The program may modify (only) the ABI permitted registers
+       * and flags, stack pointer, and the memory contents at the source pointer. *)
+      (MAYCHANGE [RSP] ,, MAYCHANGE_REGS_AND_FLAGS_PERMITTED_BY_ABI ,,
+       MAYCHANGE [memory :> bytes(a,1024)])`
+```
+
+## Platform Compatibility
+
+These x86_64 assembly proofs are expected to work on the following platforms:
+
+- **Linux x86_64**: Fully supported (native build and execution)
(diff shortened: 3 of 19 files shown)


mkannwischer (changes requested):
Thanks @jakemas - this is awesome progress!

I tried to run the proofs on a c8g.4x-large in the `nix develop .#hol_light` shell, but 
it does not work as `x86_64-linux-gnu-objdump` and `x86_64-linux-gnu-as` are not installed. 
We can probably fix cross compilation in a follow-up PR -- running it on x86 now. 

```
ubuntu@ip-172-31-34-86:~/mldsa-native tests hol_light -p mldsa_ntt
INFO  > HOL_LIGHT (1/1)    None        (native):         Starting HOL-Light proof for mldsa_ntt
ERROR > HOL_LIGHT (1/1)    None        (native):            FAILED (after 0s)
ERROR > HOL_LIGHT (1/1)    None        (native):         sh: line 1: x86_64-linux-gnu-as: command not found
sh: line 1: x86_64-linux-gnu-objdump: command not found
sh: line 1: x86_64-linux-gnu-as: command not found
make: *** [Makefile:70: mldsa/mldsa_ntt.o] Error 127

1 tests FAILED
* HOL-Light proof for mldsa_ntt
```

```
ubuntu@ip-172-31-34-86:~/mldsa-native$ make -C proofs/hol_light/x86_64 
make: Entering directory '/home/ubuntu/mldsa-native/proofs/hol_light/x86_64'
Preparing mldsa/mldsa_ntt.o ...
sh: line 1: x86_64-linux-gnu-as: command not found
AS: 
sh: line 1: x86_64-linux-gnu-objdump: command not found
OBJDUMP: 
[ -d mldsa ] || mkdir -p mldsa
cat mldsa/mldsa_ntt.S | gcc -E -xassembler-with-cpp - | tr ';' '\n' | x86_64-linux-gnu-as -o mldsa/mldsa_ntt.o -
sh: line 1: x86_64-linux-gnu-as: command not found
make: *** [Makefile:70: mldsa/mldsa_ntt.o] Error 127
make: Leaving directory '/home/ubuntu/mldsa-native/proofs/hol_light/x86_64'

````

**Review by mkannwischer — APPROVED** [review]
Thanks @jakemas.

I verified that the proof runs okay.
I tested on a m4.4xlarge machine and the proofs completed after 20393s. 

I tried to understand the main theorem that is being proven - the bounds look good and are indeed much tighter than what we require.

It currently only works on x86, but we can address that in a follow up. 
Here is a list of follow-up issues that we should tackle:
  - https://github.com/pq-code-package/mldsa-native/issues/643
- https://github.com/pq-code-package/mldsa-native/issues/644
- https://github.com/pq-code-package/mldsa-native/issues/645
- https://github.com/pq-code-package/mldsa-native/issues/646
- https://github.com/pq-code-package/mldsa-native/issues/647
- https://github.com/pq-code-package/mldsa-native/issues/648

@hanno-becker, WDYT?

File: proofs/README.md [hunk 1]
```diff
@@ -7,3 +7,7 @@ This directory contains material related to the formal verification of the sourc
 ## C verification: CBMC
 
 We use the [C Bounded Model Checker (CBMC)](https://github.com/diffblue/cbmc) to show the absence of various classes of undefined behaviour in the mldsa-native C source, including out of bounds memory accesses and integer overflows. See [proofs/cbmc](cbmc), or the [proof_guide](https://github.com/pq-code-package/mlkem-native/blob/main/proofs/cbmc/proof_guide.md).
+
+## Assembly verification: HOL-Light
+
+We use the [HOL-Light](https://github.com/jrh13/hol-light) interactive theorem prover alongside the verification infrastructure from [s2n-bignum](https://github.com/awslabs/s2n-bignum) to show the functional correctness of various highly optimized assembly routines in mlkem-native at the object-code level. See [proofs/hol_light/x86_64](hol_light/x86_64).
```
hanno-becker: "to show the functional correctness of various highly optimized assembly" doesn't yet fit given we have 'only' NTT. Also s/mlkem/mldsa.

proofs/hol_light/x86_64/README.md
```diff
@@ -0,0 +1,99 @@
+[//]: # (SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT-0)
+
+# HOL Light functional correctness proofs
+
+This directory contains functional correctness proofs for all x86_64 assembly routines
```
hanno-becker: This needs adjusting.

jakemas: Ok fixed!

For reference, when making the other READMEs feedback was that we didn't want to keep editing this over time, so to try and make it look like final state.

proofs/README.md
jakemas: fixed!

.github/workflows/hol_light.yml
```diff
@@ -0,0 +1,115 @@
...
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          fetch-depth: 0
+      - uses: ./.github/actions/setup-shell
+        with:
+          gh_token: ${{ secrets.GITHUB_TOKEN }}
+          nix-shell: 'hol_light'
+          script: |
+            make -C proofs/hol_light/x86_64 mldsa/mldsa_ntt.o
```
jakemas: When we have a quicker proof than the full NTT we can add it here, mlkem-native uses `mlkem_poly_tobytes` for example.

scripts/autogen
```diff
@@ -2363,11 +2364,86 @@ def gen_test_configs():
         gen_test_config(config_spec["path"], config_spec, default_config)
 
 
+def extract_bytecode_from_output(output_text):
```
mkannwischer: Please place those functions in the same place as they are in the mlkem-native autogen script, i.e., below`gen_citations` - that makes syncing these files much easier.

scripts/autogen
```diff
@@ -2363,11 +2364,86 @@ def gen_test_configs():
...
+    # Replace content between markers
+    pattern = rf"{re.escape(start_marker)}.*?{re.escape(end_marker)}"
+    replacement = f"{start_marker}\n{bytecode}\n{end_marker}"
+
+    updated_content = re.sub(pattern, replacement, content, flags=re.DOTALL)
+
+    update_file(filepath, updated_content)
+
+
+def update_hol_light_bytecode(dry_run):
```
mkannwischer: There is no dry-run anymore. Remove.

scripts/autogen
```diff
@@ -2423,6 +2499,10 @@ def _main():
     gen_test_configs()
     high_level_status("Generated test configs")
 
+    if args.update_hol_light_bytecode:
+        update_hol_light_bytecode(args.dry_run)
+        high_level_status("Updated HOL Light bytecode")
+
```
mkannwischer: Please place this below `high_level_status("Completed final backend synchronization")`, to align with mlkem-native.

mkannwischer (changes requested):
Thanks @jakemas. The CI looks good to me. 
Three minor comments on autogen above.

Can you also autogenerate mldsa_ntt.S please (see gen_hol_light_asm in mlkem-native)?

scripts/autogen
jakemas: Done!

scripts/autogen
jakemas: Done!

scripts/autogen
jakemas: Done!

scripts/autogen
hanno-becker: @jakemas, general note: I think we try to let comments be resolved by the person who raised them, so we can be sure they're happy. So let's leave this to @mkannwischer.

jakemas: Sorry, I am using a script/browser tool that extends CI so I can see all jobs without scrolling, however, as the discussion on this PR is growing larger and larger is it messing with the script, and jumping my page all over. I was trying to reduce the amount of content on the page at once, didn't mean to dismiss.

jargh:
That is, I'd instinctively vote for option (2) abiove if it isn't too troublesome

> (2) fix the selective swapping of source operands to VPADDD to "optimise" the instructions down to the shorter size on gcc Linux, so both compile consistently

dev/x86_64/src/ntt.S
```diff
@@ -267,6 +267,7 @@ vmovdqa		%ymm11,256*\off+224(%rdi)
 .balign 4
 .global MLD_ASM_NAMESPACE(ntt_avx2)
 MLD_ASM_FN_SYMBOL(ntt_avx2)
+endbr64
```
hanno-becker: We don't yet use those branch target instructions in mldsa-native. I would prefer to keep things uniform and introduce it all at once, and not add it here for now.

mldsa/src/native/x86_64/src/ntt.S
```diff
@@ -34,6 +34,7 @@
 MLD_ASM_FN_SYMBOL(ntt_avx2)
 
         .cfi_startproc
+        endbr64
```
hanno-becker: Same here

dev/x86_64/src/ntt.S
jakemas: If we want to autogen the asmb, I need these there, as they are required in the proof steps/simulation.

proofs/hol_light/x86_64/README.md
```diff
@@ -0,0 +1,99 @@
...
+make -C proofs/hol_light/x86_64
+```
+
+will build and run the proofs. Note that this make take hours even on powerful machines.
+
+For convenience, you can also use `tests hol_light` which wraps the `make` invocation above; see `tests hol_light --help`.
+
+## What is covered?
+
+All x86_64 assembly routines used in mldsa-native are covered. Currently this includes:
```
hanno-becker: This needs adjusting.

proofs/hol_light/x86_64/proofs/mldsa_specs.ml
```diff
@@ -0,0 +1,597 @@
...
+(* ------------------------------------------------------------------------- *)
+
+let UNIFY_INTCONG_RULE th1 th2 =
+  let p1 = dest_intconst (rand(rand(lhand(concl th1))))
+  and p2 = dest_intconst (rand(rand(lhand(concl th2)))) in
+  let d = gcd_num p1 p2 in
+  CONJ (WEAKEN_INTCONG_RULE d th1) (WEAKEN_INTCONG_RULE d th2);;
+
+(* ------------------------------------------------------------------------- *)
+(* Process list of ineqequality into standard congbounds for atomic terms.   *)
```
hanno-becker: Nit: `s/ineqequality/inequality/`

dev/x86_64/src/ntt.S
jakemas: I'd have to rewrite the proof without them, only to rewrite it when we add them back? Seems like unnecessary churn for a style of integration choice?

proofs/hol_light/x86_64/proofs/mldsa_specs.ml
jargh: Probably not the only case where a spellchecker would be worthwhile...

jakemas: fixed

proofs/hol_light/x86_64/README.md
jakemas: fixed

proofs/hol_light/x86_64/proofs/mldsa_ntt.ml
```diff
@@ -0,0 +1,4741 @@
...
+   -- &2131021878; -- &247357819; &1518161567; -- &86965173; &1708872713; &1787797779; &1638590967; -- &120646188;
+   -- &1669960606; -- &916321552; &1155548552; &2143745726; &1210558298; -- &1261461890; -- &318346816; &628664287;
+   -- &1729304568; &1422575624; &1424130038; -- &1185330464; &235321234; &168022240; &1206536194; &985155484;
+   -- &894060583; -- &898413; -- &1363460238; -- &605900043; &2027833504; &14253662; &1014493059; &863641633;
+   &1819892093; &2124962073; -- &1223601433; -- &1920467227; -- &1637785316; -- &1536588520; &694382729;
+   &235104446; -- &1045062172; &831969619; -- &300448763; &756955444; -- &260312805; &1554794072; &1339088280;
+   -- &2040058690; -- &853476187; -- &2047270596; -- &1723816713; -- &1591599803; -- &440824168; &1119856484;
+   &1544891539; &155290192; -- &973777462; &991903578; &912367099; -- &44694137; &1176904444; -- &421552614;
+   -- &818371958; &1747917558; -- &325927722; &908452108; &1851023419; -- &1176751719; -- &1354528380; -- &72690498;
+   -- &3
```
hanno-becker, line 4526: We should aim to rewrite the specs in a way that does not hardcode any code-length constants; see e.g. https://github.com/pq-code-package/mlkem-native/blob/main/proofs/hol_light/arm/proofs/mlkem_ntt.ml#L539 for how we do it in mlkem-natie. This can be a follow-up. @jakemas Can you create an issue?

jakemas: Ok done https://github.com/pq-code-package/mldsa-native/issues/672

hanno-becker, line 4547: We need to add a corresponding CBMC specification to `dev/x86_64/src/arith_native_x86_64.h`, matching the relevant part of the HOL-Light spec exactly (in particular, using exactly the same bounds). See e.g. https://github.com/pq-code-package/mlkem-native/blob/main/dev/aarch64_clean/src/arith_native_aarch64.h#L35 for how it's done for the AArch64 NTT in mlkem-native.

Then, we need to add (should be trivial) CBMC proof that the CBMC spec corresponding to the HOL-Light spec implies the spec we assume in the frontend.

This can be a follow-up, I opened https://github.com/pq-code-package/mldsa-native/issues/673

proofs/hol_light/x86_64/README.md
```diff
@@ -0,0 +1,99 @@
...
+from mldsa-native's base directory. Then
+
+```bash
+make -C proofs/hol_light/x86_64
+```
+
+will build and run the proofs. Note that this make take hours even on powerful machines.
+
+For convenience, you can also use `tests hol_light` which wraps the `make` invocation above; see `tests hol_light --help`.
+
```
hanno-becker, line 100: Can we document on which platforms this is expected to work on? Linux (x86/AArch64) and x86 Mac? The notably exception, Arm Mac, should ideally be mentioned as well, with a reference to the issue describing the problem.

dev/x86_64/src/ntt.S
hanno-becker: On second sight, we already integrate endbr64 into MLD_ASM_FN_SYMBOL. However, since the HOL-Light asm is generated with `-fcf-protection=none`, it's not emitted in the HOL-Light assembly.

We can't keep an explicit `endbrb64` in the clean code, since that contains _two_ endbr64 now. However, we should be able to keep your existing proof if we change `autogen` to build the HOL-Light ASM using `-fcf-protection=none`. Strictly speaking, one would need _two_ HOL-Light proofs here, but I'd be OK starting with 'just' the version including the `endbr64`.

So, in summary, we have two options:
* Remove the `endbr64` from the HOL-Light proof and keep `autogen` as-is
* Keep the proof as-is but adjust `autogen` for the HOL-Light ASM so it emits the `endbr64`.

.github/workflows/hol_light.yml
```diff
@@ -0,0 +1,118 @@
...
+on:
+  push:
+    branches: ["main"]
+    paths:
+     - '.github/workflows/hol_light.yml'
+     - 'proofs/hol_light/x86_64/Makefile'
+     - 'proofs/hol_light/x86_64/**/*.S'
+     - 'proofs/hol_light/x86_64/**/*.ml'
+     - 'flake.nix'
+     - 'flake.lock'
```
hanno-becker: I think we need to be pragmatic here and only re-run the proofs if something specific to HOL-Light changes. In mlkem-native, it's borderline acceptable to re-run on every change to the `flake.nix`, but here I think it's not. @mkannwischer WDYT?

hanno-becker: I think we need to be pragmatic here and only re-run the proofs if something specific to HOL-Light changes. In mlkem-native, it's borderline acceptable to re-run on every change to the `flake.nix`, but here I think it's not. @mkannwischer WDYT?

.github/workflows/hol_light.yml
```diff
@@ -0,0 +1,118 @@
...
+     - 'nix/s2n_bignum/*'
+  pull_request:
+    branches: ["main"]
+    paths:
+      - '.github/workflows/hol_light.yml'
+      - 'proofs/hol_light/x86_64/Makefile'
+      - 'proofs/hol_light/x86_64/**/*.S'
+      - 'proofs/hol_light/x86_64/**/*.ml'
+      - 'flake.nix'
+      - 'flake.lock'
```
hanno-becker: Same here

proofs/hol_light/x86_64/README.md
jakemas: Added 

```
## Platform Compatibility

These x86_64 assembly proofs are expected to work on the following platforms:

- **Linux x86_64**: Fully supported (native build and execution)
- **Linux AArch64**: Not currently supported. Requires x86_64 cross-compilation tools (`x86_64-linux-gnu-as` and `x86_64-linux-gnu-objdump`).
- **macOS x86_64**: Fully supported (native build and execution)
- **macOS ARM (Apple Silicon)**: Not currently supported. Differences in assembler behavior between clang-based (macOS) and gcc-based (Linux) toolchains can cause verification mismatches, particularly with regard to argument ordering optimizations for commutative instructions.
```

dev/x86_64/src/ntt.S
jakemas: Thank you, nice find, and discussed offline. 
Gone for the option: 
>Keep the proof as-is but adjust autogen for the HOL-Light ASM so it emits the endbr64.

Added to PR see https://github.com/pq-code-package/mldsa-native/pull/640/files#diff-8526e108afc8c311242e66c36c226b093f9d3e29c460133acc5558bc8f64fe38R1633

proofs/hol_light/x86_64/proofs/mldsa_ntt.ml
jakemas: ok , sounds good to me!

mldsa/src/native/x86_64/src/ntt.S
jakemas: now fixed

.github/workflows/hol_light.yml
mkannwischer: Agreed. We can remove `flake.nix` and flake.lock` here.

proofs/hol_light/x86_64/README.md
```diff
@@ -0,0 +1,108 @@
+[//]: # (SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT-0)
+
+# HOL Light functional correctness proofs
+
+This directory contains functional correctness proofs for x86_64 assembly routines
```
hanno-becker:
```suggestion
This directory contains functional correctness proofs for some of the x86_64 assembly routines
```

proofs/hol_light/x86_64/README.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+## Platform Compatibility
+
+These x86_64 assembly proofs are expected to work on the following platforms:
+
+- **Linux x86_64**: Fully supported (native build and execution)
+- **Linux AArch64**: Not currently supported. Requires x86_64 cross-compilation tools (`x86_64-linux-gnu-as` and `x86_64-linux-gnu-objdump`).
+- **macOS x86_64**: Fully supported (native build and execution)
+- **macOS ARM (Apple Silicon)**: Not currently supported. Differences in assembler behavior between clang-based (macOS) and gcc-based (Linux) toolchains can cause verification mismatches, particularly with regard to argument ordering optimizations for commutative instructions.
```
hanno-becker, line 81: It's only the compilation step that's problematic here, isn't it? The proof itself would work fine on Arm machines as well. We should generalize the Makefile accordingly.

.github/workflows/hol_light.yml
hanno-becker: I'll update the PR.

proofs/hol_light/x86_64/README.md
hanno-becker: I'll check if adding `toolchain_x86_64` to the `hol_light` shell is enough here.

proofs/hol_light/x86_64/README.md
hanno-becker: Will amend.

hanno-becker: Done

.github/workflows/hol_light.yml
hanno-becker: Done

proofs/hol_light/x86_64/README.md
hanno-becker: This needs a bit more thought (see https://github.com/pq-code-package/mlkem-native/pull/1303) but should not block this PR.

PR #642, opened 2025-11-09
Make: add compiler and host platform feature detection

L-series:
This commit extends the Makefiles to check both compiler and host CPU
capabilities when setting architecture flags with AUTO=1.

- Adds test/mk/compiler.mk for compile-time compiler feature detection
  (AVX2, BMI2, SSE2 for x86_64; SHA3 for AArch64)
- test/mk/auto.mk detects host CPU features and only enables
  flags when both compiler AND host support them
- Adds host_info target to display feature detection status
- Adds workaround for valgrind SHA3 incompatibility by allowing
  MK_COMPILER_SUPPORTS_SHA3=0 override on command line
- Updated .github/actions/ct-test/action.yml to disable SHA3 for
  valgrind constant-time tests

Adapted from mlkem-native commit 9da1ccd668c5c956b9c367e048a3df611dd9473
introduced in pr [#1146](https://github.com/pq-code-package/mlkem-native/pull/1146)

Fixes: https://github.com/pq-code-package/mldsa-native/issues/601

Signed-off-by: Andreas Hatziiliou <andreas.hatziiliou@savoirfairelinux.com>

### DIFF
--- a/test/mk/auto.mk
+++ b/test/mk/auto.mk
@@ -2,37 +2,115 @@
 # Copyright (c) The mldsa-native project authors
 # SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
 
-# Automatically detect system architecture and set preprocessor etc accordingly
+# Automatically detect system architecture and set preprocessor flags accordingly
+# This file detects host CPU capabilities and combines them with compiler support
+# to enable optimal compilation flags.
+
+ifndef _AUTO_MK
+_AUTO_MK :=
+
+# Helper function to check if host CPU supports a feature
+# Usage: $(call check_host_feature,feature_pattern,source_command)
+define check_host_feature
+$(shell $(2) 2>/dev/null | grep -q "$(1)" && echo 1 || echo 0)
+endef
+
+# x86_64 architecture detection
+ifeq ($(ARCH),x86_64)
+
+# Host CPU feature detection for x86_64
 ifeq ($(HOST_PLATFORM),Linux-x86_64)
-ifeq ($(CROSS_PREFIX),)
-	CFLAGS += -mavx2 -mbmi2 -mpopcnt -maes
-	CFLAGS += -DMLD_FORCE_X86_64
-else ifneq ($(findstring aarch64_be, $(CROSS_PREFIX)),)
-	CFLAGS += -DMLD_FORCE_AARCH64_EB
-else ifneq ($(findstring aarch64, $(CROSS_PREFIX)),)
-	CFLAGS += -DMLD_FORCE_AARCH64
+# Linux: Use /proc/cpuinfo
+MK_HOST_SUPPORTS_AVX2 := $(call check_host_feature,avx2,cat /proc/cpuinfo)
+MK_HOST_SUPPORTS_SSE2 := $(call check_host_feature,sse2,cat /proc/cpuinfo)
+MK_HOST_SUPPORTS_BMI2 := $(call check_host_feature,bmi2,cat /proc/cpuinfo)
+else ifeq ($(HOST_PLATFORM),Darwin-x86_64)
+# macOS: Use sysctl
+MK_HOST_SUPPORTS_AVX2 := $(call check_host_feature,AVX2,sysctl -n machdep.cpu.leaf7_features)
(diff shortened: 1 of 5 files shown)


test/mk/auto.mk
```diff
@@ -2,37 +2,114 @@
...
+endif # aarch64
+
+# AArch64 Big Endian CFLAGS configuration
+ifeq ($(ARCH),aarch64_be)
+CFLAGS += -DMLD_FORCE_AARCH64_EB
+endif # aarch64_be
+
+# RISC-V 64-bit CFLAGS configuration
+ifeq ($(ARCH),riscv64)
+CFLAGS += -DMLD_FORCE_RISCV64
```
L-series, line 100: Note that in mlkem, there is an additional cflag set for rv64, namely `CFLAGS += -march=rv64gcv`. This is added as part of a larger commit 67a1217db. Should this be added now? I don't think there will be any issue with specifying the march flag form now - however, if/when the changes from the former commit to be ported it could be included as a part of that.

test/mk/compiler.mk
```diff
@@ -0,0 +1,72 @@
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
```
mkannwischer, line 3: Add copyright header here. As this was copied from mlkem-native, it should include both mlkem-native and mldsa-native: 
```
# Copyright (c) The mlkem-native project authors
# Copyright (c) The mldsa-native project authors
```

mlkem-native is missing it in a couple of places. We should likely strengthen the linting script and add it there too.

test/mk/compiler.mk
```diff
@@ -0,0 +1,72 @@
...
+#   make MK_COMPILER_SUPPORTS_BMI2=0    # Disable BMI2 detection
+#   make MK_COMPILER_SUPPORTS_SSE2=0    # Disable SSE2 detection
+
+ifndef _COMPILER_MK
+_COMPILER_MK :=
+
+# Override ARCH for cross-compilation based on CROSS_PREFIX
+ifeq ($(CROSS_PREFIX),)
+# Normalize architecture names
+ARCH := $(shell uname -m)
```
mkannwischer, line 21: When I diff this file to mlkem-native's compiler.mk I see minor whitespace and order differences. That makes syncing these files in the future unnecessarily hard. Could you align these two please?

test/mk/config.mk
mkannwischer: This line is not present in mlkem-native. Can we safely remove it?

test/mk/auto.mk
mkannwischer: Aligning it makes sense to me. Please add. That reduces the risk that we forget it when we port 67a1217db.

mkannwischer (changes requested):
Thanks @L-series - it works fine for me on Darwin-arm64 and Linux-x86_64.
I have a few cosmetical suggestions.

test/mk/config.mk
L-series: Yes, it is removed and now the file is in line with mlkem.

test/mk/compiler.mk
L-series: Done. Files are now the same (minus the missing copyright header in mlkem). I can go ahead and take a look at the linting script and add headers if needed.

test/mk/compiler.mk
L-series: Added them to this file.

test/mk/auto.mk
L-series: Perfect, line is added.

test/mk/compiler.mk
mkannwischer: Thanks. I already took care of the missing copyright headers in https://github.com/pq-code-package/mlkem-native/pull/1290. Still need to merge it. Then we can port it to mldsa-native.

mkannwischer (approved):
Thanks @L-series. LGTM. 

I took the liberty to change 2 mlkem-native into mldsa-native.

Btw, if you write 
- Fixes: https://github.com/pq-code-package/mldsa-native/issues/601

instead of

Fixes: https://github.com/pq-code-package/mldsa-native/issues/601

in the PR description it expands the issue title.

PR #659, opened 2025-11-10
check-magic: Remember all explained magic constants

hanno-becker:
Previously, scripts/check-magic would remember only the last explained magic constant, preventing, for example, the explanation of multiple magic constants ahead of a comment block referring to all of them.

Moreover, check-magic would only lazily evaluate a provided explanation when actually finding a magic value magic the LHS of the proposed explanation. In particular, a _wrong_ explanation would only be caught if, in the rest of the file under consideration, some matching magic constant would be found. This led to an ambiguous and unused explanation being overlooked in params.h.

This commit makes check-magic more general so that
- it always checks magic value explanations when they are provided, regardless of whether they are needed or not; and,
- it remembers all magic values explained so far.

Moreover, the `round` function is instrumented to fail if it is called on an odd multiple of 1/2 -- in this case, the rounding is ambiguous (do we want round-half-down or round-half-up?); this was the source of the mismatch in params.h previously.

We also add support for `intdiv(a,b)` to an integer division which we want to assert to be without residue. This can be used instead of `//` to additionally check that the division is indeed integral.

### DIFF
--- a/scripts/check-magic
+++ b/scripts/check-magic
@@ -8,9 +8,10 @@
 #
 
 import re
+import math
 import pathlib
 
-from sympy import simplify, sympify, Function
+from sympy import simplify, sympify, Function, Rational
 
 def get_c_source_files():
     return get_files("mldsa/**/*.c")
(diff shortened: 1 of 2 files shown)


scripts/check-magic
```diff
@@ -92,6 +117,12 @@ def check_magic_numbers():
             l, g = get_magic(l)
             if g is not None:
                 magic_val, magic_expr = g
+                magic_val_check = evaluate_magic(magic_expr, magic_dict)
+                if magic_val != magic_val_check:
+                    print(f"{FAIL}:{filename}:{i}: Mismatching magic annotation: {magic_val} != {magic_expr} (= {magic_val_check})")
```
jammychiou1: Line numbers in these log messages are 0-indexed, which can sometimes be a little bit annoying. I think we can change them to be 1-indexed while we're at it, just to match the editors better.

jammychiou1 (changes requested):
Very useful improvement! Works flawlessly for the decompose comments. It would be perfect if we can also adjust the line numbering in the log messages.

scripts/check-magic
hanno-becker: Good idea, fixed.

mkannwischer (approved):

Issue #602, opened 2025-11-04
Reason about bounds of `mld_fqmul` and `mld_fqscale`


mkannwischer:
While those functions come with CBMC proofs, we should still add textual reasoning about the bounds.

https://github.com/pq-code-package/mldsa-native/blob/db475c1ac5882c12eb1c1a5906458b7f2614b68c/mldsa/src/ntt.c#L45-L55

https://github.com/pq-code-package/mldsa-native/blob/db475c1ac5882c12eb1c1a5906458b7f2614b68c/mldsa/src/ntt.c#L24-L32

jammychiou1:
I noticed that the bounds in `mld_montgomery_reduce()` are also left unexplained. I think we can address it at the same time.

https://github.com/pq-code-package/mldsa-native/blob/314ad8e597b1e627e48475ec28ba16a2f382fbb0/mldsa/src/reduce.h#L51-L59

PR #667, opened 2025-11-11
Add bounds reasoning comments to AArch64 backend

### LINKED ISSUE #529: Add bounds reasoning comments to AArch64 backend
We'll need to improve the documentation of the AArch64 backend, in particular we should add bounds reasoning comments along the lines of [what is done in mlkem-native](https://github.com/pq-code-package/mlkem-native/blob/18b4bd798288f94ebb2cc6ed5e96f90b77d71f18/dev/aarch64_clean/src/ntt.S#L208).

This can later help with writing a contract for the native functions and proving them in HOL-Light.

jammychiou1:
- ~Towards resolving #529. Still need to do decompose.~
- Resolves #529

Note that the Montgomery reduction within the basemuls is changed to the positive/subtractive variant, in order to match `mld_montgomery_reduce()` and reuse the same bound (and reasoning, when we add it later in #602). The performance is expected to be identical.

### DIFF
--- a/dev/aarch64_clean/src/poly_decompose_32_asm.S
+++ b/dev/aarch64_clean/src/poly_decompose_32_asm.S
@@ -9,12 +9,15 @@
 
 // a aliased with a0
 .macro decompose32 a1, a, temp
-        // Compute a1 = round-(a / 523776) ≈ round(a * 1074791425 /
-        // 2^49), where round-() denotes "round half down". This is
-        // exact for 0 <= a < Q. Note that half is rounded down since
-        // 1074791425 / 2^49 ≲ 1 / 523776.
+        // range: 0 <= a <= Q-1 = 32*GAMMA2
+
+        // Compute a1 = round-(a / (2*GAMMA2)) = round-(a / 523776) ≈
+        // round(a * 1074791425 / 2^49), where round-() denotes "round half
+        // down". This is exact for 0 <= a < Q. Note that half is rounded down
+        // since 1074791425 / 2^49 ≲ 1 / 523776.
         sqdmulh \a1\().4s, \a\().4s, barrett_const.4s
         srshr \a1\().4s, \a1\().4s, #18
+        // range: 0 <= a1 <= 16
 
         // If a1 = 16, i.e. a > 31*GAMMA2, proceed as if a' = a - Q was
         // given instead. (For a = 31*GAMMA2 + 1 thus a' = -GAMMA2, we
@@ -25,10 +28,14 @@
 
         // Compute remainder a0
         mls \a\().4s, \a1\().4s, gamma2_2x.4s
+        // range: -GAMMA2 < a0 <= GAMMA2
+        //
+        // This holds since a1 = round-(a / (2*GAMMA2)) was computed exactly.
 
         // If wrap-around is required, set a1 = 0 and a0 -= 1
         bic \a1\().16b, \a1\().16b, \temp\().16b
         add \a\().4s, \a\().4s, \temp\().4s
+        // range: 0 <= a1 <= 15, -GAMMA2 <= a0 <= GAMMA2
 .endm
 
         /* Parameters */
--- a/dev/aarch64_clean/src/poly_decompose_88_asm.S
+++ b/dev/aarch64_clean/src/poly_decompose_88_asm.S
@@ -9,12 +9,15 @@
 
 // a aliased with a0
 .macro decompose88 a1, a, temp
-        // Compute a1 = round-(a / 190464) ≈ round(a * 1477838209 /
-        // 2^48), where round-() denotes "round half down". This is
-        // exact for 0 <= a < Q. Note that half is rounded down since
-        // 1477838209 / 2^48 ≲ 1 / 190464.
+        // range: 0 <= a <= Q-1 = 88*GAMMA2
+
+        // Compute a1 = round-(a / (2*GAMMA2)) = round-(a / 190464) ≈
+        // round(a * 1477838209 / 2^48), where round-() denotes "round half
+        // down". This is exact for 0 <= a < Q. Note that half is rounded down
+        // since 1477838209 / 2^48 ≲ 1 / 190464.
         sqdmulh \a1\().4s, \a\().4s, barrett_const.4s
         srshr \a1\().4s, \a1\().4s, #17
+        // range: 0 <= a1 <= 44
 
         // If a1 = 44, i.e. a > 87*GAMMA2, proceed as if a' = a - Q was
         // given instead. (For a = 87*GAMMA2 + 1 thus a' = -GAMMA2, we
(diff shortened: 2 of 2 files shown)


dev/aarch64_clean/src/mld_polyvecl_pointwise_acc_montgomery_l4.S
```diff
@@ -56,50 +56,71 @@ MLD_ASM_FN_SYMBOL(polyvecl_pointwise_acc_montgomery_l4_asm)
...
     load_polys q16, q17, q18, q19, a_ptr, 3
     load_polys q20, q21, q22, q23, b_ptr, 3
+    // Bounds: |q{i}| < q  for i in 16...19
+    //                < 9q for i in 20...23
 
     pmlal v24, v25, v16, v20
     pmlal v26, v27, v17, v21
     pmlal v28, v29, v18, v22
     pmlal v30, v31, v19, v23
+    // Bounds: |q{i}| < 36q^2 < MONTGOMERY_REDUCE_STRONG_DOMAIN_MAX
```
mkannwischer: MONTGOMERY_REDUCE_STRONG_DOMAIN_MAX is removed in #668. We need to be careful here to consolidate this before we merge both.

jammychiou1: Right... Same thing for AVX2 basemuls ([here](https://github.com/pq-code-package/mldsa-native/blob/b4209badf3bb948a626dbfde8e5e789f947323d2/dev/x86_64/src/pointwise.S#L81) for example).

I believe #668 will get merged before this. Let me change this proactively. Thanks!

jammychiou1: Done

dev/aarch64_clean/src/intt.S
```diff
@@ -35,27 +35,44 @@
...
         mul      \dst\().4s, \src\().4s, \const\().4s
         mls      \dst\().4s, t2.4s,      consts.s[0]
+        // Bounds: |dst| < BARMUL_BOUND
 .endm
 
 .macro gs_butterfly a, b, root, idx0, idx1
+        // Compute a' = a + b, b' = barmul(a - b, zeta), commonly
+        // known as the Gentleman-Sande butterfly.
+        //
+        // Bounds: |a'| <= |a| + |b|
```
hanno-becker: Nit: Move the bounds comments to the _end_ of the macro?

dev/aarch64_clean/src/intt.S
```diff
@@ -235,27 +252,50 @@ intt_layer5678_start:
...
         load_next_roots_78
 
         // Layer 8
         gs_butterfly_v data0, data1, root1, root1_tw
         gs_butterfly_v data2, data3, root2, root2_tw
 
+        // Bounds: |data{i}| < 2q           for i in 0, 2
+        //                   < BARMUL_BOUND for i in 1, 3
+        //
+        // Note that since 2^31 / q > 256, the sum of all 256 coefficients does
```
hanno-becker: It may be noteworthy to give the reader a sense of how tight this is: 2^31 / q ~ 256.25.

dev/aarch64_clean/src/mld_polyvecl_pointwise_acc_montgomery_l4.S
```diff
@@ -56,50 +56,70 @@ MLD_ASM_FN_SYMBOL(polyvecl_pointwise_acc_montgomery_l4_asm)
...
+    // q{i} stores a's coefficients for i in 16...19, and b's coefficients
+    // for i in 20...23. This applies to all load_polys below.
+    //
+    // Bounds: |q{i}| < q  for i in 16...19
+    //                < 9q for i in 20...23
 
     pmull v24, v25, v16, v20
     pmull v26, v27, v17, v21
     pmull v28, v29, v18, v22
     pmull v30, v31, v19, v23
```
hanno-becker: General comment, I didn't notice this when the code was introduced: We should really use symbolic register values (using `.req`), not architectural ones. It's _much_ easier to read.

dev/aarch64_clean/src/mld_polyvecl_pointwise_acc_montgomery_l4.S
```diff
@@ -56,50 +56,70 @@ MLD_ASM_FN_SYMBOL(polyvecl_pointwise_acc_montgomery_l4_asm)
     movk wtmp, #127, lsl #16
     dup modulus.4s, wtmp
 
-    // load -q^-1 = 4236238847
-    movz wtmp, #57343
-    movk wtmp, #64639, lsl #16
+    // load q^-1 = 58728449
```
hanno-becker: We don't support it yet in assembly files, but can you add
```
/* check-magic: 58728449 == unsigned_mod(pow(MLDSA_Q, -1, 2^32), 2^32) */
```

dev/aarch64_clean/src/mld_polyvecl_pointwise_acc_montgomery_l4.S
jammychiou1: Good idea. Thanks!

While coming up with names for the 4 registers for coefficients from a and b, I wanted to use a0, ..., a3, b0, ..., b3. However b0, ..., b3 are built-in register names already.

I'll resort to use a_0, ..., a_3, b_0, ..., b_3 for now. Any other naming schemes are appreciated!

dev/aarch64_clean/src/mld_polyvecl_pointwise_acc_montgomery_l4.S
jammychiou1: Good idea. Working on it now.

dev/aarch64_clean/src/intt.S
jammychiou1: Agree

jammychiou1:
Thank you @hanno-becker and @mkannwischer for your suggestions. I've implemented them in the new patch.

I also included some annotation for the range of intermediates inside decompose.

dev/aarch64_clean/src/mld_polyvecl_pointwise_acc_montgomery_l4.S
```diff
@@ -38,78 +38,171 @@
...
 .text
 .global MLD_ASM_NAMESPACE(polyvecl_pointwise_acc_montgomery_l4_asm)
 .balign 4
 MLD_ASM_FN_SYMBOL(polyvecl_pointwise_acc_montgomery_l4_asm)
-    // load q = 8380417
-    movz wtmp, #57345
-    movk wtmp, #127, lsl #16
+    // Load MLDSA_Q = 8380417
+    movz wtmp, #0xe001
+    movk wtmp, #0x7f, lsl #16
```
hanno-becker, line 95: In the clean code, we should take the opportunity to write
```
    movz wtmp, #(MLDSA_Q & 0xFFFF)
    movk wtmp, #(MLDSA_Q >> 16), lsl #16
```
and similarly below for 58728449.

jammychiou1: I agree. However, since loading MLDSA_Q (and other constants) are also done in a few other files, I think creating a separate PR for this is more appropriate. WDYT?

mkannwischer: Sounds good to me.

PR #668, opened 2025-11-11
Remove unused spec for mld_montgomery_reduce, add bounds reasoning

### LINKED ISSUE #602: Reason about bounds of `mld_fqmul` and `mld_fqscale`

hanno-becker:
Context:

We previously had two specifications for mld_montgomery_reduce:

- A 'weak' one where the output upper bound matches the input upper bound of mld_reduce32().
- A 'strong' one where the output interval is (-Q,Q), exclusively.

Both contracts were combined into one in a somewhat ad-hoc way (the 'weak' version being encoded as pre- and post-conditions as usual, the 'strong' version being stated as an implication in the post-condition).

The combined contract was used in the CBMC proofs of
- mld_poly_pointwise_montgomery
- mld_polyvecl_pointwise_acc_montgomery

There are two more functions which rely on mld_montgomery_reduce, namely
- mld_fqmul
- mld_fqscale

Observations:
- The 'weak' contract of mld_montgomery_reduce is not needed: In the context of mld_poly_pointwise_montgomery and mld_polyvecl_pointwise_acc_montgomery, the strong version applies.
- The 'strong' version also applies in the context of mld_fqmul and mld_fqscale.

Accordingly, this commit removes the 'weak' version and re-formulates the 'strong' version in traditional pre/post condition form.

The commit also removes the macro wrapper around the input bound, which I find difficult to read. It's easier to understand if the bound is explicit.

The 'strong' version only applies to `mld_fqscale` because of the recent modification of the iNTT output bound from 3/4*Q to Q. Otherwise, an even stronger spec would have been needed. At present, one can even remove mld_fqscale and just call mld_fqmul directly, but this commit does not do this.

The commit also takes the opportunity to add some pen-and-paper bounds reasoning for mld_montgomery_reduce.

* Resolves #602


mldsa/src/reduce.h
```diff
@@ -21,41 +21,33 @@
...
+ * Returns:     Integer congruent to a * R^-1 modulo q, with absolute value
+ *                <= ceil(|a| / 2^32 + MLDSA_Q / 2)
  *
- * Arguments:   - int64_t: finite field element a
+ *              In particular, if |a| <= 2^31 * MLDSA_Q, the absolute value
+ *              of the return value is <= MLDSA_Q.
  *
- * Returns r.
+ *              For |a| < 2^31 * MLDSA_Q, the return value is even < MLDSA_Q
+ *              in absolute value. This is what the CBMC specification encodes.
```
mkannwischer: This appears to state the same thing twice?

mldsa/src/reduce.h
```diff
@@ -21,41 +21,33 @@
...
- *        -MONTGOMERY_REDUCE_DOMAIN_MAX <= a <= MONTGOMERY_REDUCE_DOMAIN_MAX,
- *      compute r == a*2^{-32} (mod MLDSA_Q) such that
- *        INT32_MIN <= r < REDUCE32_DOMAIN_MAX
+ * Description: Generic Montgomery reduction; given a 64-bit integer a, computes
+ *              32-bit integer congruent to a * R^-1 mod q, where R=2^32
  *
- *      The upper-bound on the result ensures that a result from this
- *      function can be used as an input to mld_reduce32() declared below.
+ * Arguments:   - int64_t a: input integer to be reduced, of absolute value
+ *                smaller or equal to INT64_MAX - 2^31 * MLDSA_Q.
```
mkannwischer, line 31: can we not just get rid of the larger input bound altogether and use 2^31 * MLDSA_Q, to simplify?

mldsa/src/reduce.h
```diff
@@ -68,16 +60,38 @@ __contract__(
...
    * PORTABILITY: Right-shift on a signed integer is, strictly-speaking,
    * implementation-defined for negative left argument. Here,
    * we assume it's sign-preserving "arithmetic" shift right. (C99 6.5.7 (5))
    */
   r = r >> 32;
+  /* Bounds: |r >> 32| <= ceil(|r| / 2^32)
+   *                   <= ceil(|a| / 2^32 + MLDSA_Q / 2)
+   *
+   * (Note that |a >> n| = ceil(|a| / 2^n) for negative a).
+   * This is the bound states in the documentation.
```
mkannwischer: fix grammer or remote the large bound altogether?

hanno-becker: Fixing grammar

mldsa/src/reduce.h
hanno-becker: This is deliberate. At the documentation-level, I kept things generic, but CBMC picks a specific bound.

mldsa/src/reduce.h
hanno-becker: No, one is <=, the other <. I don't have a pen-and-paper proof for the correctness of the spec with <, though CBMC seems happy with it.

mkannwischer: Sorry, sleepy me.

mldsa/src/reduce.h
mkannwischer: okay

mldsa/src/reduce.h
```diff
@@ -21,41 +21,33 @@
...
-  ensures(return_value >= INT32_MIN && return_value < REDUCE32_DOMAIN_MAX)
-
-  /* Special case - for stronger input bounds, we can ensure stronger bounds on the output */
-  ensures((a >= -MONTGOMERY_REDUCE_STRONG_DOMAIN_MAX && a < MONTGOMERY_REDUCE_STRONG_DOMAIN_MAX) ==>
-          (return_value > -MLDSA_Q && return_value < MLDSA_Q))
+  /* We don't attempt to express an input-dependent output bound
+   * as the post-condition here, as all call-sites satisfy the
+   * absolute input bound 2^31 * MLDSA_Q and higher-level
+   * reasoning can be conducted using |return_value| < MLDSA_Q. */
+  requires(a >  ((int64_t)INT32_MIN * MLDSA_Q) &&
```
jammychiou1: Any reason not to use `(((int64_t)1 << 31) * MLDSA_Q)` like below?

mldsa/src/reduce.h
```diff
@@ -70,14 +62,35 @@ __contract__(
...
+   *             <= |a| / 2^32 + MLDSA_Q / 2
+   *             <= 2^31 * MLDSA_Q / 2^32 + MLDSA_Q / 2
+   *              = MLDSA_Q.
+   *
+   * proving that |a| <= 2^31 * MLDSA_Q implies |return_value| <= MLDSA_Q.
+   * (In general, we would only have |a >> n| <= ceil(|a| / 2^n).
+   *
+   * Moreover, equality in the above chain can only occur if all intermediate
+   * inequalities are equalities, hence |a| / 2^32 = MLDSA_Q / 2, or
+   * |a| = 2^31 * MLDSA_Q.
```
jammychiou1: Is the inclusive bound still necessary? All the places that mention MONTGOMERY_REDUCE_STRONG_DOMAIN_MAX now are strictly within the bound, I think.

If only the exclusive bound is needed, I think we can use `<` on L84 and drop this paragraph here.

hanno-becker: Agree, done

mldsa/src/reduce.h
hanno-becker: Agree, done

PR #674, opened 2025-11-12
sign.c: zeroize buffers in all code paths

mkannwischer:
In various functions in sign.c, the zeroization would only happen in case the function returned correctly. In the case of failure, there would be a shortcut that returns without zeroization.
This commit fixes that ensuring zeroization for all code paths.

### DIFF
--- a/mldsa/src/sign.c
+++ b/mldsa/src/sign.c
@@ -525,6 +525,7 @@ int crypto_sign_signature_internal(uint8_t sig[CRYPTO_BYTES], size_t *siglen,
                                    const uint8_t sk[CRYPTO_SECRETKEYBYTES],
                                    int externalmu)
 {
+  int result;
   uint8_t seedbuf[2 * MLDSA_SEEDBYTES + MLDSA_TRBYTES + 2 * MLDSA_CRHBYTES];
   uint8_t *rho, *tr, *key, *mu, *rhoprime;
   mld_polyvecl mat[MLDSA_K], s1;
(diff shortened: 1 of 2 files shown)


mldsa/src/sign.c
```diff
@@ -629,7 +634,8 @@ int crypto_sign_signature(uint8_t sig[CRYPTO_BYTES], size_t *siglen,
     /* To be on the safe-side, make sure *siglen has a well-defined */
     /* value, even in the case of error.                            */
     *siglen = 0;
-    return -1;
+    result = -1;
+    goto cleanup;
```
hanno-becker, line 645: This can be argued to be unnecessary, since no buffer has been written to.

mkannwischer: A compiler could move 
```
  pre[0] = 0;
  pre[1] = (uint8_t)ctxlen;
``` 
to before conditional? 
Anyway, I did not want to get into reasoning if steps can be reordered by the compiler or not.

hanno-becker: Yes, you're right. It would only be a potential avenue if we can't solve the CBMC issues otherwise. But this function doesn't seem to be the issue anyway.

mldsa/src/sign.c
```diff
@@ -716,6 +730,7 @@ int crypto_sign_verify_internal(const uint8_t *sig, size_t siglen,
                                 int externalmu)
 {
   unsigned int i;
+  int res = 0;
```
hanno-becker: We shouldn't default to success. Change to `int res;` here and set `res = 0` once everything went through?

proofs/cbmc/crypto_sign_signature_internal/Makefile
```diff
@@ -33,7 +33,8 @@ USE_DYNAMIC_FRAMES=1
 
 # Disable any setting of EXTERNAL_SAT_SOLVER, and choose SMT backend instead
 EXTERNAL_SAT_SOLVER=
-CBMCFLAGS=--smt2
+#CBMCFLAGS=--smt2
```
hanno-becker: Sorry, my bad. Can you remove this line?

hanno-becker (changes requested):
One small issue, otherwise LGTM

mldsa/src/sign.c

proofs/cbmc/crypto_sign_signature_internal/Makefile
mkannwischer: Fixed.

PR #675, opened 2025-11-12
Serial FIPS202: Do not declare mld_poly_uniform_gamma1_4x

mkannwischer:
When MLD_CONFIG_SERIAL_FIPS202_ONLY is set, we are not using or defining mld_poly_uniform_gamma1_4x. Hence, we should also not declare it. This commit fixes that.
Without this fix, on gets unused function warnings in a monobuild marking internal functions as static.


hanno-becker:
> Without this fix, on gets unused function warnings in a monobuild marking internal functions as static.

Why did CI not flag this as an error?

hanno-becker (approved):

PR #677, opened 2025-11-12
native chknorm: 0/0xFFFFFFFF -> 0/1 

mkannwischer:
Previously the native chknorm functions would follow the C implementation,
i.e., return 0 if all coefficients are within bound and 0xFFFFFFFF otherwise.
This leads to problems with the run-time dispatch in
https://github.com/pq-code-package/mldsa-native/pull/607
as we want to use -1 to signal that the current platforms lacks the required
capabilities to run the native code, and we should fall back to the C
implementation.

This commit changes the backend API to return 1 in the failure mode.
This will allow to implement the run-time dispatch.

### DIFF
--- a/mldsa/src/poly.c
+++ b/mldsa/src/poly.c
@@ -589,8 +589,11 @@ uint32_t mld_poly_chknorm(const mld_poly *a, int32_t B)
 #if defined(MLD_USE_NATIVE_POLY_CHKNORM)
   /* TODO: proof */
   mld_assert_bound(a->coeffs, MLDSA_N, -REDUCE32_RANGE_MAX, REDUCE32_RANGE_MAX);
-  return mld_poly_chknorm_native(a->coeffs, B);
-#else
+
+  /* The native backend returns 0 if all coeffs within the bound, 1 otherwise */
+  /* Convert to 0 / 0xFFFFFFFF here */
+  return 0U - (uint32_t)mld_poly_chknorm_native(a->coeffs, B);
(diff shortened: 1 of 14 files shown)


mldsa/src/poly.c
```diff
@@ -623,7 +623,8 @@ uint32_t mld_poly_chknorm(const mld_poly *a, int32_t B)
     t |= mld_ct_cmask_neg_i32(B - 1 - mld_ct_abs_i32(a->coeffs[i]));
   }
 
-  return t;
+  /* Return 1, if t = 0xFFFFFFFF */
+  return t & 1;
```
hanno-becker: If the goal is to not leak which secret is out of bound, this is a problematic change as the compiler can recognize that if you do `t |= ...` repeatedly and in the and merely test `t & 1`, you can abort early once the low-bit is set.

hanno-becker: Have you considered just changing the native API, but not the frontend?

Alternatively, I think one would need to do something like
```c
t = mld_value_barrier_u32(t | mld_ct_cmask_neg_i32(B - 1 - mld_ct_abs_i32(a->coeffs[i])));
```

mkannwischer: Okay, agreed - this would need more work.
As changing this is not necessary to unblock #607, I removed changing the singature of the C version of chknorm. We can consider that separately.

hanno-becker (approved):

Issue #603, opened 2025-11-04
Support randombytes() with a failure mode


mkannwischer:
Some randombytes implementations may not always succeed.
With the randombytes API, we cannot support that case: 
```
void mlk_randombytes(uint8_t *out, size_t outlen)
```

The expected behavior is that the failure gets escalated to the caller.

PR #689, opened 2025-11-12
API: add failure mode support for randombytes()

L-series:
This PR adds failure mode support for the randombytes() interface.

Marking as draft as there are a few points for which I need clarifications. These are the following:

1. What is to be done regarding the OQS_randombytes API integration? From what I understand this is defined as having return type `void` on the liboqs side. Do we make a PR to change this on their end also or do we simply modify the return type of our inline `mld_randombytes` functions defined in files `integration/liboqs/config_*.h`?
2. Is it appropriate that I am adding the `CHECK(x)` macro to `bench/test_components_mldsa.c`? 
3. Should I add a test which verifies the behavior in case `randombytes()` fails?

As an aside. It could be a good idea to factor our the `CHECK` macro into its own test header file as the code is currently duplicated in many files across the project. @hanno-becker @mkannwischer please let me know your thoughts.

### DIFF
--- a/.github/actions/config-variations/action.yml
+++ b/.github/actions/config-variations/action.yml
@@ -62,6 +62,7 @@ runs:
         extra_env: 'ASAN_OPTIONS=detect_leaks=1'
         examples: false # Some examples use a custom config themselves
         alloc: false # Requires custom config
+        rng_fail: false # Requires custom config
     - name: "Custom zeroization (explicit_bzero)"
       if: ${{ inputs.tests == 'all' || contains(inputs.tests, 'custom-zeroize') }}
       uses: ./.github/actions/multi-functest
@@ -76,6 +77,7 @@ runs:
         opt: ${{ inputs.opt }}
         examples: false # Some examples use a custom config themselves
         alloc: false # Requires custom config
+        rng_fail: false # Requires custom config
     - name: "Custom native capability functions (static ON)"
       if: ${{ inputs.tests == 'all' || contains(inputs.tests, 'native-cap-ON') }}
       uses: ./.github/actions/multi-functest
@@ -90,6 +92,7 @@ runs:
         opt: ${{ inputs.opt }}
         examples: false # Some examples use a custom config themselves
         alloc: false # Requires custom config
+        rng_fail: false # Requires custom config
     - name: "Custom native capability functions (static OFF)"
       if: ${{ inputs.tests == 'all' || contains(inputs.tests, 'native-cap-OFF') }}
       uses: ./.github/actions/multi-functest
@@ -104,6 +107,7 @@ runs:
         opt: ${{ inputs.opt }}
         examples: false # Some examples use a custom config themselves
         alloc: false # Requires custom config
+        rng_fail: false # Requires custom config
     - name: "Custom native capability functions (ID_AA64PFR1_EL1 detection)"
       if: ${{ (inputs.tests == 'all' || contains(inputs.tests, 'native-cap-ID_AA64PFR1_EL1')) && runner.os == 'Linux' && runner.arch == 'ARM64' }}
       uses: ./.github/actions/multi-functest
@@ -118,6 +122,7 @@ runs:
         opt: ${{ inputs.opt }}
         examples: false # Some examples use a custom config themselves
         alloc: false # Requires custom config
+        rng_fail: false # Requires custom config
     - name: "Custom native capability functions (CPUID AVX2 detection)"
       if: ${{ (inputs.tests == 'all' || contains(inputs.tests, 'native-cap-CPUID_AVX2')) && runner.os == 'Linux' && runner.arch == 'X64' }}
       uses: ./.github/actions/multi-functest
@@ -132,6 +137,7 @@ runs:
         opt: ${{ inputs.opt }}
         examples: false # Some examples use a custom config themselves
         alloc: false # Requires custom config
+        rng_fail: false # Requires custom config
     - name: "No ASM"
       if: ${{ inputs.tests == 'all' || contains(inputs.tests, 'no-asm') }}
       uses: ./.github/actions/multi-functest
@@ -146,6 +152,7 @@ runs:
         opt: ${{ inputs.opt }}
         examples: false # Some examples use a custom config themselves
         alloc: false # Requires custom config
+        rng_fail: false # Requires custom config
     - name: "Serial FIPS202 (no batched Keccak)"
       if: ${{ inputs.tests == 'all' || contains(inputs.tests, 'serial-fips202') }}
       uses: ./.github/actions/multi-functest
@@ -160,6 +167,7 @@ runs:
         opt: ${{ inputs.opt }}
         examples: false # Some examples use a custom config themselves
         alloc: false # Requires custom config
+        rng_fail: false # Requires custom config
     - name: "Custom randombytes"
       if: ${{ inputs.tests == 'all' || contains(inputs.tests, 'custom-randombytes') }}
       uses: ./.github/actions/multi-functest
@@ -174,6 +182,7 @@ runs:
         opt: ${{ inputs.opt }}
         examples: false # Some examples use a custom config themselves
         alloc: false # Requires custom config
+        rng_fail: false # Requires custom config
     - name: "Custom memcpy"
       if: ${{ inputs.tests == 'all' || contains(inputs.tests, 'custom-memcpy') }}
       uses: ./.github/actions/multi-functest
@@ -188,6 +197,7 @@ runs:
         opt: ${{ inputs.opt }}
         examples: false # Some examples use a custom config themselves
         alloc: false # Requires custom config
+        rng_fail: false # Requires custom config
     - name: "Custom memset"
       if: ${{ inputs.tests == 'all' || contains(inputs.tests, 'custom-memset') }}
       uses: ./.github/actions/multi-functest
@@ -202,6 +212,7 @@ runs:
         opt: ${{ inputs.opt }}
         examples: false # Some examples use a custom config themselves
         alloc: false # Requires custom config
+        rng_fail: false # Requires custom config
     - name: "Custom stdlib (memcpy + memset)"
       if: ${{ inputs.tests == 'all' || contains(inputs.tests, 'custom-stdlib') }}
       uses: ./.github/actions/multi-functest
@@ -216,3 +227,4 @@ runs:
         opt: ${{ inputs.opt }}
         examples: false # Some examples use a custom config themselves
         alloc: false # Requires custom config
+        rng_fail: false # Requires custom config
--- a/.github/actions/functest/action.yml
+++ b/.github/actions/functest/action.yml
@@ -60,6 +60,9 @@ inputs:
   alloc:
     description: Determine whether to run alloc tests or not
     default: "true"
+  rng_fail:
+    description: Determine whether to run rng fail tests or not
+    default: "true"
   extra_args:
     description: Additional arguments to pass to the tests script
     default: ""
@@ -81,6 +84,7 @@ runs:
           echo EXAMPLES="${{ inputs.examples == 'true' && 'examples' || 'no-examples' }}" >> $GITHUB_ENV
           echo STACK="${{ inputs.stack == 'true' && 'stack' || 'no-stack' }}" >> $GITHUB_ENV
           echo ALLOC="${{ inputs.alloc == 'true' && 'alloc' || 'no-alloc' }}" >> $GITHUB_ENV
+          echo RNGFAIL="${{ inputs.rng_fail == 'true' && 'rng-fail' || 'no-rng-fail' }}" >> $GITHUB_ENV
       - name: Setup nix
         uses: ./.github/actions/setup-shell
         with:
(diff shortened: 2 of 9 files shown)


hanno-becker:
Thank you @L-series!

> What is to be done regarding the OQS_randombytes API integration? From what I understand this is defined as having return type void on the liboqs side. Do we make a PR to change this on their end also or do we simply modify the return type of our inline mld_randombytes functions defined in files integration/liboqs/config_*.h?

So long as OQS' randombytes has return type `void`, we would merely need to change the wrapper to

```c
#define MLD_CONFIG_CUSTOM_RANDOMBYTES
#if !defined(__ASSEMBLER__)
#include <oqs/rand.h>
#include <stdint.h>
#include "../../mldsa/src/sys.h"
static MLD_INLINE int mld_randombytes(uint8_t *ptr, size_t len)
{
  OQS_randombytes(ptr, len);
  return 0;
}
```

or am I missing something?

> Is it appropriate that I am adding the CHECK(x) macro to bench/test_components_mldsa.c?

Yes, I see no problem with that.

> Should I add a test which verifies the behavior in case randombytes() fails?

The CBMC proofs will need to cover the new behavior, but yes, we also need have specific tests that trigger failure of `randombytes()` at various points. This will require a custom configuration and a dedicated test; it's not clear on first sight where this should live. It does not fit into the existing tests in tests/*, nor would it make sense to add a new test class (alongside func, KAT, ACVP, stack, bench) for just this. I tend towards a custom 'example', despite being more of a test than an example -- but that line is blurry anyway.

mldsa/src/sign.c
```diff
@@ -709,7 +712,11 @@ int crypto_sign_signature(uint8_t sig[CRYPTO_BYTES], size_t *siglen,
 
   /* Randomized variant of ML-DSA. If you need the deterministic variant,
    * call crypto_sign_signature_internal directly with all-zero rnd. */
-  mld_randombytes(rnd, MLDSA_RNDBYTES);
+  if (mld_randombytes(rnd, MLDSA_RNDBYTES) != 0)
+  {
+    *siglen = 0;
+    return -1;
+  }
```
mkannwischer, line 802: This skips zeroization of pre which has already been written to. Please instead goto cleanup.

examples/basic_rng_failure/mldsa_native/mldsa
mkannwischer, line 1: I'm not a fan of adding such tests as examples - examples are intended to showcase how an integration of mldsa-native into other libraries can work.

Have you considered implementing this test in a similar way as the `break_pct_config`? It may be a little tricky to make it fail in specific algorithms. 
If that does not work well, I'd prefer adding the tests to test_mldsa.c guarded by a pre-processor flag and adding a specific CI job just for that.
I think that is preferable over polluting the examples/

mkannwischer (changes requested):
Thanks @L-series. 

Could you please rebase on top of main - there has been some restructuring to the config.h? 

ALso could you please add the new failure cases to the documentation of the top level functions in sign.h and mldsa_native.h?

examples/basic_rng_failure/mldsa_native/mldsa
L-series: Hello Matthias , thank you for the review. I'll rebase the patchset and fix the implementation of the test cases as you have suggested. I believe it should be possible to implement it as is `break_pct_config` !

mldsa/src/sign.c
L-series: Fixed and added  documentation to sign.h and mldsa_native.h

examples/basic_rng_failure/mldsa_native/mldsa
L-series: Ive gone ahead and implemented it similarly to `break_pct`. Please let me know what you think!

hanno-becker (changes requested):
Thank you @L-series for pushing this, I think this will be a valuable addition!

I don't think we should have a `MLD_CONFIG_RANDOMBYTES_FAILURE_TEST` option. I agree that there is some similarity to the PCT breakage test, but the difference here is that `randombytes` is externally controlled, so an injection of error can happen through a dedicated implementation of `randombytes` and need not happen from the mldsa-native source.

I think we should remove `MLD_CONFIG_RANDOMBYTES_FAILURE_TEST` and instead have a separate test file `test/test_rng_fail.c` or similar, which uses a custom `randombytes` implementation injecting failures at various points. One way this could be achieved is by first doing a 'dry-run' of ML-DSA API from `test_rng_fail.c` and having the instrumented `randombytes` implementation merely count (in a global variable) how many times it has been invoked. And then, say there have been `N` invocations of `randombytes`, `test_rng_fail.c` runs the standard tests _again_ N times, and in the i-th invocation, `randombytes` is configured to fail at the i-th invocation, but working beforehand. This way, we can trigger the failure handling for _all_ invocations of `randombytes`.

What do you think?

mldsa/src/sign.c
```diff
@@ -735,7 +743,11 @@ int crypto_sign_signature_extmu(uint8_t sig[MLDSA_CRYPTO_BYTES], size_t *siglen,
 
   /* Randomized variant of ML-DSA. If you need the deterministic variant,
    * call crypto_sign_signature_internal directly with all-zero rnd. */
-  mld_randombytes(rnd, MLDSA_RNDBYTES);
+  if (mld_randombytes(rnd, MLDSA_RNDBYTES) != 0)
+  {
+    *siglen = 0;
+    return -1;
```
gilles-peskine-arm: I am investigating integrating mldsa-native into TF-PSA-Crypto (formerly known as Mbed TLS). In our library, we would need to report the specific error code from the `randombytes` call (e.g. to distinguish a possibly temporary lack of entropy from a system integration failure). We aren't actually going to use `randombytes` at all because [we'll use `keypair_internal` and `signature_internal` anyway](https://github.com/pq-code-package/mldsa-native/issues/603#issuecomment-3672009944). However, it's likely that other users with similar system designs would want to retrieve the specific error code from their `randombytes` function.

test/custom_memcpy_config.h
```diff
@@ -628,6 +630,29 @@ static MLD_INLINE void *mld_memcpy(void *dest, const void *src, size_t n)
  *****************************************************************************/
 /* #define MLD_CONFIG_SERIAL_FIPS202_ONLY */
 
+/******************************************************************************
+ * Name:        MLD_CONFIG_RANDOMBYTES_FAILURE_TEST
```
gilles-peskine-arm: > I think we should remove `MLD_CONFIG_RANDOMBYTES_FAILURE_TEST` and instead have a separate test file `test/test_rng_fail.c` or similar, which uses a custom randombytes implementation injecting failures at various points.

Several examples already comes with a test-only RNG. How about instrumenting that to have a failure mode, instead of adding a separate example?

mldsa/src/randombytes.h
```diff
@@ -13,12 +13,29 @@
 
 #if !defined(MLD_CONFIG_NO_RANDOMIZED_API)
 #if !defined(MLD_CONFIG_CUSTOM_RANDOMBYTES)
-void randombytes(uint8_t *out, size_t outlen);
-static MLD_INLINE void mld_randombytes(uint8_t *out, size_t outlen)
+int randombytes(uint8_t *out, size_t outlen);
```
gilles-peskine-arm, line 17: Maybe `randombytes` should be declared with `MLD_API_MUST_CHECK_RETURN_VALUE`? (If it's available here, I don't really understand the API vs implementation namespaces.)

L-series:
@hanno-becker 

Thanks  for the feedback! I completely agree with your approach, using a custom randombytes implementation  is  cleaner than having a compile-time config option for what is essentially external dependency testing. The dry-run strategy to automatically discover and test all randombytes call sites is  likely more maintainable.

I'll proceed to:

* Remove `MLD_CONFIG_RANDOMBYTES_FAILURE_TEST` from the config
and create test/test_rng_fail.c with the custom randombytes implementation
* Implementing the N-runs approach you described

hanno-becker:
@L-series Please see the recently merged allocation test https://github.com/pq-code-package/mlkem-native/pull/1424 in mlkem-native (WIP PR for mldsa-native is https://github.com/pq-code-package/mldsa-native/pull/810) for how to approach the rng-failure test -- we should strive to have this uniform.

mldsa/src/randombytes.h
hanno-becker: Yes, I fully agree

test/custom_memcpy_config.h
hanno-becker: We just did something similar for allocations https://github.com/pq-code-package/mlkem-native/pull/1424 I hope we can follow the same approach.

mldsa/src/sign.c
hanno-becker: We started refining error codes in https://github.com/pq-code-package/mldsa-native/pull/801 which introduced fallible allocation; we should continue with this and have `MLD_ERR_RNG_FAILURE` or similar for failing RNGs.

mldsa/src/randombytes.h
L-series: Done!

test/custom_memcpy_config.h
L-series: I used the alloc test as inspiration for the randombytes test. It is now reworked.

mldsa/src/sign.c
L-series: Done, added `MLD_ERR_RNG_FAIL`

mkannwischer (changes requested):
Thanks @L-series. The test looks much better to me now.
Please also add it to `scripts/tests` - otherwise large parts of the CI don't actually run it.

Do you want some help with the CBMC proofs? I can take a look later.

mldsa/src/sign.c
```diff
@@ -61,7 +61,8 @@ __contract__(
   requires(memory_no_alias(sk, MLDSA_CRYPTO_SECRETKEYBYTES))
   ensures(return_value == 0
 	  || return_value == MLD_ERR_FAIL
-	  || return_value == MLD_ERR_OUT_OF_MEMORY)
+	  || return_value == MLD_ERR_OUT_OF_MEMORY
+         || return_value == MLD_ERR_RNG_FAIL)
```
hanno-becker: Nit: Indentation

mldsa/src/sign.c
```diff
@@ -779,7 +783,12 @@ int crypto_sign_signature(uint8_t sig[MLDSA_CRYPTO_BYTES], size_t *siglen,
 
   /* Randomized variant of ML-DSA. If you need the deterministic variant,
    * call crypto_sign_signature_internal directly with all-zero rnd. */
-  mld_randombytes(rnd, MLDSA_RNDBYTES);
+  if (mld_randombytes(rnd, MLDSA_RNDBYTES) != 0)
+  {
+    *siglen = 0;
```
hanno-becker: This is not needed

mldsa/src/sign.c
```diff
@@ -818,7 +827,11 @@ int crypto_sign_signature_extmu(uint8_t sig[MLDSA_CRYPTO_BYTES], size_t *siglen,
...
```
hanno-becker, line 843: We should jump to `cleanup` section here which includes `mld_zeroize(rnd, ..)`, even though `mld_randombytes` failed.

test/test_rng_fail.c
```diff
@@ -0,0 +1,185 @@
...
+
+  TEST_RNG_FAILURE("crypto_sign_signature_extmu",
+                   mld_signature_extmu(sig, &siglen, mu, sk));
+  return 0;
+}
+
+int main(void)
+{
+  printf("Testing ML-DSA parameter set %d\n", MLD_CONFIG_API_PARAMETER_SET);
+  printf("========================================\n");
```
hanno-becker: Nit: I'd suggest to drop this, it doesn't provide full context and is also not present in `test_alloc.c`. The individual prints in `test_xxx_rng_failure()` are expressive enough.

test/test_rng_fail.c
```diff
@@ -0,0 +1,185 @@
...
+  {
+    return 1;
+  }
+
+  if (test_signature_extmu_rng_failure() != 0)
+  {
+    return 1;
+  }
+
+  printf("\nAll RNG failure tests PASSED!\n");
```
hanno-becker: Drop this in alignment with `test_alloc.c`?

test/test_rng_fail.c
```diff
@@ -0,0 +1,185 @@
...
+
+  if (test_signature_rng_failure() != 0)
+  {
+    return 1;
+  }
+
+  if (test_signature_extmu_rng_failure() != 0)
+  {
+    return 1;
+  }
```
hanno-becker: I'd be inclined to just cover the same set of functions as in `test_alloc.c`, regardless of whether (currently) they call RNG or not. For example, `crypto_sign` also indirectly uses RNG, and the present test would not catch a bad error propagation in the call path from `crypto_sign -> crypto_sign_signature`.

hanno-becker (changes requested):
Thank you very much @L-series, mostly looks good! Some smaller comments.

mldsa/src/sign.c
```diff
@@ -360,7 +361,10 @@ int crypto_sign_keypair(uint8_t pk[MLDSA_CRYPTO_PUBLICKEYBYTES],
 {
   MLD_ALIGN uint8_t seed[MLDSA_SEEDBYTES];
   int ret;
-  mld_randombytes(seed, MLDSA_SEEDBYTES);
+  if (mld_randombytes(seed, MLDSA_SEEDBYTES) != 0)
+  {
+    return MLD_ERR_RNG_FAIL;
```
mkannwischer: We should zeroize `seed` in this case, i.e., change to the `goto cleanup` approach used elsewhere.

mldsa/src/sign.h
```diff
@@ -101,6 +101,8 @@
  *                  used and an allocation via MLD_CUSTOM_ALLOC returned NULL.
  *              - MLD_ERR_FAIL: Other kinds of failure, incl. PCT failure
  *                  if MLD_CONFIG_KEYGEN_PCT is enabled.
+ *              - MLD_ERR_RNG_FAIL: Returned indirectly from mld_randombytes
+ *                  if random number generation fails.
```
mkannwischer: Can we first list the explcit failure reasons, and then the generic `MLD_ERR_FAIL`, here? 
Also please use the same description. I think `Random number generation failed.` is sufficient.

mldsa/mldsa_native.h
mkannwischer, line 1: same changes as I suggested for sign.h should be applied.

mldsa/src/sign.c
L-series: Done!

mldsa/src/sign.h
L-series: Done! Unified the comments for MLD_ERR_RNG_FAIL, moved them above ERR_FAIL in sign.h and mldsa_native.h

test/mk/components.mk
```diff
@@ -119,9 +134,14 @@ $(BUILD_DIR)/$(1)/bin/test_alloc$(subst mldsa,,$(1)): LDLIBS += -L$(BUILD_DIR) -
 $(BUILD_DIR)/$(1)/bin/test_alloc$(subst mldsa,,$(1)): $(BUILD_DIR)/$(1)/test/test_alloc.c.o $(BUILD_DIR)/lib$(1)_alloc.a $(call MAKE_OBJS, $(BUILD_DIR)/$(1), $(wildcard test/notrandombytes/*.c))
 endef
 
+# Special rule for test_rng_fail - link against rng_fail libraries with custom randombytes config
+define ADD_SOURCE_RNG_FAIL
+$(BUILD_DIR)/$(1)/bin/test_rng_fail$(subst mldsa,,$(1)): LDLIBS += -L$(BUILD_DIR) -l$(1)_rng_fail
+$(BUILD_DIR)/$(1)/bin/test_rng_fail$(subst mldsa,,$(1)): $(BUILD_DIR)/$(1)/test/test_rng_fail.c.o $(BUILD_DIR)/lib$(1)_rng_fail.a
+endef
```
hanno-becker, line 141: Do we need a custom build here? Can we not use the default build but just link against the customized randombytes?

test/mk/rules.mk
```diff
@@ -113,3 +113,34 @@ $(BUILD_DIR)/mldsa87/alloc/%.S.o: %.S $(CONFIG)
 	$(Q)echo "  AS      $@"
 	$(Q)[ -d $(@D) ] || mkdir -p $(@D)
 	$(Q)$(CC) -c -o $@ $(CFLAGS) $<
+
+$(BUILD_DIR)/mldsa44/rng_fail/%.c.o: %.c $(CONFIG)
+	$(Q)echo "  CC      $@"
+	$(Q)[ -d $(@D) ] || mkdir -p $(@D)
+	$(Q)$(CC) -c -o $@ $(CFLAGS) $<
```
hanno-becker, line 120: Same comment here: I'm not sure we need a separate build.

hanno-becker (approved):
Looks great, thanks a lot @L-series 

The only thing that should be improved is the build: We don't need a separate build for the fail-RNG test, but can just link a different randombytes. This will avoid the build time for `make test` creep up even further.

I leave it to you whether you want to do this now or as a follow-up.

mkannwischer:
> Looks great, thanks a lot @L-series
> 
> The only thing that should be improved is the build: We don't need a separate build for the fail-RNG test, but can just link a different randombytes. This will avoid the build time for `make test` creep up even further.
> 
> I leave it to you whether you want to do this now or as a follow-up.

@L-series could you do that as a follow-up, please? Then we can already get the restructuring PRs going without conflicting with this PR.

Issue #625, opened 2025-11-06
CI: Add cross tests for RVV


mkannwischer:
Port https://github.com/pq-code-package/mlkem-native/commit/9c93e942675308b828e7ba4db25a20726728368b

PR #692, opened 2025-11-13
CI: Add cross tests for RVV

willieyz:
- Resolves: #625

- This PR replaces the existing RISC-V64 tests with multiple RVV RISC-V64 tests that use different VLEN values (128/256/512/1024).


mkannwischer (approved):
LGTM. Thanks @willieyz!

Issue #620, opened 2025-11-06
`MLD_SYS_LITTLE_ENDIAN` not set when using MSVC


mkannwischer:
I noticed in https://github.com/pq-code-package/mldsa-native/actions/runs/19122367507/job/54645299261?pr=618, that when using MSVC, we are using the portable code path in https://github.com/pq-code-package/mldsa-native/blob/f7fccd93ad5cf88ad735df4c422fceb5c4e47cf7/mldsa/src/fips202/keccakf1600.c#L51.

That should be fixed.
Probably we want to fix this in mlkem-native first: https://github.com/pq-code-package/mlkem-native/issues/1283

PR #704, opened 2025-11-17
sys.h: Detect little endian when compiling with MSVC

mkannwischer:
This commit adds detection for endianness with MSVC - MSVC does not define __BYTE_ORDER__, and, hence our current detection does not capture it.

MSVC currently supports 3 platforms: x86, x86_64, and AArch64 (AArch32 was recently removed in VS 2026 18.0)
See
https://learn.microsoft.com/en-us/cpp/overview/supported-platforms-visual-cpp

x86 and x86_64 are always little endian.
AArch64 can be either little endian or big endian, but MSVC/Windows only supports little endian:
https://learn.microsoft.com/en-us/cpp/build/arm64-windows-abi-conventions

For AArch32, I can't find explicit statements that MSVC was always limited to little endian, so I've commited that.

See
https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros

- Resolves https://github.com/pq-code-package/mldsa-native/issues/620

### DIFF
--- a/mldsa/src/sys.h
+++ b/mldsa/src/sys.h
@@ -22,6 +22,15 @@
 #error "__BYTE_ORDER__ defined, but don't recognize value."
 #endif
 #endif /* __BYTE_ORDER__ */
+
+/* MSVC does not define __BYTE_ORDER__. However, MSVC only supports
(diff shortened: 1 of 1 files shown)


hanno-becker (approved):
I had looked into this independently and came to the same conclusion.

PR #706, opened 2025-11-17
Standardize on signed integers in quantifiers and loops.

rod-chapman:
Conjecture: it is better for proof performance and stability to use signed "int" types for quantifiers, loop indexing expressions, for loop counters. This PR implements this change to gain data on its impact on proof and runtime performance.

Specifically, use signed "int" for:
1. for loop counters
2. quantified expressions
3. variables and struct fields that count or index into arrays
4. formal parameters for array indexes, offsets and counts

Additionally, strengthen pre-conditions on polyvecl_add() and polyveck_add() to stabilize proof of those functions for all parameter sets.

Opening a PR here to get Benchmark results.

### DIFF
--- a/dev/aarch64_clean/meta.h
+++ b/dev/aarch64_clean/meta.h
@@ -32,6 +32,7 @@
 
 
 #if !defined(__ASSEMBLER__)
+#include <stddef.h>
 #include "../api.h"
 #include "src/arith_native_aarch64.h"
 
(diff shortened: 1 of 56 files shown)


mkannwischer:
@rod-chapman, are you still looking into this, or can this be closed?

rod-chapman:
The proof performance results were noticeably better, but there was an (unexplained) runtime performance penalty in the C back-end that I never managed to diagnose. Suggest we close this for now.

Issue #44, opened 2025-03-04
C Implementation: Public Key from Private


jakemas:
### Create a utility function that constructs public key from private key.

It is often useful when serializing asymmetric key pairs to populate both the public and private elements, given only the private element. For this to be possible, an algorithm utility function is often provided to derive key material. ML-DSA does not support this in the reference implementation.

#### Background ML-DSA keypairs 

An ML-DSA private key is constructed of the following elements: (ref https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf)
```
sk = (
      rho, // public random seed (32-bytes)
      tr,  // public key hash (64-bytes)
      key, // private random seed (32-bytes) (utilized during sign)
      t0,  // polynomial vector: encodes the least significant bits of public-key polynomial t, facilitating certain computational efficiencies.
      s1,  // secret polynomial vectors. These vectors contain polynomials with coefficients in a specified range, 
      s2.  // serving as the secret components in the lattice-based structure of ML-DSA.
)
```

An ML-DSA public key is constructed of the following elements:
```
pk = (
      rho, // public random seed (32-bytes)
      t1.  // compressed representation of the public key polynomial 
)
```

- The vector t is decomposed into two parts:
- `t1`: Represents the higher-order bits of `t`.
- `t0`: Represents the lower-order bits of `t`.

One can see that to reconstruct the public key from the private key, one must:
1. Extract all elements from `sk`, using the existing function in `/ml_dsa_ref/packing.c`: `ml_dsa_unpack_sk`
    1. This will provide `sk = (rho, tr, key, t0, s1, s2)`.
2. Reconstruct `A` using `rho` with the existing function in `/ml_dsa_ref/polyvec.c`: `ml_dsa_polyvec_matrix_expand`
3. Reconstruct `t` from `t = A*s1 + s2`
4. Drop `d` lower bits from `t` to get `t1`
5. Pack `rho`, `t1` into public key.
6. Verify `pk` matches expected value, by comparing SHAKE256(pk) + `tr` (unpacked from secret key).

mkannwischer:
@jakemas, are you still planning to work on this? If not, could you unassign yourself?

mkannwischer:
Due to lack of feedback, I'm removing this from v1.0.0-alpha. 
Note that we can always add new APIs later on.

Issue #713, opened 2025-11-18
Make memory backing top-level allocations configurable


hanno-becker:
The core APIs allocate a number of large objects on the stack as separate local variables. For example, `crypto_sign_verify_internal` starts like this:
```c
int crypto_sign_verify_internal(const uint8_t *sig, size_t siglen,
                                const uint8_t *m, size_t mlen,
                                const uint8_t *pre, size_t prelen,
                                const uint8_t pk[CRYPTO_PUBLICKEYBYTES],
                                int externalmu)
{
  unsigned int i;
  int res;
  MLD_ALIGN uint8_t buf[MLDSA_K * MLDSA_POLYW1_PACKEDBYTES];
  MLD_ALIGN uint8_t rho[MLDSA_SEEDBYTES];
  MLD_ALIGN uint8_t mu[MLDSA_CRHBYTES];
  MLD_ALIGN uint8_t c[MLDSA_CTILDEBYTES];
  MLD_ALIGN uint8_t c2[MLDSA_CTILDEBYTES];
  mld_poly cp;
  mld_polyvecl mat[MLDSA_K], z;
  mld_polyveck t1, w1, tmp, h;
  ...
```

This causes a large amount of stack usage (>100K for ML-DSA-87) which may be unacceptable for some environments. Specifically, some environments might not have enough memory at all, while other may, but require heap rather than stack allocation. At the moment, mldsa-native provides no flexibility to specify the source of allocation.

**Task:** Provide a means to configure where the 'core objects' in the top-level APIs should be allocated from.

**Outline:**
1. As a first experiment, move all larger objects into a single `struct`, one per high-level API; it can be declared inline. The code-motion is straightforward, but CBMC will pose a challenge because we cannot use `object_whole` anymore.
2. Once 1. works, make minimal experiments moving objects in a `union` if they are not used at the same time. Again, check what CBMC does.
3. If 1. and 2. pan out, configure the allocation mechanism by providing a configurable macro for the struct allocation. By default, this should be implemented as a stack allocation. There will also need to be a 'free'-macro which for the stack case will be a no-op.
4. Implement a new example implementing the allocation via the heap. The allocation macro should call `alloc` and the free macro `free`.

PR #714, opened 2025-11-18
Add `crypto_sign_pk_from_sk` to top-level API

jakemas:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/44

### DIFF
--- a/mldsa/src/ct.h
+++ b/mldsa/src/ct.h
@@ -92,12 +92,18 @@ __contract__(ensures(return_value == 0)) { return (int64_t)mld_ct_get_optblocker
 static MLD_INLINE uint32_t mld_ct_get_optblocker_u32(void)
 __contract__(ensures(return_value == 0)) { return (uint32_t)mld_ct_get_optblocker_u64(); }
 
+static MLD_INLINE uint8_t mld_ct_get_optblocker_u8(void)
+__contract__(ensures(return_value == 0)) { return (uint8_t)mld_ct_get_optblocker_u64(); }
+
 /* Opt-blocker based implementation of value barriers */
 static MLD_INLINE int64_t mld_value_barrier_i64(int64_t b)
 __contract__(ensures(return_value == b)) { return (b ^ mld_ct_get_optblocker_i64()); }
 
 static MLD_INLINE uint32_t mld_value_barrier_u32(uint32_t b)
 __contract__(ensures(return_value == b)) { return (b ^ mld_ct_get_optblocker_u32()); }
+
+static MLD_INLINE uint8_t mld_value_barrier_u8(uint8_t b)
+__contract__(ensures(return_value == b)) { return (b ^ mld_ct_get_optblocker_u8()); }
 
 
 #else  /* !MLD_USE_ASM_VALUE_BARRIER */
@@ -109,6 +115,13 @@ __contract__(ensures(return_value == b))
 }
 
 static MLD_INLINE uint32_t mld_value_barrier_u32(uint32_t b)
+__contract__(ensures(return_value == b))
+{
+  __asm__("" : "+r"(b));
+  return b;
+}
+
+static MLD_INLINE uint8_t mld_value_barrier_u8(uint8_t b)
 __contract__(ensures(return_value == b))
 {
   __asm__("" : "+r"(b));
@@ -239,6 +252,48 @@ __contract__(
 
 #if !defined(__ASSEMBLER__)
 #include <string.h>
+
+/*************************************************
+ * Name:        mld_ct_memcmp
+ *
+ * Description: Compare two arrays for equality in constant time.
+ *
+ * Arguments:   const void *a: pointer to first byte array
+ *              const void *b: pointer to second byte array
+ *              size_t len:    length of the byte arrays
+ *
+ * Returns 0 if the byte arrays are equal, a non-zero value otherwise
+ **************************************************/
+static MLD_INLINE uint8_t mld_ct_memcmp(const void *a, const void *b,
+                                        const size_t len)
+__contract__(
+  requires(len <= UINT16_MAX)
+  requires(memory_no_alias(a, len))
+  requires(memory_no_alias(b, len))
+  ensures((return_value == 0) == forall(i, 0, len, (((const uint8_t *)a)[i] == ((const uint8_t *)b)[i])))
+)
+{
+  const uint8_t *a_bytes = (const uint8_t *)a;
+  const uint8_t *b_bytes = (const uint8_t *)b;
+  uint8_t r = 0, s = 0;
+  unsigned i;
+
+  for (i = 0; i < len; i++)
+  __loop__(
+    invariant(i <= len)
+    invariant((r == 0) == (forall(k, 0, i, (a_bytes[k] == b_bytes[k])))))
+  {
+    r |= a_bytes[i] ^ b_bytes[i];
+    /* s is useless, but prevents the loop from being aborted once r=0xff. */
+    s ^= a_bytes[i] ^ b_bytes[i];
+  }
+
+  /*
+   * XOR twice with s, separated by a value barrier, to prevent the compile
+   * from dropping the s computation in the loop.
+   */
+  return (uint8_t)((mld_value_barrier_u32((uint32_t)r) ^ s) ^ s);
+}
 
 /*************************************************
  * Name:        mld_zeroize
--- a/mldsa/src/sign.c
+++ b/mldsa/src/sign.c
@@ -28,6 +28,7 @@
 #include <string.h>
 
 #include "cbmc.h"
+#include "ct.h"
 #include "debug.h"
 #include "packing.h"
 #include "poly.h"
@@ -48,6 +49,8 @@
 #define mld_H MLD_ADD_PARAM_SET(mld_H)
 #define mld_attempt_signature_generation \
   MLD_ADD_PARAM_SET(mld_attempt_signature_generation)
+#define mld_compute_t0_t1_tr_from_sk_components \
+  MLD_ADD_PARAM_SET(mld_compute_t0_t1_tr_from_sk_components)
 /* End of parameter set namespacing */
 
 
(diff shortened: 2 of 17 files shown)


mldsa/src/sign.h
```diff
@@ -64,6 +64,7 @@
   MLD_NAMESPACE_KL(verify_pre_hash_shake256)
 #define mld_prepare_domain_separation_prefix \
   MLD_NAMESPACE_KL(prepare_domain_separation_prefix)
+#define pk_from_sk MLD_NAMESPACE_KL(pk_from_sk)
```
hanno-becker: Nit: Add `mld_` prefix
```suggestion
#define mld_pk_from_sk MLD_NAMESPACE_KL(pk_from_sk)
```

mldsa/src/sign.h
```diff
@@ -686,4 +687,25 @@ __contract__(
...
+/*************************************************
+ * Name:        pk_from_sk
+ *
+ * Description: Derives public key from secret key with validation.
+ *              Checks that t0 and tr stored in sk match recomputed values.
+ *
+ * Arguments:   - uint8_t pk[CRYPTO_PUBLICKEYBYTES]: output public key
+ *              - const uint8_t sk[CRYPTO_SECRETKEYBYTES]: input secret key
+ *
+ * Returns 0 on success, -1 if validation fails (corrupted secret key)
```
hanno-becker: Is there a reference to FIPS204 one can provide here?

hanno-becker: Nit: s/corrupted/invalid

mldsa/src/sign.h
```diff
@@ -686,4 +687,25 @@ __contract__(
...
+ **************************************************/
+MLD_MUST_CHECK_RETURN_VALUE
+MLD_EXTERNAL_API
+int pk_from_sk(uint8_t pk[CRYPTO_PUBLICKEYBYTES],
```
mkannwischer: If this is supposed to be a public API, then we should add it to mldsa_native.h

mldsa/src/sign.h
```diff
@@ -686,4 +687,25 @@ __contract__(
   ensures(return_value <= MLD_DOMAIN_SEPARATION_MAX_BYTES)
 );
 
+/*************************************************
+ * Name:        pk_from_sk
```
mkannwischer: Should this be renamed to `crypto_sign_pk_from_sk` to align with the other APIs?

mldsa/src/sign.c
```diff
@@ -174,6 +177,84 @@ __contract__(
...
+  requires(memory_no_alias(t0, sizeof(mld_polyveck)))
+  requires(memory_no_alias(t1, sizeof(mld_polyveck)))
+  requires(memory_no_alias(tr, MLDSA_TRBYTES))
+  requires(memory_no_alias(rho, MLDSA_SEEDBYTES))
+  requires(memory_no_alias(s1, sizeof(mld_polyvecl)))
+  requires(memory_no_alias(s2, sizeof(mld_polyveck)))
+  requires(forall(l0, 0, MLDSA_L, array_bound(s1->vec[l0].coeffs, 0, MLDSA_N, MLD_POLYETA_UNPACK_LOWER_BOUND, MLDSA_ETA + 1)))
+  requires(forall(k0, 0, MLDSA_K, array_bound(s2->vec[k0].coeffs, 0, MLDSA_N, MLD_POLYETA_UNPACK_LOWER_BOUND, MLDSA_ETA + 1)))
+  assigns(object_whole(t0))
+  assigns(object_whole(t1))
```
mkannwischer: In light of https://github.com/pq-code-package/mldsa-native/issues/716, this should be changed to memory_slice

mldsa/src/sign.c
```diff
@@ -174,6 +177,84 @@ __contract__(
...
+  assigns(memory_slice(tr, MLDSA_TRBYTES))
+  ensures(forall(k1, 0, MLDSA_K, array_bound(t0->vec[k1].coeffs, 0, MLDSA_N, -(1<<(MLDSA_D-1)) + 1, (1<<(MLDSA_D-1)) + 1)))
+  ensures(forall(k2, 0, MLDSA_K, array_bound(t1->vec[k2].coeffs, 0, MLDSA_N, 0, 1 << 10)))
+)
+{
+  mld_polyvecl mat[MLDSA_K], s1hat;
+  mld_polyveck t;
+  uint8_t pk_tmp[CRYPTO_PUBLICKEYBYTES];
```
mkannwischer: This design requires re-packing the public key in both call sites afterwards. Can we avoid that?
Maybe just add the packed pk as an additional output?

mldsa/src/sign.c
```diff
@@ -1130,6 +1184,67 @@ size_t mld_prepare_domain_separation_prefix(
...
+  int res;
+
+  /* Unpack secret key */
+  mld_unpack_sk(rho, tr, key, &t0, &s1, &s2, sk);
+
+  /* Declassify secret components before validation */
+  MLD_CT_TESTING_DECLASSIFY(tr, MLDSA_TRBYTES);
+  MLD_CT_TESTING_DECLASSIFY(&t0, sizeof(t0));
+  MLD_CT_TESTING_DECLASSIFY(&s1, sizeof(s1));
+  MLD_CT_TESTING_DECLASSIFY(&s2, sizeof(s2));
```
mkannwischer: Who we need to declassify these? Can we remove this and instead declassify the result of the ct_memcmp? That's sort of the point of it.

mkannwischer (changes requested):
Thanks @jakemas! 
Two blockers: 
 - The function needs to be added to mldsa_native.h
 - I think some of the declassifications need to be removed

I also wonder if we should rename the function to `crypto_sign_pk_from_sk` to align with other functions?

mldsa/src/sign.h
jakemas: Unfortunately not, there is no mention of this in FIPS 204.

mldsa/src/sign.h
jakemas: Changed to `crypto_sign_pk_from_sk` as Matthias suggested

mldsa/src/sign.h
jakemas: changed!

mldsa/src/sign.h
jakemas: added!

mldsa/src/sign.h
jakemas: done!

mldsa/src/sign.c
jakemas: ok done

mldsa/src/sign.c
jakemas: Ok, added PK as additional output

mldsa/src/sign.c
jakemas: Ok -- wasn't exactly sure on this part, thanks!

mkannwischer (approved):
Thanks @jakemas for the changes! This looks good to me now.

mldsa/src/sign.h
hanno-becker: No, this is still as before

mldsa/src/sign.c
```diff
@@ -1130,6 +1184,62 @@ size_t mld_prepare_domain_separation_prefix(
...
+  res = 0;
+
+cleanup:
+  /* @[FIPS204, Section 3.6.3] Destruction of intermediate values. */
+  mld_zeroize(&s1, sizeof(s1));
+  mld_zeroize(&s2, sizeof(s2));
+  mld_zeroize(&t0, sizeof(t0));
+  mld_zeroize(&t0_computed, sizeof(t0_computed));
+  mld_zeroize(key, sizeof(key));
+  mld_zeroize(tr_computed, sizeof(tr_computed));
```
hanno-becker: Why are `tr`, `rho`, and `t1` omitted?

mkannwischer: and rho too

mldsa/src/sign.h
mkannwischer: I fixed it

mldsa/src/sign.c
```diff
@@ -1130,6 +1184,64 @@ size_t mld_prepare_domain_separation_prefix(
...
+
+  /* Unpack secret key */
+  mld_unpack_sk(rho, tr, key, &t0, &s1, &s2, sk);
+
+  /* Recompute t0, t1, tr, and pk from rho, s1, s2 */
+  mld_compute_t0_t1_tr_from_sk_components(&t0_computed, &t1, tr_computed, pk,
+                                          rho, &s1, &s2);
+
+  /* Declassify public key */
+  MLD_CT_TESTING_DECLASSIFY(pk, CRYPTO_PUBLICKEYBYTES);
```
hanno-becker: Nit: This declassification could (should?) be moved to _after_ the validation.

test/test_mldsa.c
```diff
@@ -151,6 +151,65 @@ static int test_sign_pre_hash(void)
...
+
+  /* Test with corrupted t0 in secret key - should fail validation */
+  memcpy(sk_corrupted, sk, CRYPTO_SECRETKEYBYTES);
+  /* Corrupt a byte in the t0 portion of the secret key */
+  sk_corrupted[MLDSA_SEEDBYTES + MLDSA_TRBYTES + MLDSA_SEEDBYTES + 10] ^= 1;
+
+  rc = crypto_sign_pk_from_sk(pk_derived, sk_corrupted);
+
+  /* Constant time: Declassify to check result */
+  MLD_CT_TESTING_DECLASSIFY(&rc, sizeof(int));
```
hanno-becker, line 183: Why do we need this declassification here, but not above?

test/test_mldsa.c
```diff
@@ -151,6 +151,65 @@ static int test_sign_pre_hash(void)
...
+  /* Test with corrupted tr in secret key - should fail validation */
+  memcpy(sk_corrupted, sk, CRYPTO_SECRETKEYBYTES);
+  /* Corrupt a byte in the tr portion of the secret key */
+  /* tr starts at offset 2 * MLDSA_SEEDBYTES (after rho and key) */
+  sk_corrupted[2 * MLDSA_SEEDBYTES + 10] ^= 1;
+
+  rc = crypto_sign_pk_from_sk(pk_derived, sk_corrupted);
+
+  /* Constant time: Declassify to check result */
+  MLD_CT_TESTING_DECLASSIFY(&rc, sizeof(int));
```
hanno-becker, line 200: Same

test/test_mldsa.c
```diff
@@ -151,6 +151,65 @@ static int test_sign_pre_hash(void)
...
+  uint8_t pk_derived[CRYPTO_PUBLICKEYBYTES];
+  uint8_t sk[CRYPTO_SECRETKEYBYTES];
+  uint8_t sk_corrupted[CRYPTO_SECRETKEYBYTES];
+  int rc;
+
+  /* Generate a keypair */
+  CHECK(crypto_sign_keypair(pk, sk) == 0);
+
+  /* Derive public key from secret key */
+  CHECK(crypto_sign_pk_from_sk(pk_derived, sk) == 0);
```
hanno-becker, line 166: @mkannwischer Is there a way to test/assert in valgrind that certain memory _is_ (still) classified? For failing `crypto_sign_pk_from_sk` we may want to check that `pk` is still secret (see comment above).

mldsa/src/sign.c
hanno-becker: Actually... if the function fails, `pk` should probably be 0'ed.

hanno-becker (changes requested):
One question regarding declassification/zeroing of `pk` in case of failure remaining.

This change will make https://github.com/pq-code-package/mldsa-native/issues/713 harder, unfortunately, but we can cross that bridge when we get to it.

test/test_mldsa.c
mkannwischer: I don't think we should this is a good idea. The pk should be a public output regardless of what input was passed to the function. We should zero out the public key and keep the declassification.

hanno-becker: That's OK by me as well.

mldsa/src/sign.c
jakemas: Ok, what would you like here? Move the declassification, remove it, and 0 instead?

mldsa/src/sign.c
jakemas: they are now 
https://github.com/pq-code-package/mldsa-native/pull/714/files#diff-10cdaf2571bc06c70997f7cc5aa75e472dbf5054f0cb71e548c7cdbd79a53f3dR1233-R1241

test/test_mldsa.c
jakemas: would you like it added above or removed here?

test/test_mldsa.c
jakemas: would you like it added above or removed here?

test/test_mldsa.c
jakemas: Ok so the consensus is that we should zero out the public key and keep the declassification. Is this the same for `rc` also in your above comments Hanno?

jakemas:
> One question regarding declassification/zeroing of `pk` in case of failure remaining.
> 
> This change will make #713 harder, unfortunately, but we can cross that bridge when we get to it.

Ok, let me know what you think -- I've added the zero of `pk` in failure case. If I have the declassifications in the incorrect place, please let me know specifically which ones and exactly where you would like me to move them and I'll do it :)

mldsa/src/sign.c
```diff
@@ -1131,6 +1185,62 @@ size_t mld_prepare_domain_separation_prefix(
...
+  res = 0;
+
+cleanup:
+  /* @[FIPS204, Section 3.6.3] Destruction of intermediate values. */
+  mld_zeroize(&s1, sizeof(s1));
+  mld_zeroize(&s2, sizeof(s2));
+  mld_zeroize(&t0, sizeof(t0));
+  mld_zeroize(&t0_computed, sizeof(t0_computed));
+  mld_zeroize(key, sizeof(key));
+  mld_zeroize(tr_computed, sizeof(tr_computed));
```
hanno-becker: `tr` is not zeroized here

mldsa/src/sign.c
hanno-becker: In `cleanup`, I'd suggest to zeroize `pk` if `res` is not 0, and _then_ unconditionally declassify it.

mldsa/src/sign.c
hanno-becker: And what about `t1`?

mldsa/src/sign.c
hanno-becker: I don't think all of those have been added, can you check?

mldsa/src/sign.c
```diff
@@ -1131,6 +1185,62 @@ size_t mld_prepare_domain_separation_prefix(
...
+
+  /* Validate tr using constant-time comparison */
+  res = mld_ct_memcmp(tr, tr_computed, MLDSA_TRBYTES);
+  /* Declassify comparison result */
+  MLD_CT_TESTING_DECLASSIFY(&res, sizeof(res));
+  if (res != 0)
+  {
+    res = -1;
+    goto cleanup;
+  }
```
hanno-becker: Suggestion: This can be streamlined and further hardened 
```c
res0 = mld_ct_memcmp(&t0, &t0_computed, sizeof(mld_polyveck));
res1 = mld_ct_memcmp(tr, tr_computed, MLDSA_TRBYTES);
res = mld_value_barrier_u8(res0 | res1);

/* Declassify the final result of the validity check. */
MLD_CT_TESTING_DECLASSIFY(&res, sizeof(res));
if (res != 0)
{
  mld_zeroize(pk, CRYPTO_PUBLICKEYBYTES);
}

cleanup:
   ...
```

Also, it needs to be documented that the function leaks the validity of the `sk`.

jakemas: Ok, this required the addition of `mld_value_barrier_u8` and `ct_get_optblocker_u8` funcitons and contracts, so I made this change, and added these.

mldsa/src/sign.c
jakemas: https://github.com/pq-code-package/mldsa-native/pull/714/files#diff-10cdaf2571bc06c70997f7cc5aa75e472dbf5054f0cb71e548c7cdbd79a53f3dR1219-R1228
```
  /* @[FIPS204, Section 3.6.3] Destruction of intermediate values. */
  mld_zeroize(rho, sizeof(rho));
  mld_zeroize(tr, sizeof(tr));
  mld_zeroize(tr_computed, sizeof(tr_computed));
  mld_zeroize(key, sizeof(key));
  mld_zeroize(&s1, sizeof(s1));
  mld_zeroize(&s2, sizeof(s2));
  mld_zeroize(&t0, sizeof(t0));
  mld_zeroize(&t0_computed, sizeof(t0_computed));
  mld_zeroize(&t1, sizeof(t1));
  ```

File: mldsa/src/sign.c (hunk 9 above)
Comment (jakemas) [reply mldsa/src/sign.c]: done

File: mldsa/src/ct.h [hunk 15]
```diff
@@ -240,6 +253,48 @@ __contract__(
...
+    r |= a_bytes[i] ^ b_bytes[i];
+    /* s is useless, but prevents the loop from being aborted once r=0xff. */
+    s ^= a_bytes[i] ^ b_bytes[i];
+  }
+
+  /*
+   * XOR twice with s, separated by a value barrier, to prevent the compile
+   * from dropping the s computation in the loop.
+   */
+  return (uint8_t)((mld_value_barrier_u32((uint32_t)r) ^ s) ^ s);
```
hanno-becker, line 295: This should not have been merged. The value barrier is in the wrong place, allowing the compiler to elide `s`, and somehow the CT mask from mlkem-native was dropped.

Issue #722, opened 2025-11-19
`make quickcheck CFLAGS="-std=c11"` fails


mkannwischer:
We usually pass CFLAGS as environment variables:
```
CFLAGS="-std=c11" make quickcheck
```
works perfectly fine 

However, it should also work to pass it as an argument to make:

```
make quickcheck CFLAGS="-std=c11"
```

Currently this leads to an obscure error when running the ML-DSA-65 KAT tests. 
```
set -o pipefail;  test/build/mldsa65/bin/gen_KAT65 | shasum -a 256 | cut -d " " -f 1 | xargs ./META.sh ML-DSA-65 kat-sha256
META.yml ML-DSA-65 kat-sha256: FAIL (2ff0ddcd0dc08b746aa04853d6f84c82c6c8ac38783c9061aed78e29c1698ae5 != 33e7eb7e3b4965fc8b8274ebf8a222c519008ec242b3f753d1bc240f1ee3cb1f)
make: *** [Makefile:58: run_kat_65] Error 123
```

The ML-DSA-44 KATs work fine, so I suspect it it compiling the wrong parameter set.

PR #725, opened 2025-11-20
Make: Support CFLAGS as both env var and make argument

mkannwischer:
We usually pass arguments to make through environment variables: CFLAGS="-std=c11" make quickcheck

It currently does not work to pass it as a make argument: make quickcheck CFLAGS="-std=c11"

This commit fixes that and extends the C90 CI to test both variants.

- Resolves https://github.com/pq-code-package/mldsa-native/issues/722

I stapled on a cleanup of the Makefiles to align with mlkem-native

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,12 +32,6 @@ endif
 
 W := $(EXEC_WRAPPER)
 
-# Detect available SHA256 command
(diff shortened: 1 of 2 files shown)


hanno-becker (dismissed):

hanno-becker (changes requested):
Can you update this in line with https://github.com/pq-code-package/mlkem-native/pull/1316?

mkannwischer:
This breaks benchmarking. See https://github.com/pq-code-package/mlkem-native/pull/1327.

PR #734, opened 2025-11-21
CBMC: Allow specification of per-proof timeouts

### LINKED ISSUE #733: CBMC: Allow specification of timeouts
When CBMC proofs blow up in CI, it is very tedious to find out which function(s) are actually the problematic ones. One needs to crawl through the log and reverse engineer from the presence of _successful_ steps which functions have _not_ completed.

**Task:** 
- [ ] Modify `run-cbmc-proofs.py` so one can pass a per-proof timeout
- [ ] Set the proof timeout in CI to 30min (for now) and report "TIMEOUT" in the result table upon timeout, so one can always inspect the table in the CI results to see which functions were problematic.

30min is larger than what we'd want in CI, but it is still useful to know whether a proof works in principle, but is just slow. It is also good enough to prevent the CI from running for hours and wasting money on a failing / inacceptably long proof.

**Acceptance Criteria:**
- [ ] `run-cbmc-proofs.py` has an additional `--per-proof-timeout argument which defaults to 1800s (30min) and bounds the amount of time spent on a _fixed_ proof.
- [ ] When a proof times out, the corresponding entry in the summary table is "TIMEOUT" rather then "XXX s" upon success.
- [ ] If at least one proof times out, the return value of `run-cbmc-proofs.py` is non-zero.
- [ ] `tests cbmc` has an additional `--per-proof-timeout` argument which is passed on to `run-cbmc-proofs.py`
- [ ] CI is adjusted to CBMC proofs run with timeout of 30min; this should be spelled out explicitly even though it's also the implicit default.
- [ ] A local test of `tests cbmc --per-proof-timeout=15` has been run to confirm that the mechanism works, and that the return value upon _any_ timeout is non-zero.

hanno-becker:
This commit adds support for per-proof timeouts in CBMC proofs to prevent CI from running for hours on failing or slow proofs. The implementation leverages Litani's existing timeout mechanism via the CBMC_TIMEOUT environment variable.

- Add `--per-proof-timeout` argument to `run-cbmc-proofs.py` (default: 1800s)
- Pass timeout to Litani via `CBMC_TIMEOUT` environment variable
- Modify `lib/summarize.py` to detect `timeout_reached` flag and display "timeout" instead.
- Add `--per-proof-timeout` argument to `tests cbmc` command
- Set explicit 30-minute timeout in CI workflow

Fixes #733

### DIFF
--- a/.github/actions/cbmc/action.yml
+++ b/.github/actions/cbmc/action.yml
@@ -54,5 +54,5 @@ runs:
         shell: ${{ env.SHELL }}
         run: |
           echo "::group::cbmc_${{ inputs.mldsa_parameter_set }}"
-          tests cbmc --mldsa_parameter_set ${{ inputs.mldsa_parameter_set }};
+          tests cbmc --mldsa_parameter_set ${{ inputs.mldsa_parameter_set }} --per-proof-timeout 1800;
           echo "::endgroup::"
--- a/proofs/cbmc/lib/summarize.py
+++ b/proofs/cbmc/lib/summarize.py
@@ -90,17 +90,31 @@ def _get_status_and_proof_summaries(run_dict):
     for proof_pipeline in run_dict["pipelines"]:
         if proof_pipeline["name"] == "print_tool_versions":
             continue
-        status_pretty_name = proof_pipeline["status"].title().replace("_", " ")
+
+        # Check if any job timed out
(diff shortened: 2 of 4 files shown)


mkannwischer (approved):
Nice addition! LGTM and appears to work well in #730.

Issue #732, opened 2025-11-21
Port: Hoist default C backend into separate functions


willieyz:
- Related: #607 
- Blocks: #470
- Reference from mlkem-native: https://github.com/pq-code-package/mlkem-native/pull/1195

We have completed PR #607. The next step is to separate the C backend functions.  
The following is the plan for this issue:

- Purpose: this porting aims to separate the native backend functions into two sets: fallback C functions and native functions.

- Steps: This task is based on the [runtime-dispatch PR #607](https://github.com/pq-code-package/mldsa-native/pull/607).  
  - The following functions will be touched:
    - `mld_ntt_native`, `mld_intt_native`, `mld_poly_permute_bitrev_to_custom`
    - `mld_rej_uniform_native`, `mld_rej_uniform_eta2_native`, `mld_rej_uniform_eta4_native`
    - `mld_poly_decompose_32_native`, `mld_poly_decompose_88_native`
    - `mld_poly_caddq_native`
    - `mld_poly_use_hint_32_native`, `mld_poly_use_hint_88_native`
    - `mld_poly_chknorm`
    - `mld_polyz_unpack_17_native`, `mld_polyz_unpack_19_native`
    - `mld_poly_pointwise_montgomery_native`
    - `mld_polyvecl_pointwise_acc_montgomery_l4_native`, `l5_native`, `l7_native`
    - **FIPS202**:  
      - `mld_keccak_f1600_x1_native`  
      - `mld_keccak_f1600_x4_native[x2_v84a]`  
      - `mld_keccak_f1600_x4_native[x4_v8a]`  
      - `mld_keccak_f1600_x4_native[x4_v8a_v84a]`  
      - `mld_keccakf1600x4_permute`

> Note: If any new C functions are created during this task, corresponding CBMC proof contracts should also be implemented.
- Tests
  - Run test scripts with all argument variations, using both `static ON` and `OFF` configurations.
  - Run `quickcheck` with different configurations (`static ON` and `OFF`).
  - Perform CBMC proof tests for each new C function.

PR #735, opened 2025-11-24
Port: Hoist default C backend into separate functions

willieyz:
- Resolves: #732  
- Purpose: This porting aims to separate the native backend functions into two distinct sets:  
  - Fallback C backend functions  
  - Native backend functions  
  This work is based on the runtime-dispatch changes introduced in [PR #607](https://github.com/pq-code-package/mldsa-native/pull/607).

- Steps:
  - For each function improved by runtime dispatch in [PR #607](https://github.com/pq-code-package/mldsa-native/pull/607).:
    - Separate a additional fallback C backend function from the original implementation.
    - Add a CBMC proof for the C backend function.
    - Update the existing original implementation's CBMC `USE_FUNCTION_CONTRACTS` with the C backend function.
  
- Tests:
  - Run all test scripts with all argument variations under both static ON and static OFF configurations.
  - Run all examples: `tests examples`.
  - Run all relevant CBMC proofs for each modified or newly added function.

- The following is the checking table for fallback C backend and native backend:

| Original | Native | C backend |
|:------:|:------:|:------:|
| `mld_poly_ntt`   | `mld_ntt_native`   | `mld_poly_ntt_c`   |
| `mld_poly_invntt_tomont`   | `mld_intt_native`   | `mld_poly_invntt_tomont_c`   |
|`mld_polymat_permute_bitrev_to_custom`|`mld_poly_permute_bitrev_to_custom`|`((void)mat);`<br> `/* Nothing to do */ `|
|`mld_rej_uniform`|`mld_rej_uniform_native`|`mld_rej_uniform_c`|
|`mld_rej_eta`|`mld_rej_uniform_eta2_native`<br> `mld_rej_uniform_eta4_native`|`mld_rej_eta_c`|
|mld_poly_decompose|`mld_poly_decompose_32_native`<br>`mld_poly_decompose_88_native`|`mld_poly_decompose_c`|
|mld_poly_caddq|`mld_poly_caddq_native`|`mld_poly_caddq_c`|
|`mld_poly_use_hint`|`mld_poly_use_hint_32_native`<br>`mld_poly_use_hint_88_native`|`mld_poly_use_hint_c`|
|`mld_poly_chknorm`|`mld_poly_chknorm_native`|`mld_poly_chknorm_c`|
|`mld_polyz_unpack`|`mld_polyz_unpack_17_native`<br>`mld_polyz_unpack_19_native`|`mld_polyz_unpack_c`|
|`mld_poly_pointwise_montgomery`|`mld_poly_pointwise_montgomery_native`|`mld_poly_pointwise_montgomery_c`|
|`mld_polyvecl_pointwise_acc_montgomery`|`mld_polyvecl_pointwise_acc_montgomery_l4_native`<br>`mld_polyvecl_pointwise_acc_montgomery_l5_native`<br>`mld_polyvecl_pointwise_acc_montgomery_l7_native`|`mld_polyvecl_pointwise_acc_montgomery_c`|
|`mld_keccakf1600_permute`|`mld_keccak_f1600_x1_native`|`mld_keccakf1600_permute_c`|
|`mld_keccakf1600x4_permute`|`mld_keccak_f1600_x2_v84a_asm` * 2<br>`mld_keccak_f1600_x4_scalar_v8a_hybrid_asm` <br>`mld_keccak_f1600_x4_scalar_v8a_v84a_hybrid_asm`|`mld_keccakf1600_permute` * 4|

### DIFF
--- a/mldsa/src/poly.c
+++ b/mldsa/src/poly.c
@@ -648,50 +648,18 @@ void mld_polyt0_unpack(mld_poly *r, const uint8_t a[MLDSA_POLYT0_PACKEDBYTES])
                    (1 << (MLDSA_D - 1)) + 1);
 }
 
-/* Reference: explicitly checks the bound B to be <= (MLDSA_Q - 1) / 8).
- * This is unnecessary as it's always a compile-time constant.
- * We instead model it as a precondition.
- * Checking the bound is performed using a conditional arguing
- * that it is okay to leak which coefficient violates the bound (while the
- * coefficient itself must remain secret).
- * We instead perform everything in constant-time.
- * Also it is sufficient to check that it is smaller than
- * MLDSA_Q - REDUCE32_RANGE_MAX > (MLDSA_Q - 1) / 8).
- */
-MLD_INTERNAL_API
-uint32_t mld_poly_chknorm(const mld_poly *a, int32_t B)
+MLD_STATIC_TESTABLE uint32_t mld_poly_chknorm_c(const mld_poly *a, int32_t B)
+__contract__(
+  requires(memory_no_alias(a, sizeof(mld_poly)))
+  requires(0 <= B && B <= MLDSA_Q - REDUCE32_RANGE_MAX)
+  requires(array_bound(a->coeffs, 0, MLDSA_N, -REDUCE32_RANGE_MAX, REDUCE32_RANGE_MAX))
+  ensures(return_value == 0 || return_value == 0xFFFFFFFF)
+  ensures((return_value == 0) == array_abs_bound(a->coeffs, 0, MLDSA_N, B))
+)
 {
   unsigned int i;
   uint32_t t = 0;
   mld_assert_bound(a->coeffs, MLDSA_N, -REDUCE32_RANGE_MAX, REDUCE32_RANGE_MAX);
-#if defined(MLD_USE_NATIVE_POLY_CHKNORM)
-  {
-    /* TODO: proof */
-    int ret;
-    int success;
-    /* The native backend returns 0 if all coefficients are within the bound,
-     * 1 if at least one coefficient exceeds the bound, and
-     * -1 (MLD_NATIVE_FUNC_FALLBACK) if the platform does not have the
-     * required capabilities to run the native function.
-     */
-    ret = mld_poly_chknorm_native(a->coeffs, B);
-
-    success = (ret != MLD_NATIVE_FUNC_FALLBACK);
-    /* Constant-time: It would be fine to leak the return value of chknorm
-     * entirely (as it is fine to leak if any coefficient exceeded the bound or
-     * not). However, it is cleaner to perform declassification in sign.c.
-     * Hence, here we only declassify if the native function returned
-     * MLD_NATIVE_FUNC_FALLBACK or not (which solely depends on system
-     * capabilities).
-     */
-    MLD_CT_TESTING_DECLASSIFY(&success, sizeof(int));
-    if (success)
-    {
-      /* Convert 0 / 1 to 0 / 0xFFFFFFFF here */
-      return 0U - (uint32_t)ret;
-    }
-  }
-#endif /* MLD_USE_NATIVE_POLY_CHKNORM */
   for (i = 0; i < MLDSA_N; ++i)
   __loop__(
     invariant(i <= MLDSA_N)
@@ -720,6 +688,49 @@ uint32_t mld_poly_chknorm(const mld_poly *a, int32_t B)
   }
 
   return t;
+}
+
+/* Reference: explicitly checks the bound B to be <= (MLDSA_Q - 1) / 8).
+ * This is unnecessary as it's always a compile-time constant.
+ * We instead model it as a precondition.
+ * Checking the bound is performed using a conditional arguing
+ * that it is okay to leak which coefficient violates the bound (while the
+ * coefficient itself must remain secret).
+ * We instead perform everything in constant-time.
+ * Also it is sufficient to check that it is smaller than
+ * MLDSA_Q - REDUCE32_RANGE_MAX > (MLDSA_Q - 1) / 8).
+ */
+MLD_INTERNAL_API
+uint32_t mld_poly_chknorm(const mld_poly *a, int32_t B)
+{
+#if defined(MLD_USE_NATIVE_POLY_CHKNORM)
+  /* TODO: proof */
+  int ret;
+  int success;
+  mld_assert_bound(a->coeffs, MLDSA_N, -REDUCE32_RANGE_MAX, REDUCE32_RANGE_MAX);
+  /* The native backend returns 0 if all coefficients are within the bound,
+   * 1 if at least one coefficient exceeds the bound, and
+   * -1 (MLD_NATIVE_FUNC_FALLBACK) if the platform does not have the
+   * required capabilities to run the native function.
+   */
+  ret = mld_poly_chknorm_native(a->coeffs, B);
+
+  success = (ret != MLD_NATIVE_FUNC_FALLBACK);
+  /* Constant-time: It would be fine to leak the return value of chknorm
+   * entirely (as it is fine to leak if any coefficient exceeded the bound or
+   * not). However, it is cleaner to perform declassification in sign.c.
+   * Hence, here we only declassify if the native function returned
+   * MLD_NATIVE_FUNC_FALLBACK or not (which solely depends on system
+   * capabilities).
+   */
+  MLD_CT_TESTING_DECLASSIFY(&success, sizeof(int));
+  if (success)
+  {
+    /* Convert 0 / 1 to 0 / 0xFFFFFFFF here */
+    return 0U - (uint32_t)ret;
+  }
+#endif /* MLD_USE_NATIVE_POLY_CHKNORM */
+  return mld_poly_chknorm_c(a, B);
 }
 
 #else  /* !MLD_CONFIG_MULTILEVEL_NO_SHARED */
--- a/proofs/cbmc/poly_chknorm_c/Makefile
+++ b/proofs/cbmc/poly_chknorm_c/Makefile
@@ -4,11 +4,11 @@
 include ../Makefile_params.common
 
 HARNESS_ENTRY = harness
-HARNESS_FILE =poly_chknorm_harness
+HARNESS_FILE =poly_chknorm_c_harness
 
 # This should be a unique identifier for this proof, and will appear on the
 # Litani dashboard. It can be human-readable and contain spaces if you wish.
-PROOF_UID = poly_chknorm
+PROOF_UID = poly_chknorm_c
 
 DEFINES +=
 INCLUDES +=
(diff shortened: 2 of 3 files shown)


proofs/cbmc/poly_ntt/Makefile
```diff
@@ -20,7 +20,7 @@ PROOF_SOURCES += $(PROOFDIR)/$(HARNESS_FILE).c
 PROJECT_SOURCES += $(SRCDIR)/mldsa/src/poly.c
 
 CHECK_FUNCTION_CONTRACTS=$(MLD_NAMESPACE)poly_ntt
-USE_FUNCTION_CONTRACTS=$(MLD_NAMESPACE)ntt
+USE_FUNCTION_CONTRACTS=$(MLD_NAMESPACE)ntt mld_poly_ntt_c
```
mkannwischer: This should not include the `ntt` as that code was moved.

mldsa/src/poly.c
```diff
@@ -142,6 +142,19 @@ void mld_poly_shiftl(mld_poly *a)
   mld_assert_bound(a->coeffs, MLDSA_N, 0, MLDSA_Q);
 }
 
+static void mld_poly_ntt_c(mld_poly *a)
```
mkannwischer: To be able to implement the unit tests later the `static` need to be changed to `MLD_STATIC_TESTABLE`. Maybe we already do that now so it does not have to be changed later.

proofs/cbmc/poly_ntt/Makefile
willieyz: Hello, @mkannwischer , thank you for your review,
I had removed this from `poly_ntt`'s CBMC Makefile,  thank you!

mldsa/src/poly.c
willieyz: Done, now all *_c function had a `MLD_STATIC_TESTABLE` instead of `static`.
Thank you for your review.

mldsa/src/polyvec.c
```diff
@@ -323,11 +325,55 @@ void mld_polyvecl_pointwise_poly_montgomery(mld_polyvecl *r, const mld_poly *a,
   mld_assert_abs_bound_2d(r->vec, MLDSA_L, MLDSA_N, MLDSA_Q);
 }
 
+MLD_STATIC_TESTABLE void mld_polyvecl_pointwise_acc_montgomery_c(
+    mld_poly *w, const mld_polyvecl *u, const mld_polyvecl *v)
+
```
mkannwischer: nit: spurious new line.

mldsa/src/poly.c
```diff
@@ -142,6 +142,19 @@ void mld_poly_shiftl(mld_poly *a)
...
 
+MLD_STATIC_TESTABLE void mld_poly_ntt_c(mld_poly *a)
+__contract__(
+  requires(memory_no_alias(a, sizeof(mld_poly)))
+  requires(array_abs_bound(a->coeffs, 0, MLDSA_N, MLDSA_Q))
+  assigns(memory_slice(a, sizeof(mld_poly)))
+  ensures(array_abs_bound(a->coeffs, 0, MLDSA_N, MLD_NTT_BOUND))
+)
+{
+  mld_assert_abs_bound(a->coeffs, MLDSA_N, MLDSA_Q);
```
hanno-becker, line 163: We currently check the pre-bound _twice_, which will make esp. cross testing slower. I think it makes sense to have the assertions here and not at the higher level, since we want them to be present in forthcoming unit tests; however, can we move the additional bounds check in `mld_poly_ntt` into the `MLD_USE_NATIVE_NTT` branch?

mldsa/src/poly.c
```diff
@@ -142,6 +142,19 @@ void mld_poly_shiftl(mld_poly *a)
...
+  mld_ntt(a->coeffs);
```
hanno-becker, line 164: Why do we have the indirection `mld_poly_ntt_c -> mld_ntt`? Can we just inline `mld_ntt` here?

mldsa/src/poly.c
```diff
@@ -186,6 +186,20 @@ void mld_poly_ntt(mld_poly *a)
...
+MLD_STATIC_TESTABLE void mld_poly_invntt_tomont_c(mld_poly *a)
+__contract__(
+  requires(memory_no_alias(a, sizeof(mld_poly)))
+  requires(array_abs_bound(a->coeffs, 0, MLDSA_N, MLDSA_Q))
+  assigns(memory_slice(a, sizeof(mld_poly)))
+  ensures(array_abs_bound(a->coeffs, 0, MLDSA_N, MLD_INTT_BOUND))
+)
+{
+  mld_assert_abs_bound(a->coeffs, MLDSA_N, MLDSA_Q);
+  mld_invntt_tomont(a->coeffs);
```
hanno-becker, line 193: Same comments as for the fwd NTT

mldsa/src/poly.c
```diff
@@ -237,16 +264,7 @@ void mld_poly_pointwise_montgomery(mld_poly *c, const mld_poly *a,
...
-  __loop__(
-    invariant(i <= MLDSA_N)
-    invariant(array_abs_bound(c->coeffs, 0, i, MLDSA_Q))
-  )
-  {
-    c->coeffs[i] = mld_montgomery_reduce((int64_t)a->coeffs[i] * b->coeffs[i]);
-  }
-
-  mld_assert_abs_bound(c->coeffs, MLDSA_N, MLDSA_Q);
+  mld_poly_pointwise_montgomery_c(c, a, b);
```
hanno-becker, line 258: The use of assertions is not consistent here. We should either a) use pre/post assertions _everywhere_ -- in which case we should have a post-condition assertion here as well, or b) we allow ourselves -- for test performance -- to omit 'obviously' redundant pre/post conditions, such as when we call from `foo` straight into `foo_c`.

WDYT @mkannwischer @willieyz

mldsa/src/poly.c
```diff
@@ -289,7 +289,57 @@ void mld_poly_power2round(mld_poly *a1, mld_poly *a0, const mld_poly *a)
   mld_assert_bound(a1->coeffs, MLDSA_N, 0, ((MLDSA_Q - 1) / MLD_2_POW_D) + 1);
 }
 
+#define POLY_UNIFORM_NBLOCKS \
```
hanno-becker, line 283: All macros should be namespaced

hanno-becker: Ok, this is a pre-existing issue.

mkannwischer: Let's do this as a follow-up. I opened https://github.com/pq-code-package/mldsa-native/issues/766 for this.

mldsa/src/poly.c
mkannwischer: Yes, we can. We can follow the same approach as in mlkem-native and just eliminate ntt.c altogether. 
That is a larger change than I would like to have in this PR though. I opened https://github.com/pq-code-package/mldsa-native/issues/767.

mldsa/src/poly.c
mkannwischer: Yes, let's do it in a follow-up. https://github.com/pq-code-package/mldsa-native/issues/767.

mldsa/src/poly.c
mkannwischer: Same comment applies to mlkem-native though. Do you suggest to change it there as well?

mldsa/src/poly.c
mkannwischer: I don't follow. You say we are missing a post-condition assertion here. So are you suggesting to add a `  mld_assert_abs_bound(c->coeffs, MLDSA_N, MLDSA_Q);` after ` mld_poly_pointwise_montgomery_c(c, a, b);`? I can't find any case in mlkem-native and mldsa-native where we have a post-condition assertion after the call to `foo_c()`. It seems redundant as the `foo_c()` function will have the post-condition assertion.

mldsa/src/poly.c
mkannwischer: I'm in favour of this change as it will speed up tests.
@willieyz, can you please change it in this PR and also in mlkem-native?

mldsa/src/poly.c
hanno-becker: I'm saying the assertions are inconsistent. We have a pre-condition assert at the toplevel, but no post-condition check. Indeed the pre-condition check could be moved into the native branch since it, too, is duplicated with the precondition assert in `foo_c`. If one wanted to be pedantic and consistent, though, and _always_ have pre/post asserts matching a function's contract, I'd be OK with duplicated post-conditions as well.

mkannwischer: Okay, let's move the pre-condition into the native branch as proposed in the other comment.

mldsa/src/poly.c
willieyz: Hello @mkannwischer and @hanno-becker, thank you for the review.
I have updated this PR according to your comments and moved all duplicate pre-bound assertions into the native branch.
In addition, I opened a PR addressing the same issue in mlkem-native: https://github.com/pq-code-package/mlkem-native/pull/1334.
Thank you!

mkannwischer (approved):
Thanks for the changes @willieyz. LGTM.

@hanno-becker, WDYT?

mldsa/src/poly.c
```diff
@@ -320,15 +368,12 @@ __contract__(
   ensures(array_bound(a, 0, return_value, 0, MLDSA_Q))
 )
 {
-  unsigned int ctr, pos;
-  uint32_t t;
-  mld_assert_bound(a, offset, 0, MLDSA_Q);
-
 /* TODO: CBMC proof based on mld_rej_uniform_native */
 #if defined(MLD_USE_NATIVE_REJ_UNIFORM)
+  int ret;
```
hanno-becker, line 373: Scope extension for `ret` wasn't necessary, but does not harm

mldsa/src/polyvec.c
```diff
@@ -323,54 +325,95 @@ void mld_polyvecl_pointwise_poly_montgomery(mld_polyvecl *r, const mld_poly *a,
...
+{
+#if defined(MLD_USE_NATIVE_POLYVECL_POINTWISE_ACC_MONTGOMERY_L4) && \
+    MLD_CONFIG_PARAMETER_SET == 44
+  /* TODO: proof */
+  int ret;
+  mld_assert_bound_2d(u->vec, MLDSA_L, MLDSA_N, 0, MLDSA_Q);
+  mld_assert_abs_bound_2d(v->vec, MLDSA_L, MLDSA_N, MLD_NTT_BOUND);
+  ret = mld_polyvecl_pointwise_acc_montgomery_l4_native(
+      w->coeffs, (const int32_t(*)[MLDSA_N])u->vec,
+      (const int32_t(*)[MLDSA_N])v->vec);
```
hanno-becker, line 384: This cast business is not very nice. I wonder if we should modify the native API to streamline this. Not an issue for this PR.

hanno-becker (approved):
LGTM, thank you @willieyz !

mldsa/src/polyvec.c
mkannwischer: mlkem-native has 
```
    ret = mlk_polyvec_basemul_acc_montgomery_cached_k2_native(
        r->coeffs, (const int16_t *)a, (const int16_t *)b,
        (const int16_t *)b_cache);
```
Is that better?

PR #746, opened 2025-11-26
Examples: multilevel build

willieyz:
- Resolves: #577 
- This PR add a multilevel build example demonstrating how to build mldsa-native with support for all three security levels while sharing level-independent code.

- This example builds the library three times in separate build directories(build/mldsa{44,65,87}). ML-DSA-44 is built with `MLD_CONFIG_MULTILEVEL_WITH_SHARED` to include all level-independent code, while ML-DSA-65 and ML-DSA-87 are built with `MLD_CONFIG_MULTILEVEL_NO_SHARED` to exclude shared code. All builds use the common namespace prefix 'mldsa' with automatic suffixes (44/65/87) for level-dependent functions.

This commit made following changes:
- Add `examples/multilevel_build/` with:
  - `Makefile`: Build system for all three security levels
  - `main.c`: Test program exercising all three ML-DSA variants
  - `mldsa_native_all.h`: Unified header exposing all three APIs
  - `expected_signatures.h`: For validation
  - `README.md`: Documentation
  - Symlinks to mldsa source and test RNG
- Add MULTILEVEL_BUILD test type in scripts/tests
- Add multilevel_build to tests script
- Add clean target for multilevel_build in root Makefile

- reference from mlkem-native's multilevel_build example.


examples/multilevel_build/expected_signatures.h
```diff
@@ -0,0 +1,886 @@
+/*
```
hanno-becker: Can we not symlink to examples/monolithic_build_multilevel_native/expected_signatures.h here?

willieyz: Hello, @hanno-becker , thank you for your review!
Sorry, I did not follow up here, maybe I miss something...

For the `multilevel_build` example, the `expected_signatures.h` I added is a regular file and not a symlink to `examples/monolithic_build_multilevel_native/expected_signatures.h`  :
```
MacBookPro:multilevel_build willieyz$ ls -la /Users/willieyz/mldsa-native/examples/multilevel_build/expected_signatures.h
-rw-r--r-- 1 willieyz staff 66312 Nov 27 14:19 /Users/willieyz/mldsa-native/examples/multilevel_build/expected_signatures.h
``` 

May I clarify what you meant?
Are you suggesting that in this PR we should remove the existing symlink at
examples/monolithic_build_multilevel_native/expected_signatures.h
and replace it with a regular file as well?

Or did I misunderstand your comment?
Sorry if this is a very basic question, thank you for your patience and your reply!

mkannwischer: I agree with @hanno-becker here. Please use a symlink.

PR #750, opened 2025-11-28
verify: Switch to constant-time memcmp

mkannwischer:
In the very end of verify one has to compare the input challenge to the re-computed challenge. If they are equal (and some previous checks on h and z passed), the signature is valid.
Currently, our constant-time tests do not declassify the message and we, hence, need to declassify in this final step.
Before thi commit, the declassification would happen on the recomputed challenge just before the memcmp.
Now that a constant-time memcmp was added in
https://github.com/pq-code-package/mldsa-native/pull/714, we might as well use that and declassify only the result on the memcmp which feels a bit more naturual and is easier to justify.


mldsa/src/sign.c
```diff
@@ -850,28 +850,21 @@ int crypto_sign_verify_internal(const uint8_t *sig, size_t siglen,
...
+   * comparison of the challenge hash c with the re-computed challenge hash c2.
+   * This only leaks if the signature is valid or not.
    */
-  MLD_CT_TESTING_DECLASSIFY(c2, sizeof(c2));
-  for (i = 0; i < MLDSA_CTILDEBYTES; ++i)
-  __loop__(
-    invariant(i <= MLDSA_CTILDEBYTES)
-  )
+  MLD_CT_TESTING_DECLASSIFY(&c_valid, sizeof(uint8_t));
+  if (c_valid != 0)
```
hanno-becker: You could keep this entirely CT by using `mld_ct_sel_int32` here.

hanno-becker (approved):

mldsa/src/sign.c
mkannwischer: I gave it a try - please take a look.

hanno-becker (dismissed):

PR #756, opened 2025-12-01
linting: add shellcheck for shell script lint

L-series:
Add linting for shell scripts in the scipts directory as well as any
*.sh file accross the project using shellcheck.

- Resolves https://github.com/pq-code-package/mldsa-native/issues/907
- Ports https://github.com/pq-code-package/mlkem-native/pull/1329


scripts/lint
```diff
@@ -106,6 +104,62 @@ gh_error_simple()
...
+
+run-shellcheck()
+{
+  if ! command -v shellcheck >/dev/null; then
+    gh_error_simple "Shellcheck missing" "shellcheck is not installed"
+    error "Lint shellcheck"
+    SUCCESS=false
+    gh_summary_failure "Lint shellcheck"
+    return 0
+  fi
```
hanno-becker, line 115: Is `shellcheck` guaranteed to be available in our nix environments?

L-series: It seems to be present by default on the runners, however I now explicitly added it to utils.nix

scripts/format
```diff
@@ -29,36 +27,36 @@ error()
...
 nixpkgs-fmt "$ROOT"
 
 info "Formatting shell scripts"
-if ! command -v shfmt 2>&1 >/dev/null; then
+if ! command -v shfmt >/dev/null 2>&1; then
   error "shfmt not found. Are you running in a nix shell? See BUILDING.md."
   exit 1
 fi
-shfmt -s -w -l -i 2 -ci -fn $(shfmt -f $(git grep -l '' :/))
+shfmt -s -w -l -i 2 -ci -fn "$(shfmt -f "$(git grep -l '' :/)")"
```
hanno-becker: I am not a shell expert, but this looks wrong? Do we really want to pass the list of files as a single quoted argument to `shfmt` here?

hanno-becker: This fails locally for me. How come the CI does not catch this?

scripts/lint
hanno-becker: Thank you! At least locally on my Mac I didn't have it, but the your changes to the nix shell work for me.

scripts/format
L-series: Hey Hanno, I just spent some time looking at this - it turns out that the command 
` shfmt -s -w -l -i 2 -ci -fn "$(shfmt -f "$(git grep -l '' :/)")" `
has the following odd behavior:

1. It fails loudly (in the sense that it was printing an incorrect list of files which needed to be formatted to the console) - I missed this when rerunning my lint script. 
2. In the context of `checkerr`, since we aren't calling the functions with the redirection of stderr to stdout, it receives an empty `$out$` as `shfmt` was writing those logs to stderr. 
3. Since `$out` was empty and the return value was 1, `success` would still remain true and it was appearing as if the linting had succeeded. 

In general it would probably be wise to have `checkerr` fail if the return value is anything other than 0. I can implement this change if you'd like. 

I've fixed the syntax and now CI should be passing.

scripts/format
```diff
@@ -29,36 +27,40 @@ error()
...
+if ! command -v shfmt >/dev/null 2>&1; then
   error "shfmt not found. Are you running in a nix shell? See BUILDING.md."
   exit 1
 fi
-shfmt -s -w -l -i 2 -ci -fn $(shfmt -f $(git grep -l '' :/))
+
+# Find all scripts in the repo
+ALL_FILES=$(git grep -l '' :/)
+SHELL_SCRIPTS=$(echo "$ALL_FILES" | xargs shfmt -f)
+shfmt -s -l -i 2 -ci -fn $SHELL_SCRIPTS
```
hanno-becker: Why have we lost `-w`?

hanno-becker: Will this work if scripts ever contain spaces in their file names?

scripts/lint
```diff
@@ -130,7 +149,7 @@ fi
 gh_group_end
 
 gh_group_start "Linting c files with clang-format"
-checkerr "Lint C" "$(clang-format $(git ls-files ":/*.c" ":/*.h") --Werror --dry-run 2>&1 | grep "error:" | cut -d ':' -f 1,2 | tr ':' ' ')"
+checkerr "Lint C" "$(clang-format "$(git ls-files ":/*.c" ":/*.h")" --Werror --dry-run 2>&1 | grep "error:" | cut -d ':' -f 1,2 | tr ':' ' ')"
```
hanno-becker: Is this suffering from the same quoting issue as shfmt that we discussed?

hanno-becker: @L-series Gentle ping :)

scripts/format
L-series: Added back the missing -w and fixed the command to use xargs as per the `lint` script. I've checked and this will not work for files which contain spaces in the filename, however the original implementation also suffers from this issue. The output of renaming `lint` to `lint new` and calling shfmt on it:

```
shfmt -s -w -l -i 2 -ci -fn $(shfmt -f $(git grep -l '' :/))
lstat scripts/lint: no such file or directory
lstat new: no such file or directory
```

 I'll spend some time tomorrow thinking on how to resolve this.

scripts/lint
L-series: Hello, sorry for the late reply - I was away on Holiday! This is indeed suffering from the same issue. I've pushed a fix which involves xargs, as per the case of shfmt!

scripts/format
L-series: We could use a loop of the sort to bypass this issue:

```
git grep -l '' :/ | while IFS= read -r file; do
  shfmt -f "$file"
done | while IFS= read -r shellfile; do
  shfmt -s -w -l -i 2 -ci -fn "$shellfile"
done
```

scripts/format
```diff
@@ -29,36 +27,39 @@ error()
...
   exit 1
 fi
 
 nixpkgs-fmt "$ROOT"
 
 info "Formatting shell scripts"
-if ! command -v shfmt 2>&1 >/dev/null; then
-  error "shfmt not found. Are you running in a nix shell? See CONTRIBUTING.md."
+if ! command -v shfmt >/dev/null 2>&1; then
+  error "shfmt not found. Are you running in a nix shell? See BUILDING.md."
```
hanno-becker: This looks like a rebase-slip? I think it should be CONTRIBUTING.md here and in the other instances

L-series: Oops, good catch. Fixed it. Pushing the patch to mlkem now as well

PR #762, opened 2025-12-01
Better explanation for Barrett division in decompose

### LINKED ISSUE #654: Come up with a better explanation for round(-half-up) vs round-half-down in decompose
Decompose() naturally needs round-half-down. It can be computed with rounding mulhi which always round half up, but our current explanation ([here](https://github.com/pq-code-package/mldsa-native/blob/45e64ca9d22a2121757b2044820f517539b2cf9c/dev/aarch64_clean/src/poly_decompose_32_asm.S#L12-L15) for example) doesn't justify this very clearly. See https://github.com/pq-code-package/mldsa-native/pull/411#discussion_r2371057773 and https://github.com/pq-code-package/mldsa-native/pull/629#discussion_r2508911467 for more context.

jammychiou1:
- Resolves #654

Along the way, we also fix some minor details in comments for AVX2 decompose:
- The `floor()` in `floor((f + 127) >> 7)` was somewhat unnecessary as the usual semantic for the right-shift operator `>>` has integer output anyway. Seeing as the right-shift operator is not used in other explanation comments, we decided to rewrite it as division by `2^7` for better consistency.
- The bound of `f1''` is correct but the proof was misleading. The new proof should be clearer.

### DIFF
--- a/.github/workflows/all.yml
+++ b/.github/workflows/all.yml
@@ -83,3 +83,11 @@ jobs:
     needs: [ base ]
     uses: ./.github/workflows/baremetal.yml
     secrets: inherit
+  isabelle:
+    name: Isabelle
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    needs: [ base ]
+    uses: ./.github/workflows/isabelle.yml
+    secrets: inherit
--- a/dev/x86_64/src/poly_decompose_32_avx2.c
+++ b/dev/x86_64/src/poly_decompose_32_avx2.c
@@ -75,7 +75,7 @@ void mld_poly_decompose_32_avx2(int32_t *a1, int32_t *a0)
     /*
      * Compute f1 = round-(f1' / B) ≈ round(f1' * 1025 / 2^22). This is exact
      * for 0 <= f1' < 2^16. See mld_decompose() in mldsa/src/rounding.h for the
-     * proof.
+     * proof, and proofs/isabelle/compress for a formalization of the argument.
      *
      * round(f1' * 1025 / 2^22) is in turn computed in 2 steps as
      * round(floor(f1' * 1025 / 2^16) / 2^6). The mulhi computes f1'' =
--- a/dev/x86_64/src/poly_decompose_88_avx2.c
+++ b/dev/x86_64/src/poly_decompose_88_avx2.c
@@ -76,7 +76,7 @@ void mld_poly_decompose_88_avx2(int32_t *a1, int32_t *a0)
     /*
      * Compute f1 = round-(f1' / B) ≈ round(f1' * 11275 / 2^24). This is exact
      * for 0 <= f1' < 2^16. See mld_decompose() in mldsa/src/rounding.h for the
-     * proof.
+     * proof, and proofs/isabelle/compress for a formalization of the argument.
      *
      * round(f1' * 11275 / 2^24) is in turn computed in 2 steps as
      * round(floor(f1' * 11275 / 2^16) / 2^8). The mulhi computes f1'' =
--- a/proofs/isabelle/compress/Makefile
+++ b/proofs/isabelle/compress/Makefile
@@ -1,0 +1,16 @@
+# Copyright (c) The mldsa-native project authors
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+
+ISABELLE_VERSION ?= Isabelle2025
+ISABELLE_HOME ?= /Applications/$(ISABELLE_VERSION).app/bin
+
+all:
+	$(ISABELLE_HOME)/isabelle build -D .
+
+clean:
+	$(ISABELLE_HOME)/isabelle build -c -D .
+
+jedit:
+	$(ISABELLE_HOME)/isabelle jedit -D . ML-DSA_Compress.thy
+
+.PHONY: all clean jedit
--- a/proofs/isabelle/compress/README.md
+++ b/proofs/isabelle/compress/README.md
@@ -1,0 +1,44 @@
+[//]: # (SPDX-License-Identifier: CC-BY-4.0)
+
+# ML-DSA Compression Proofs
+
+Isabelle/HOL proofs for the Barrett division used in ML-DSA's `Decompose` routine.
+
+## Theory Files
+
+- [Rounding.thy](Rounding.thy) — Round-half-down rounding and its stability properties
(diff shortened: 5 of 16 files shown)


dev/x86_64/src/poly_decompose_32_avx2.c
```diff
@@ -72,10 +72,17 @@ void mld_poly_decompose_32_avx2(__m256i *a1, __m256i *a0, const __m256i *a)
...
      * Compute f1 = round-(f1' / B) ≈ round(f1' * 1025 / 2^22). This is exact
      * for 0 <= f1' < 2^16. Note that half is rounded down since 1025 / 2^22 ≲
-     * 1 / 4092.
+     * 1 / 4092, so (for example) f1' = B / 2 = 2046 is mapped to
+     *
+     *   round(2046 * 1025 / 2^22) = round(2046 * (1 / 4092 - epsilon))
+     *                             = round(1 / 2 - epsilon') = 0,
+     *
+     * where epsilon = 1 / 4092 - 1025 / 2^22 and epsilon' = 2046 * eps are both
+     * tiny but positive numbers.
```
jammychiou1: @hanno-becker May I have your review on this explanation? If this looks good, I plan to also apply this to the aarch64 implementation and resolve #654. Thank you!

dev/x86_64/src/poly_decompose_88_avx2.c
```diff
@@ -73,10 +73,17 @@ void mld_poly_decompose_88_avx2(__m256i *a1, __m256i *a0, const __m256i *a)
...
      * Compute f1 = round-(f1' / B) ≈ round(f1' * 11275 / 2^24). This is exact
      * for 0 <= f1' < 2^16. Note that half is rounded down since 11275 / 2^24 ≲
-     * 1 / 1488.
+     * 1 / 1488, so (for example) f1' = B / 2 = 744 is mapped to
+     *
+     *   round(744 * 11275 / 2^24) = round(744 * (1 / 1488 - epsilon))
+     *                             = round(1 / 2 - epsilon') = 0,
+     *
+     * where epsilon = 1 / 1488 - 11275 / 2^24 and epsilon' = 744 * eps are both
+     * tiny but positive numbers.
```
jammychiou1: Same here

dev/x86_64/src/poly_decompose_32_avx2.c
hanno-becker: @jammychiou1 Apologies for the slow reply.

I think if we want to fix #654 we should provide a general explanation.

Here's an attempt:
```
The approximation error for `f1' / B ≈ f1' * 1025 / 2^22` is `f1' * (1025/2^22 - 1/B)`. 
For `eps := 1025/2^22 - 1/B` we have `eps = 1/4290772992 ~ 2^{-31.99} < 2^{-31}`. 
Hence `|f1'| * eps < 2^{-15}`. Given `B = 4092 ~ 2^12`, we thus have `|f1'| * eps < 1/B`. 
On the other hand, `1/B` is the spacing between the integral multiples of `1/B`, which
includes all rounding boundaries `n + 1/2` (since `B` is even). Hence, if `f1' / B` is not 
of the form `n + 1/2`, then moving from `f1' / B` to `f1' * 1025 / 2^22` does not cross 
a rounding boundary, and hence `round(f1' / B) = round(f1' * 1025 / 2^22)`, and it doesn't
matter on either side which version of rounding one uses. 
If `f1' / B` _is_ of the form `n + 1/2`, then `f1' * 1025 / 2^22` is slightly below it 
(and _not_ a multiple of `n + 1/2`), hence `round-(f1' / B) = round(f1' * 1025 / 2^22)`; 
where the round-down on the LHS is essential, and on the RHS the type of rounding 
again does not matter.
```

jammychiou1: @hanno-becker I'm very sorry for the long delay. I wanted to work on this, but I had to prioritize other tasks in the last few weeks...

Thank you for the detailed suggestion. This explanation is very clear. I created a new patch that incorporates the explanation with some modifications: I reworded it slightly according to the context and added some details that are unobvious to me at first. Please take a look at your convenience. Thank you very much and happy holidays!

proofs/isabelle/compress/Makefile
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) The mldsa-native project authors
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+
+ISABELLE_VERSION ?= Isabelle2025
```
mkannwischer, line 4: The most recent version is Isabelle2025-2

.github/workflows/all.yml
```diff
@@ -83,3 +83,10 @@ jobs:
     needs: [ base ]
     uses: ./.github/workflows/baremetal.yml
     secrets: inherit
+  isabelle:
+    name: Isabelle
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    uses: ./.github/workflows/isabelle.yml
```
mkannwischer, line 92: add `needs: [ base  ]` here.

proofs/isabelle/compress/Makefile
hanno-becker: Yes, I just can't change to it yet (unrelated to this PR).

proofs/isabelle/compress/README.md
```diff
@@ -0,0 +1,24 @@
...
+- [Barrett_Division.thy](Barrett_Division.thy) — Barrett approximation and correctness theorem
+- [ML-DSA_Compress.thy](ML-DSA_Compress.thy) — ML-DSA specific instantiations for C/AVX2 and AArch64
+
+## Building
+
+```
+isabelle build -D .
+```
+
+Tested with Isabelle2025.1.
```
mkannwischer: Works fine with Isabelle2025-2, too.

.github/workflows/all.yml
hanno-becker: Done

proofs/isabelle/compress/Makefile
mkannwischer: Is there a standard way to make it use any installed Isabelle version? Or am I expected to set ISABELLE_VERSION in my env? If so, please add it to the README.

hanno-becker: Added

mkannwischer: thx!

PR #777, opened 2025-12-08
Backend unit tests

willieyz:
- Resolves: #470 

During this pull request, we also opened another PR for the mld_polyz_unpack backend unit test to address the issue we encountered while constructing the unit test, as detailed in #779.

- This PR add backend unit test for all function with `MLD_STATIC_TESTABLE` except FIPS202:

| Native | C backend |Unit test|
|:------:|:------:|:------:|
 | `mld_ntt_native`   | `mld_poly_ntt_c` |`test_native_ntt`|
 | `mld_intt_native`   | `mld_poly_invntt_tomont_c`   |`test_native_invntt_tomont`|
|`mld_poly_permute_bitrev_to_custom`|`((void)mat);`<br> `/* Nothing to do */ `|Skip(same reason as mlkem-native)|
|`mld_rej_uniform_native`|`mld_rej_uniform_c`|Skip(same reason as mlkem-native)|
|`mld_rej_uniform_eta2_native`<br> `mld_rej_uniform_eta4_native`|`mld_rej_eta_c`|Skip(same reason as mlkem-native)|
|`mld_poly_decompose_32_native`<br>`mld_poly_decompose_88_native`|`mld_poly_decompose_c`|`test_native_decompose`|
|`mld_poly_caddq_native`|`mld_poly_caddq_c`|`test_native_caddq`|
|`mld_poly_use_hint_32_native`<br>`mld_poly_use_hint_88_native`|`mld_poly_use_hint_c`|`test_native_use_hint`|
|`mld_poly_chknorm_native`|`mld_poly_chknorm_c`|`test_native_poly_chknorm`|
|`mld_polyz_unpack_17_native`<br>`mld_polyz_unpack_19_native`|`mld_polyz_unpack_c`|Fix overread problem in #784<br>`test_native_polyz_unpack`|
|`mld_poly_pointwise_montgomery_native`|`mld_poly_pointwise_montgomery_c`|`test_native_pointwise_montgomery`|
|`mld_polyvecl_pointwise_acc_montgomery_l4_native`<br>`mld_polyvecl_pointwise_acc_montgomery_l5_native`<br>`mld_polyvecl_pointwise_acc_montgomery_l7_native`|`mld_polyvecl_pointwise_acc_montgomery_c`|`test_native_polyvecl_pointwise_acc_montgomery`|

- Steps:
  - Add the test infrastructure to the Makefile and `scripts/tests`.
  - Update the build rules in `test/mk/` for component testing.
[...]

PR #784, opened 2025-12-10
Avoid overread in polyz_unpack (AArch64 + x86_64)

mkannwischer:
polyz_unpack_17_avx2/polyz_unpack_197_avx2 unpacks polynnomials with
coefficients packed into 18/20 bits each.
To do so, it currently loads 32 bytes into a 256-bit register and discards
the last 14/12-bytes. This is problematic in the last iteration as this
overreads the buffer presenting a potential safety problem.

This is a violoation of the API contract which only requires 18*32/20*32-bytes.
Hence, this commit eliminates the overread by adding a special handling for
the last iteration.

Luckily, this does not presenta problem in the current code base as all call
sites have sufficient padding (by chance):
1) In mld_poly_uniform_gamma1, the buffer is padded to full SHAKE256 blocks,
   i.e., for MLDSA_POLYZ_PACKEDBYTES=576 => ceil(576/136)*136=680, and
             MLDSA_POLYZ_PACKEDBYTES=640 => ceil(640/136)*136=680.
2) In mld_unpack_sig where the signature is unpacked, the z component is
   followed by the hint component which is 83 bytes.

This means both overreads are fine.

We did, however, find this problem in the unit tests that are being implemented
in https://github.com/pq-code-package/mldsa-native/pull/777.

The same problem is present in the AArch64 code - which is also fixed in this PR.

### DIFF
--- a/dev/aarch64_clean/src/polyz_unpack_17_asm.S
+++ b/dev/aarch64_clean/src/polyz_unpack_17_asm.S
@@ -65,7 +65,7 @@ MLD_ASM_FN_SYMBOL(polyz_unpack_17_asm)
 
 polyz_unpack_17_loop:
         ldr q1, [buf, #16]
-        ldr q2, [buf, #32]
+        ldr s2, [buf, #32]
         ldr q0, [buf], #36
 
         tbl v4.16b, {v0.16b}, idx0.16b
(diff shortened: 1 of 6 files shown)


hanno-becker:
@mkannwischer I updated the PR with a perhaps simpler approach to fixing the issue in AVX2. Please take a look. The AArch64 fix LGTM.

Issue #767, opened 2025-12-02
Inline ntt.c


mkannwischer:
https://github.com/pq-code-package/mldsa-native/pull/735 hoisted out the C portion of `poly_ntt` into `poly_ntt_c` resulting in now three levels of indirection `poly_ntt` -> `poly_ntt_c` -> `mld_ntt`. Similar for the `invntt`.
We should remove that last level of indirection and simply inline `mld_ntt` into `poly_ntt_c` (similar for `invntt`). Consequently, the ntt.c/h files will no longer be needed. This also aligns with mlkem-native.

PR #787, opened 2025-12-11
Inline ntt.c

willieyz:
- Resolves: #767 

- The current code has three levels of indirection for NTT operations:
`poly_ntt` → `poly_ntt_c` → `mld_ntt`, we inline ntt.c and align with the mlkem-native approach.

- This PR made follwoing changes:
  - Inline NTT implementation `mld_ntt` and `mld_invntt_tomont` directly into `poly_ntt_c` and `poly_invntt_tomont_c`
  - Move NTT helper functions (`mld_ntt_layer`, `mld_invntt_layer`, `mld_ntt_butterfly_block`, `mld_fqmul`, `mld_fqscale`) from `ntt.c` to `poly.c` as static functions
  - Remove `ntt.c` and `ntt.h` files entirely
  - Move NTT bound constants to `poly.h`
  - Remove CBMC contract for `mld_ntt` and `mld_invntt_tomont`
  - Update CBMC proofs's Makefile and harness by replace ntt.c/h to poly.c/h

- Test:
  - `tests all`
  - `OPT=0 CFLAGS=-std=c90 make quickcheck`


mldsa/src/poly.c
```diff
@@ -152,6 +152,128 @@ void mld_poly_shiftl(mld_poly *a)
...
+ *                10   --     14
+ *                   11   --     15
+ * - start=4, len=2
+ *   This would compute the following two butterflies
+ *          4 -- 6
+ *             5 -- 7
+ */
+
+/* Reference: Embedded in `ntt()` in the reference implementation @[REF]. */
+static MLD_INLINE void mld_ntt_butterfly_block(int32_t r[MLDSA_N],
```
mkannwischer, line 203: This `MLD_INLINE` was not there before and also isn't present in mlkem-native.
Why was it added? Let's benchmark.

mldsa/src/poly.c
```diff
@@ -152,6 +152,128 @@ void mld_poly_shiftl(mld_poly *a)
...
+ *
+ * Compute one layer of forward NTT
+ *
+ * Parameters:
+ * - r:     Pointer to base of polynomial
+ * - layer: Indicates which layer is being applied.
+ */
+
+/* Reference: Embedded in `ntt()` in the reference implementation @[REF]. */
+static MLD_INLINE void mld_ntt_layer(int32_t r[MLDSA_N], const unsigned layer)
```
mkannwischer, line 252: same

mldsa/src/poly.c
```diff
@@ -181,6 +317,69 @@ void mld_poly_ntt(mld_poly *a)
...
+  /* check-magic: 41978 == pow(2,64-8,MLDSA_Q) */
+  const int32_t f = 41978;
+  /* Bounds: MLD_INTT_BOUND is MLDSA_Q, so the bounds reasoning is just
+   * a special case of that in mld_fqmul(). */
+  return mld_montgomery_reduce((int64_t)a * f);
+}
+
+/* Reference: Embedded into `invntt_tomont()` in the reference implementation
+ * @[REF] */
+static MLD_INLINE void mld_invntt_layer(int32_t r[MLDSA_N], unsigned layer)
```
mkannwischer, line 346: same

mkannwischer (approved):
The performance improvements on c7a are absurd (and it is quite suprising that it also strongly affects the optimized builds where this code isn't actually used).

There is a CBMC proof performance regression for ML-DSA-86, but it isn't too bad and we can investigate that in a follow-up.

Issue #774, opened 2025-12-08
Port: Speed up `make` from mlkem-native


mkannwischer:
https://github.com/pq-code-package/mlkem-native/pull/1350

PR #788, opened 2025-12-11
Port: Speed up make

willieyz:
- Resolves: #774
- Rebased on: #777

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,17 +32,16 @@ endif
 
 W := $(EXEC_WRAPPER)
 
-# Detect available SHA256 command
-SHA256SUM := $(shell command -v shasum >/dev/null 2>&1 && echo "shasum -a 256" || (command -v sha256sum >/dev/null 2>&1 && echo "sha256sum" || echo ""))
-ifeq ($(SHA256SUM),)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -32,17 +32,25 @@ endif
...
 W := $(EXEC_WRAPPER)
 
+BUILD_DIR ?= test/build
+
+
 # Detect available SHA256 command
 SHA256SUM := $(shell command -v shasum >/dev/null 2>&1 && echo "shasum -a 256" || (command -v sha256sum >/dev/null 2>&1 && echo "sha256sum" || echo ""))
 ifeq ($(SHA256SUM),)
 $(error Neither 'shasum' nor 'sha256sum' found. Please install one of these tools.)
 endif
```
mkannwischer: While at it, can you move this block to the same place it is in mlkem-native?

willieyz: Hello @mkannwischer, thank you for your review.
I have moved this block to the same position as in mlkem-native, specifically between test: and run_kat_44:. I also double-checked the differences between the Makefiles in mldsa-native and mlkem-native and aligned them as closely as possible.

jakemas (approved):
Great work!

PR #792, opened 2025-12-15
sign mem: reuse w/w0 buffer

mkannwischer:
This is the first of a series of commit bringing down the memory consumption of signature_internal.
Memory consumption is reduced by placing buffers whose lifetime does not overlap into a union starting with thw w and w0 buffer. CBMC contracts and proofs are adjusted where necessary.

- Hoisted out from #791 

`-fstack-usage` of `signature_internal` of ML-DSA-86 reduces from 129616 to 121536 (aarch64, -Os, gcc14).

### DIFF
--- a/mldsa/src/polyvec.h
+++ b/mldsa/src/polyvec.h
@@ -484,7 +484,7 @@ void mld_polyveck_decompose(mld_polyveck *v1, mld_polyveck *v0,
 __contract__(
   requires(memory_no_alias(v1,  sizeof(mld_polyveck)))
   requires(memory_no_alias(v0, sizeof(mld_polyveck)))
-  requires(memory_no_alias(v, sizeof(mld_polyveck)))
+  requires(v0 == v || memory_no_alias(v, sizeof(mld_polyveck)))
   requires(forall(k0, 0, MLDSA_K,
     array_bound(v->vec[k0].coeffs, 0, MLDSA_N, 0, MLDSA_Q)))
   assigns(memory_slice(v1, sizeof(mld_polyveck)))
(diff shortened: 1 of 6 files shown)


jakemas:
Looks good to me! Confirmed CBMC passes locally with these changes.

jakemas (approved):

mldsa/src/polyvec.h
```diff
@@ -484,7 +484,7 @@ void mld_polyveck_decompose(mld_polyveck *v1, mld_polyveck *v0,
 __contract__(
   requires(memory_no_alias(v1,  sizeof(mld_polyveck)))
   requires(memory_no_alias(v0, sizeof(mld_polyveck)))
-  requires(memory_no_alias(v, sizeof(mld_polyveck)))
+  requires(v0 == v || memory_no_alias(v, sizeof(mld_polyveck)))
```
hanno-becker, line 487: Is it worth keeping the non-aliased version, or can we rewrite the code so it uses the aliased version everywhere?

hanno-becker (commented):
It looks like the CBMC runtime got quite a lot worse, from 7/7/11 to 11/15/21.

We can of course just hope for the best and merge this, but seeing that this is only the very beginning of a series of changes, I am wary that it's not going to work for long.

hanno-becker:
The culprit is `mld_attempt_signature_generation`, which now takes 1041s in CI for ML-DSA-87.

hanno-becker (changes requested):
Rather than ignoring the CBMC slowdown, I'd prefer if we could spend more time in this toy-example to understand what, if anything, we can do to keep the CBMC runtime largely where it was.

@tautschnig told me that if we never use `union`s for data-presentation changes and set `--slice-formula`, it _should_ not matter (Michael, chime in if I mispresent anything), but this example does not seem to confirm it:
```
CBMCFLAGS=--external-smt2-solver $(PROOF_ROOT)/lib/z3_no_bv_extract --z3
CBMCFLAGS += --slice-formula
CBMCFLAGS += --no-array-field-sensitivity
```

Issue #795, opened 2025-12-16
Namespace `STACK_SIZE` in various assembly files


mkannwischer:
`STACK_SIZE` should be `MLD_STACK_SIZE` to avoid clashing with symbols in consuming libraries.
It currentl occurs invarious places, e.g., in
https://github.com/pq-code-package/mldsa-native/blob/8ec5b2caa748ff3993c47471670568d703473a61/dev/aarch64_opt/src/rej_uniform_eta4_asm.S#L15

PR #796, opened 2025-12-16
Namespace STACK_SIZE

willieyz:
- Resolves: #795

- The STACK_SIZE should namespace as MLD_STACK_SIZE to avoid clashing with symbols in consuming libraries.


mkannwischer (changes requested):
While at it, could you also namespace the macros in `KeccakP_1600_times4_SIMD256.c`, please? That's actually more important than the assembly ones.

PR #798, opened 2025-12-17
`decompose`: Remove separate input argument

mkannwischer:
To reduce stack usage, polyveck_decompose should be able to operate on one of the outputs aliasing the input. This is already fully supported but wasn't used to ease CBMC proofs/contracts at a cost of requiring additional stack space which is undesirable.

This commit is an alternative to
https://github.com/pq-code-package/mldsa-native/pull/792 which extended the CBMC contracts to allow aliasing.

This commit instead, changes the signature of the decompose functions to take 2 arguments instead of 3, where the second output also acts as and input. The corresponding poly_ and native functions are adjusted accordingly. CBMC contracts/proofs and unit tests are adjusted to the new signature. The single callsite of polyveck_decompose is adjusted.

### DIFF
--- a/dev/aarch64_clean/src/poly_decompose_32_asm.S
+++ b/dev/aarch64_clean/src/poly_decompose_32_asm.S
@@ -40,8 +40,7 @@
 
         /* Parameters */
         a1_ptr          .req x0     // Output polynomial with coefficients c1
-        a0_ptr          .req x1     // Output polynomial with coefficients c0
-        a_ptr           .req x2     // Input polynomial
+        a0_ptr          .req x1     // Input/Output polynomial with coefficients c0
 
         count           .req x3
 
@@ -76,10 +75,10 @@ MLD_ASM_FN_SYMBOL(poly_decompose_32_asm)
         mov count, #(64/4)
 
 poly_decompose_32_loop:
-        ldr q1, [a_ptr, #1*16]
-        ldr q2, [a_ptr, #2*16]
-        ldr q3, [a_ptr, #3*16]
-        ldr q0, [a_ptr], #4*16
+        ldr q0, [a0_ptr, #0*16]
+        ldr q1, [a0_ptr, #1*16]
+        ldr q2, [a0_ptr, #2*16]
+        ldr q3, [a0_ptr, #3*16]
 
         decompose32 v5, v1, v24
         decompose32 v6, v2, v24
@@ -102,7 +101,6 @@ poly_decompose_32_loop:
 
         .unreq a1_ptr
         .unreq a0_ptr
-        .unreq a_ptr
         .unreq count
         .unreq q
         .unreq q_bound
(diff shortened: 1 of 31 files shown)


dev/aarch64_clean/src/poly_decompose_32_asm.S
```diff
@@ -76,10 +76,10 @@ MLD_ASM_FN_SYMBOL(poly_decompose_32_asm)
         mov count, #(64/4)
 
 poly_decompose_32_loop:
+        ldr q0, [a_ptr, #0*16]
         ldr q1, [a_ptr, #1*16]
         ldr q2, [a_ptr, #2*16]
         ldr q3, [a_ptr, #3*16]
-        ldr q0, [a_ptr], #4*16
```
hanno-becker: I would suggest to make the aliasing explicit here and use `a0` instead of `a`. Otherwise, it's very confusing not to see `a_ptr` incremented.

dev/aarch64_clean/src/poly_decompose_88_asm.S
```diff
@@ -74,10 +74,10 @@ MLD_ASM_FN_SYMBOL(poly_decompose_88_asm)
 
         mov count, #(64/4)
 poly_decompose_88_loop:
+        ldr q0, [a_ptr, #0*16]
         ldr q1, [a_ptr, #1*16]
         ldr q2, [a_ptr, #2*16]
         ldr q3, [a_ptr, #3*16]
-        ldr q0, [a_ptr], #4*16
```
hanno-becker: Same here

mldsa/src/poly_kl.h
```diff
@@ -21,18 +21,21 @@
  *              Assumes coefficients to be standard representatives.
  *
  * Arguments:   - mld_poly *a1: pointer to output polynomial with coefficients
- *c1
- *              - mld_poly *a0: pointer to output polynomial with coefficients
- *c0
- *              - const mld_poly *a: pointer to input polynomial
+ *                              c1
+ *              - mld_poly *a0: pointer to input/output polynomial. Output
+ *                              polynomial has coefficients a0
```
hanno-becker: c0?

hanno-becker: The use of a vs. c doesn't seem to be consistent throughout the documentation

hanno-becker (changes requested):
Some minor points on readability and documentation, otherwise LGTM

dev/aarch64_clean/src/poly_decompose_88_asm.S
mkannwischer: Fixed.

dev/aarch64_clean/src/poly_decompose_32_asm.S
mkannwischer: Fixed.

mldsa/src/poly_kl.h
mkannwischer: Fixed.

hanno-becker (approved):

PR #800, opened 2025-12-17
Sign stack usage: Re-use y/h buffer

mkannwischer:
This commit is the second commit working towards bringing down the memory consumption of signature_internal. It combines the y and h buffer as those lifetime does not overlap.

`-fstack-usage` of `signature_internal` of ML-DSA-86 reduces from 121536 to 114368 (aarch64, -Os, gcc14).


hanno-becker:
@mkannwischer To double-check: Can you try adding a helper function which takes the pointers to the helper structures (incl. `y` and `h`) as arguments, and simply asserting that `y` and `h` be the same? I wonder if that would lead to a different handling of read/writes than when `y` and `h` are visibly (to CBMC) members of a union.

mkannwischer:
We merged #818  instead.

PR #801, opened 2025-12-17
Make allocation of large structures/buffers configurable

hanno-becker:
This commit introduces the configuration option

MLD_CONFIG_CUSTOM_ALLOC_FREE

which allows users to provide custom macros

MLD_CUSTOM_ALLOC(v, T, N)
MLD_CUSTOM_FREE(v, T, N)

that should be used in place of the default stack-allocation to allocate/free large internal structures. Those macros are then wrapped into MLD_ALLOC/MLD_FREE and used in the source. Importantly, MLD_FREE adds zeroization ahead of calling MLD_CUSTOM_FREE, so the latter need not take care of this.

The macros are put to use in sign.c, but not yet further down the call-stack.

The option is documented as experimental and (hence) unstable so we have freedom to adjust this ahead of v2.

A custom configuration is added which implements the macros based on C11's aligned_alloc. We enable the leak sanitizer in the tests for this configuration to catch any missing calls to MLD_FREE.

Since allocation can fail, additional error paths are introduced to the functions using MLD_ALLOC. This uniformly follows the pattern of all pointers being initialized to NULL before allocation, and a cleanup section calling MLD_FREE on them before returning the result.

To distinguish between out-of-memory failure and other kinds of functional failures, we introduce named error codes MLD_ERR_FAIL and MLD_ERR_OUT_OF_MEMORY. This will likely need further refinement, but this can only be done in v2.


mldsa/src/common.h
```diff
@@ -155,6 +155,62 @@
...
+#endif
+
+#if !defined(MLD_CONFIG_CUSTOM_ALLOC_FREE)
+/* Default: stack allocation */
+
+#define MLD_ALLOC(v, T, N)      \
+  MLD_ALIGN T mld_alloc_##v[N]; \
+  T *v = mld_alloc_##v
+
+#include "ct.h"
```
mkannwischer, line 183: Should this also have TODO note that this is a circular include that we should eliminate? (ct.h includes common.h)

mldsa/src/sign.c
```diff
@@ -614,70 +669,79 @@ int crypto_sign_signature_internal(
...
+    invariant(forall(k2, 0, MLDSA_K, array_abs_bound(t0->vec[k2].coeffs, 0, MLDSA_N, MLD_NTT_BOUND)))
+    invariant(forall(k3, 0, MLDSA_L, array_abs_bound(s1->vec[k3].coeffs, 0, MLDSA_N, MLD_NTT_BOUND)))
+    invariant(forall(k4, 0, MLDSA_K, array_abs_bound(s2->vec[k4].coeffs, 0, MLDSA_N, MLD_NTT_BOUND)))
+    invariant(ret == MLD_ERR_FAIL)
   )
   {
-    int attempt_result;
     /* Reference: this code explicitly checks for exhaustion of nonce     */
-    /* values to provide predictable termination and results in that case */
+    /* values to provide predictable termination and rets in that case */
```
mkannwischer: rets?

mldsa/src/sign.c
```diff
@@ -782,31 +861,46 @@ int crypto_sign_verify_internal(const uint8_t *sig, size_t siglen,
 {
   unsigned int i;
   int res;
```
mkannwischer: Should we consistently use ret?

mldsa/src/sign.c
```diff
@@ -1221,20 +1342,24 @@ int crypto_sign_pk_from_sk(uint8_t pk[MLDSA_CRYPTO_PUBLICKEYBYTES],
     mld_zeroize(pk, MLDSA_CRYPTO_PUBLICKEYBYTES);
```
mkannwischer, line 1346: Should this zeroize also happen if ret != 0?

mldsa/src/sign.h
```diff
@@ -96,7 +96,10 @@
  *              - uint8_t sk[MLDSA_CRYPTO_SECRETKEYBYTES]:   output private key
  *              - const uint8_t seed[MLDSA_SEEDBYTES]: input random seed
  *
- * Returns 0 (success) or -1 (PCT failure)
+ * Returns:     - 0: Success
+ *              - MLD_ERR_FAIL: PCT failure
+ *              - MLD_ERR_OUT_OF_MEMORY: If MLD_CONFIG_CUSTOM_ALLOC_FREE is
+ *                  used and an allocation via MLD_CUSTOM_ALLOC returned NULL.
```
mkannwischer, line 101: This comment should be added to all functions - e.g., crypto_sign_keypair is missing it.

mldsa/src/sign.h
```diff
@@ -271,7 +279,7 @@ __contract__(
   assigns(memory_slice(sig, MLDSA_CRYPTO_BYTES))
   assigns(object_whole(siglen))
   ensures((return_value == 0 && *siglen == MLDSA_CRYPTO_BYTES) ||
-          (return_value == -1 && *siglen == 0))
+          ((return_value == MLD_ERR_FAIL || return_value == MLD_ERR_OUT_OF_MEMORY) && *siglen == 0))
```
mkannwischer, line 295: I wonder if it is worth to model the failure behaviour as a part of the contract. It makes the contract harder to read and ultimately, the calling code should not make any assumptions about *siglen in case of failures anyway. WDYT?

mldsa/mldsa_native.h
mkannwischer, line 1: `pk_from_sk` lacks the documentation of the failure cases.

mldsa/src/sign.c
hanno-becker: Ok

mldsa/src/sign.c
hanno-becker: Oops

mldsa/src/sign.c
hanno-becker: Yes

mldsa/src/common.h
hanno-becker: Yes

mldsa/src/sign.h
hanno-becker: Yes

mldsa/mldsa_native.h
hanno-becker: Yes

mldsa/src/sign.h
hanno-becker: Ok, I can move this into the implementation.

hanno-becker: Hmm... on the other hand, this is internal documentation anyway.

hanno-becker: I changed it to
```c
  ensures(return_value == 0 || return_value == MLD_ERR_FAIL ||
          return_value == MLD_ERR_OUT_OF_MEMORY)
  ensures(return_value == 0 ==> *siglen == MLDSA_CRYPTO_BYTES)
  ensures(return_value != 0 ==> *siglen == 0)
```

mkannwischer:
> Thanks @hanno-becker for porting this - this looks like a lot of work.
> 
> A couple of small nits - nothing serious.
> 
> The commit message talks about "The macros are put to use in kem.c and indcpa.c, but not yet further down the call-stack." - that should be changed.



> Please adjust the commit message - see above,

I adjusted it.

PR #805, opened 2025-12-19
x86_64 Backend: Remove `<immintrin.h>` dependency from `arith_native_x86_64.h`

willieyz:
- Related: #806 

- Purpose
  - While adding CBMC proofs for `mld_ntt_avx2`, we identified a need to remove the dependency on `intrinsics.h`.

- Details
  - This PR refactors the x86_64 AVX2 implementation to remove intrinsic types from public function interfaces while preserving full AVX2 functionality.

- Changes
  - Removed `#include <immintrin.h>` from:
    - `src/arith_native_x86_64.h`
    - `align.h`  
  - All AVX2-related function signatures declared in  `meta.h`, `arith_native_x86_64.h`, and `native/x86_64/src/*_avx2.c` have been updated to use standard C types (`int32_t`) instead of intrinsic types (`__m256i`).
  - Updated all AVX2 function parameters from `__m256i *` to `int32_t *`.
  - Updated corresponding function declarations in `arith_native_x86_64.h` to use standard C types.  
  - AVX2 functionality is preserved internally via explicit casting to `__m256i *` within AVX2 implementation files.
    - When changing parameters from `__m256i *a` to `int32_t *a`, array indexing was adjusted:
  - Accesses were updated from `&a[i]` to `&a[8 * i]`.
  - Removed the inclusion of `cbmc.h` from `mldsa/src/common.h`.
  - Removed `#include "common.h"` from `mldsa/src/params.h`.
  - Removed `aligh.h`.

- Testing
Tested on both aarch64 and x86_64:
```sh
autogen && format && lint && make clean && tests all

PR #806, opened 2025-12-19
CBMC: Prove x86_64 NTT adheres to native API contract in api.h

willieyz:
- Based on: #805

- This PR add CBMC proof for mld_ntt_avx2, during this PR, we found the dependency of `<immintrin.h>` will cause the failed result for CBMC. we open #805 to remove this dependency to solve this problem.

### DIFF
--- a/examples/basic_lowram/mldsa_native/mldsa_native_config.h
+++ b/examples/basic_lowram/mldsa_native/mldsa_native_config.h
@@ -437,6 +437,9 @@
  *              Set this option and define `mld_sys_check_capability` if you
  *              want to use a custom method to dispatch between implementations.
  *
+ *              Return value 1 indicates that a capability is supported.
+ *              Return value 0 indicates that a capability is not supported.
+ *
  *              If this option is not set, mldsa-native uses compile-time
  *              feature detection only to decide which backend to use.
  *


mkannwischer (approved):
Thanks @willieyz. This LGTM.


@hanno-becker, do you agree?

dev/x86_64/src/arith_native_x86_64.h
```diff
@@ -33,7 +33,18 @@
 extern const uint8_t mld_rej_uniform_table[256][8];
 
 #define mld_ntt_avx2 MLD_NAMESPACE(ntt_avx2)
-void mld_ntt_avx2(int32_t *r, const int32_t *mld_qdata);
+void mld_ntt_avx2(int32_t *r, const int32_t *qdata)
+/* This must be kept in sync with the HOL-Light specification
+ * in proofs/hol_light/x86_64/proofs/mldsa_ntt.ml */
```
hanno-becker, line 38: Please also leave a similar note in `proofs/hol_light/x86_64/proofs/mldsa_ntt.ml` (see the Arm proof files)

mldsa/src/sys.h
```diff
@@ -230,6 +230,9 @@ typedef enum
 #include "cbmc.h"
 
 static MLD_INLINE int mld_sys_check_capability(mld_sys_cap cap)
+__contract__(
+  ensures(return_value == 0 || return_value == 1)
```
hanno-becker, line 234: While at it, can you add this requirement to the (textual) documentation of `MLD_CONFIG_CUSTOM_CAPABILITY_FUNC`:
```
 *              Set this option and define `mld_sys_check_capability` if you
 *              want to use a custom method to dispatch between implementations.
 *
 *              Return value 1 indicates that a capability is supported.
 *              Return value 0 indicates that a capability is not supported.
 *
```

willieyz: Thank you for your review!
I had add above content at the textual documentation of `MLD_CONFIG_CUSTOM_CAPABILITY_FUNC ` in `mldsa_native_config.h`, also re-run `autogen` to update for all config files.

dev/x86_64/src/arith_native_x86_64.h
willieyz: Hello, @hanno-becker , thank you for your review!
I had add following at the start of `Subroutine correctness theorems` in `/proofs/hol_light/x86_64/proofs/mldsa_ntt.ml` (reference from mlkem-native existing hol-light proof)
```
(* NOTE: This must be kept in sync with the CBMC specification
 * in mldsa/src/native/x86_64/src/arith_native_x86_64.h *)
```

PR #807, opened 2025-12-19
Add missing secret key coefficient validation

sgmenda:
### Issues:

`crypto_sign_pk_from_sk` was not verifying that s1 and s2 polynomial coefficients were within the valid range [-η, η].

### Description of changes: 

Implements missing validation in `crypto_sign_pk_from_sk` to verify that s1 and s2 polynomial coefficients are within the valid range [-η, η].

Motivated by https://github.com/aws/aws-lc/pull/2874 where we discovered, via [a Wycheproof test vector](https://github.com/C2SP/wycheproof/blob/e3c37e9db0f85a762dfcef1642b046bd31090ca4/testvectors_v1/mldsa_44_sign_noseed_test.json#L626-L646), that AWS-LC's MLDSA implementation was accepting secret keys with out-of-bounds s1/s2 coefficients.

### Call-outs:

- Covers all three MLDSA parameter sets (44, 65, 87) and both s1 and s2 polynomial vectors  
- Tests both underflow and overflow conditions at key boundary positions (first/last coefficients and polynomials)

### Testing:

The new functional tests and the existing tests all pass.
```bash
make run_func
make run_unit
make run_kat
```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the ISC license.

### DIFF
--- a/test/src/test_mldsa.c
+++ b/test/src/test_mldsa.c
@@ -5,7 +5,13 @@
 
 #include <stddef.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
+
+#include "../../mldsa/src/fips202/fips202.h"
+#include "../../mldsa/src/packing.h"
+#include "../../mldsa/src/params.h"
+#include "../../mldsa/src/polyvec.h"
 #include "../../mldsa/src/sign.h"
 #include "../../mldsa/src/sys.h"
 #include "../notrandombytes/notrandombytes.h"
@@ -27,6 +33,23 @@
       return 1;                                               \
     }                                                         \
   } while (0)
+
+/* Enum to specify which vector to corrupt in tests */
+typedef enum
+{
+  CORRUPT_S1,
+  CORRUPT_S2
+} corrupt_vector_t;
+
+/* Struct to define a coefficient corruption test case */
+typedef struct
+{
+  corrupt_vector_t vector;  /* Which vector to corrupt (s1 or s2) */
+  unsigned int poly_idx;    /* Polynomial index within the vector */
+  unsigned int coeff_idx;   /* Coefficient index within the polynomial */
+  int32_t corruption_value; /* Value to set the coefficient to */
+  const char *description;  /* Description of the test case */
+} corruption_test_case_t;
 
 
 static int test_sign_core(uint8_t pk[MLDSA_CRYPTO_PUBLICKEYBYTES],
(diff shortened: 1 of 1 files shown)


sgmenda:
@hanno-becker yep, exactly. `crypto_sign_pk_from_sk` already does validation like recomputing the hash, but is missing this particular check. https://github.com/pq-code-package/mldsa-native/blob/b7e9a2688b0094bc7ddbef6052afac72720f3019/mldsa/mldsa_native.h#L672-L685

jakemas:
Thank you for this! In my mind, my first thought was to modify the function `mld_unpack_sk/mld_polyvecl_unpack_eta` to perform the bounds check within this function itself -- the consequences of said approach would effect much more of the implementation as a whole, of course. I'll explain by thinking below:

In my mind, the fundamental issue is this function: https://github.com/pq-code-package/mldsa-native/blob/main/mldsa/src/polyvec.h#L666-L685
```
 * Name:        mld_polyvecl_unpack_eta
 *
 * Description: Unpack polynomial vector with coefficients in
 *              [-MLDSA_ETA,MLDSA_ETA].
 *
 * Arguments:   - mld_polyvecl *p: pointer to output polynomial vector
 *              - const uint8_t *r: input byte array with
 *                                  bit-packed polynomial vector
MLD_INTERNAL_API
void mld_polyvecl_unpack_eta(
    mld_polyvecl *p, const uint8_t r[MLDSA_L * MLDSA_POLYETA_PACKEDBYTES])
__contract__(
  requires(memory_no_alias(r,  MLDSA_L * MLDSA_POLYETA_PACKEDBYTES))
  requires(memory_no_alias(p, sizeof(mld_polyvecl)))
  assigns(memory_slice(p, sizeof(mld_polyvecl)))
  ensures(forall(k1, 0, MLDSA_L,
    array_bound(p->vec[k1].coeffs, 0, MLDSA_N, MLD_POLYETA_UNPACK_LOWER_BOUND, MLDSA_ETA + 1)))
);

```

Should not be bound to `MLD_POLYETA_UNPACK_LOWER_BOUND`, but by `MLDSA_ETA` itself. We made this modification in https://github.com/pq-code-package/mldsa-native/commit/cf01977101f8930a1584df9574e2ac3f99a2fef3 see:

```
/*
 * polyeta_unpack produces coefficients in [-MLDSA_ETA,MLDSA_ETA] for
 * well-formed inputs (i.e., those produced by polyeta_pack).
 * However, when passed an arbitrary byte array, it may produce smaller values,
 * i.e, values in [MLD_POLYETA_UNPACK_LOWER_BOUND,MLDSA_ETA]
 * Even though this should never happen, we use use the bound for arbitrary
 * inputs in the CBMC proofs.
 */
```

Essentially, what I feel the crux of this PR is getting at, is changing the above decision to use the bounds for well-formed inputs, and fail on the "arbitrary input" bound. What are thoughts on this? The current approach solves the issue in a way I hadn't first considered, and I just want to know the pros/cons of this over allowing `polyeta_pack` to unpack outside of `MLDSA_ETA` range.

This all came to me when creating the [CBMC contract](https://github.com/pq-code-package/mldsa-native/blob/main/mldsa/src/sign.c#L208-L209) for the `pk_from_sk` function, I had to make the bounds for use `MLD_POLYETA_UNPACK_LOWER_BOUND` rather than `MLDSA_ETA`, and chased the history down.

mkannwischer:
Thanks @sgmenda! Is there any reason for not re-using the constant-time `chknorm` function that we already have lying around?
I implemented an alternative patch using `chknorm` in https://github.com/pq-code-package/mldsa-native/pull/841.

initsecret:
@mkannwischer nope, that's a better fix. I hadn't thought of that. Thanks for the suggestion.

mkannwischer:
> @initsecret General Q: The standard does not seem to designate a list of recommended validity checks, does it?

This paragraph on secret key validity that I can find: 
image

PR #810, opened 2025-12-21
Add test for failing dynamic allocation

hanno-becker:
* Based on #801 

```
Implements allocation failure testing by using a custom allocator that
can fail at specific points.

The test also checks that freeing happens in reverse order to
allocation, allowing simple bump allocator implementations.
```

### DIFF
--- a/scripts/tests
+++ b/scripts/tests
@@ -212,6 +212,7 @@ class TEST_TYPES(Enum):
     SIZE = 17
     BASIC_DETERMINISTIC = 18
     UNIT = 19
+    ALLOC = 20
 
     def is_benchmark(self):
         return self in [TEST_TYPES.BENCH, TEST_TYPES.BENCH_COMPONENTS]
@@ -286,6 +287,8 @@ class TEST_TYPES(Enum):
             return "Measurement Code Size"
         if self == TEST_TYPES.UNIT:
             return "Unit Test"
+        if self == TEST_TYPES.ALLOC:
+            return "Alloc Test"
 
     def make_dir(self):
         if self == TEST_TYPES.BRING_YOUR_OWN_FIPS202:
@@ -351,6 +354,8 @@ class TEST_TYPES(Enum):
             return "size"
         if self == TEST_TYPES.UNIT:
             return "unit"
+        if self == TEST_TYPES.ALLOC:
+            return "alloc"
 
     def make_run_target(self, scheme):
         t = self.make_target()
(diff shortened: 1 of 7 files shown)


test/test_alloc.c
```diff
@@ -0,0 +1,570 @@
...
+  alloc_stack[alloc_stack_top].addr = addr;
+  alloc_stack[alloc_stack_top].size = size;
+  alloc_stack[alloc_stack_top].file = file;
+  alloc_stack[alloc_stack_top].line = line;
+  alloc_stack[alloc_stack_top].var = var;
+  alloc_stack[alloc_stack_top].type = type;
+  alloc_stack_top++;
+}
+
+static void alloc_tracker_pop(void *addr, const char *file, int line,
```
mkannwischer: Same comment as in mlkem-native: Please also assert the size of the FREE.

test/test_alloc.c
```diff
@@ -0,0 +1,570 @@
...
+    exit(1);
+  }
+
+  if (print_debug_info == 1)
+  {
+    fprintf(stderr, "Alloc #%d: %s %s (%d bytes) at %s:%d\n", alloc_counter,
+            type, var, (int)sz, file, line);
+  }
+
+  alloc_tracker_push(p, sz, file, line, var, type);
```
mkannwischer, line 186: To ease maintainance, I'd suggest to align this closely to mlkem-native.

test/test_alloc.c
```diff
@@ -0,0 +1,570 @@
...
+  {
+    return 1;
+  }
+
+  if (test_pk_from_sk_alloc_failure() != 0)
+  {
+    return 1;
+  }
+
+  return 0;
```
mkannwischer, line 570: Did you intentionally skip the following functions, or should those be added?
- crypto_sign_signature_pre_hash_internal
- crypto_sign_verify_pre_hash_internal 
- mld_prepare_domain_separation_prefix

It's probably best to add them now to not risk forgetting them in case they do add alllocations in there later.

hanno-becker: I left those out intentionally because they require exercising with different parameters for full coverage, and that is already achieved by the way they are called from the higher level functions?

test/test_alloc.c
hanno-becker: Done

test/test_alloc.c
hanno-becker: done

scripts/tests
```diff
@@ -1379,6 +1417,13 @@ def cli():
         parents=[common_parser],
     )
 
+    # alloc arguments
+    alloc_parser = cmd_subparsers.add_parser(
+        "alloc",
+        help="Run the alloc tests for all parameter sets",
+        parents=[common_parser],
+    )
```
mkannwischer: To ease maintenance, please reduce the diff to the mlkem-native tests script.

hanno-becker: Done

PR #811, opened 2025-12-22
HOL-Light: Speed up NTT proof

### LINKED ISSUE #648: HOL-Light: Autogenerate constant tables for NTT
The twiddle (and other constant) tables used by the HOL-Light proof for the forward NTT should be auto-generated.
See https://github.com/pq-code-package/mlkem-native/blob/90fed62ae9361d0da7760817c229ec50fcbfa2e6/scripts/autogen#L925

mkannwischer:
This commit ports @jargh's improvements to the NTT proof in s2n-bignum
which should result in vast proof performance improvements (1.6x- 4.3x
according to the PR).

- Ports https://github.com/awslabs/s2n-bignum/pull/325

While at it, I also introduced auto-generation for the twiddle factors
- Resolves #648


scripts/autogen
```diff
@@ -1129,6 +1165,46 @@ def gen_avx2_fwd_ntt_zetas(mult=False):
...
+        yield from [pow(modulus, -1, 2**32)] * 8
+        # MLD_AVX2_DIV_QINV
+        yield from [
+            signed_mod(pow(modulus, -1, 2**32) * pow(2, 64 - 8, modulus), 2**32)
+        ] * 8
+        # MLD_AVX2_DIV_QINV
+        yield from ([pow(2, 64 - 8, modulus)] * 8)
+
+        yield from gen_avx2_fwd_ntt_zetas(mult=True)
+        yield from gen_avx2_fwd_ntt_zetas(mult=False)
```
hanno-becker, line 1191: This should be shared with the C code, i.e. we should auto-generate the entire qdata array in `consts.c` from a helper function in autogen, and use the same helper function to emit the array embedded in the HOL-Light proof. Otherwise, there's risk the two run out of sync.

hanno-becker: Deferred as agreed below.

PR #814, opened 2025-12-24
CBMC: Use instrumented malloc/free for MLD_ALLOC/MLD_FREE

hanno-becker:
* Port of: https://github.com/pq-code-package/mlkem-native/pull/1405

The default implementation of MLD_ALLOC and MLD_FREE uses stack
allocation, which the compiler can assume not to fail. This means
that CBMC does not exercise the cleanup paths which handle fallible
dynamic allocation -- a significant proof gap.

This commit changes the configuration uses for the CBMC proofs
to use the (instrumented) calls to malloc/free for MLD_ALLOC/MLD_FREE.

Importantly, we set --malloc-may-fail to model allocation as fallible,
and --malloc-fail-null to return NULL in case of allocation failure.


mkannwischer (commented):
Thanks @hanno-becker. The changes itself look good to me, but CBMC fails due to timeouts. Marking as draft until we can resolve that.

hanno-becker:
This would be good to have for v1. Marking v1 so at least we're forced to make a deliberate decision about it.

mkannwischer:
Quite a peformance hit, but probably tolerable at least temporarily. @hanno-becker, is this ready for review?

hanno-becker:
@mkannwischer I experimented some more with CBMC but could not get the proofs to go through faster than the current status. Maybe @rod-chapman has ideas when he's back from PTO.

I think the performance loss is acceptable for the increase in proof scope and recommend we move forward with this.

mkannwischer (approved):
Thanks @hanno-becker. LGTM

PR #815, opened 2025-12-25
sign stack usage: Re-use y/h buffer

mkannwischer:
This commit is the second commit working towards bringing down the memory consumption of signature_internal. It combines the y and h buffer as those lifetime does not overlap.

We work around https://github.com/diffblue/cbmc/issues/8813 by introducing a no-op mld_cbmc_workaround_8813 whose contract states that y and h point to the same piece of memory of correct size.

- Alternative to #800


mkannwischer:
We merged #818  instead.

Issue #785, opened 2025-12-10
Extend macro typo checker for wrongly ported MLK_XXX / MLKEM_XXX macros


hanno-becker:
We often port material from mlkem-native to mldsa-native, in which case it is easy to forget to rename `MLK_XXX` to `MLD_XXX` in comments.

We should extend `scripts/autogen` to not only check for macro typos in the `MLD_XXX` namespace, but also flag any `MLK_XXX` or `MLKEM_XXX` macros.

PR #816, opened 2025-12-26
Autogen: Extend macro typo checker for MLK_XXX/ MLKEM_XXX

willieyz:
- Resolves: #785
- This PR extend the macro typo checker in autogen, for checking wrongly port MLK_XXX/ MLKEM_XXX

### DIFF
--- a/mldsa/mldsa_native.S
+++ b/mldsa/mldsa_native.S
@@ -247,7 +247,7 @@
 #undef MLD_NAMESPACE_KL
 #undef MLD_NAMESPACE_PREFIX
 #undef MLD_NAMESPACE_PREFIX_KL
-#undef MLK_UNION_OR_STRUCT
+#undef MLD_UNION_OR_STRUCT
 #undef mld_memcpy
 #undef mld_memset
 /* mldsa/src/packing.h */
--- a/mldsa/mldsa_native.c
+++ b/mldsa/mldsa_native.c
@@ -243,7 +243,7 @@
 #undef MLD_NAMESPACE_KL
 #undef MLD_NAMESPACE_PREFIX
 #undef MLD_NAMESPACE_PREFIX_KL
-#undef MLK_UNION_OR_STRUCT
+#undef MLD_UNION_OR_STRUCT
 #undef mld_memcpy
 #undef mld_memset
 /* mldsa/src/packing.h */
--- a/mldsa/src/common.h
+++ b/mldsa/src/common.h
@@ -278,9 +278,9 @@
  * is resolved
  */
 #if defined(MLD_CONFIG_REDUCE_RAM)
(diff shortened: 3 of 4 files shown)


hanno-becker (changes requested):
Thank you @willieyz. As I understand, with the change we would still allow `MLK_FOO` if there is `#define MLK_FOO` somewhere. That's not intended. We should not have any `MLK_XXX` or `MLKEM_XXX` macros in the mldsa-native code base.

scripts/autogen
```diff
@@ -27,6 +27,11 @@ montgomery_factor = pow(2, 32, modulus)
 _RE_DEFINED = re.compile(r"defined\(([^)]+)\)")
 _RE_MARKDOWN_CITE = re.compile(r"\[\^(?P<id>\w+)\]")
 _RE_C_CITE = re.compile(r"@\[(?P<id>\w+)")
+_FORBIDDEN_PREFIXES = "MLK_|MLKEM_"
+_RE_MACRO_CHECK = re.compile(
```
hanno-becker: `_RE_MACRO_CHECK` is overwritten again on line 37.

I suggest you keep it simple and just add a `_RE_MLKEM_MACRO_CHECK` which like, `_RE_MACRO_CHECK`, just with `MLK/MLKEM`, and then you add a separate check for their presence rather than extending the existing one.

willieyz: Hello, @hanno-becker , thank you for your review, 
I had change the macro checking method according to your suggestion,
Now all checking logic related with MLK/MLKEM will be processing in separate part, and mange the macro in additional `_RE_MLKEM_MACRO_CHECK`, thank you for your help!

scripts/autogen
```diff
@@ -1805,6 +1807,16 @@ def check_macro_typos_in_file(filename, macro_check):
                 f"Likely typo {txt} in {filename}:{line_no}? Not a defined macro."
             )
 
+    # Separate check for wrongly ported MLK/MLKEM macros
```
hanno-becker, line 1829: Can you run this check first? Otherwise, if you wrongly port `#define MLK_FOO ...` but use `MLD_FOO` in the code, you'll get the error "Undefined macro MLD_FOO" from the first check, but it would be clearer to get the second error.

hanno-becker (changes requested):
See comment

scripts/autogen
willieyz: Hello @hanno-becker, thank you for your help.
You are correct. Thank you for pointing this issue out. I have adjusted the order based on your suggestion and verified it with selftests.

hanno-becker (approved):

scripts/autogen
hanno-becker: Thanks!

PR #818, opened 2025-12-29
sign stack usage: Re-use y/h buffer 

mkannwischer:
Alternative to 
 - #800 
 - #815

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,8 @@ make test
 
 We use the [C Bounded Model Checker (CBMC)](https://github.com/diffblue/cbmc) to prove absence of various classes of undefined behaviour in C, including out of bounds memory accesses and integer overflows. The proofs cover all C code in [mldsa/src/*](mldsa) and [mldsa/src/fips202/*](mldsa/src/fips202) involved in running mldsa-native with its C backend. See [proofs/cbmc](proofs/cbmc) for details.
 
+**Note:** The `MLD_CONFIG_REDUCE_RAM` configuration option is not currently covered by CBMC proofs.
+
 HOL-Light functional correctness proofs can be found in [proofs/hol_light/x86_64](proofs/hol_light/x86_64). So far, the following functions have been proven correct:
 
 - x86_64 NTT [ntt.S](mldsa/src/native/x86_64/src/ntt.S)


mldsa/mldsa_native_config.h
```diff
@@ -650,6 +650,25 @@
...
+ * Name:        MLD_CONFIG_REDUCE_RAM
+ *
+ * Description: Set this to reduce RAM usage.
+ *
+ *              At this point, the configuration option merely enables the use
+ *              of unions for the major allocations vastly reducing stack usage.
+ *              Presently, unions cannot be used by default due to performance
+ *              issues with the CBMC proofs.
+ *              Once https://github.com/diffblue/cbmc/issues/8813 is resolved,
+ *              they can be enabled by default.
```
hanno-becker: Nit: I don't think this detail belongs to the config. There are also other optimizations such as generating the matrix ad-hoc which are clearly trade-offs between performance and memory usage, and will likely always be behind an option.

Focus instead on the relevant metrics: How much RAM does this currently save for each mode/function?

I would also indicate that this is still experimental and work in progress and no stability guarantee is made for it.

mldsa/src/sign.c
```diff
@@ -490,20 +490,36 @@ __contract__(
   unsigned int n;
   uint32_t z_invalid, w0_invalid, h_invalid;
   int ret;
+  typedef
+  /* TODO: Remove the following workaround for
+   * https://github.com/diffblue/cbmc/issues/8813 */
+#if defined(MLD_CONFIG_REDUCE_RAM)
```
hanno-becker: Nit: Wrap this into a `#define`, like `MLK_UNION_OR_STRUCT`.

examples/README.md
```diff
@@ -11,6 +11,11 @@ See [basic](basic) for a basic example of how to build a single instance of mlds
 ## Basic_deterministic
 
 See [basic_deterministic](basic_deterministic) for a basic example of how to build a single instance of mldsa-native without `randombytes()` implementation. This allows users to build mldsa-native using only the deterministic API when randomized functions are not required.
+
+## Basic_lowram
+
+See [basic_lowram](basic_lowram) for a basic example of how to build a single instance of mldsa-native with reduced RAM usage (`MLD_CONFIG_REDUCE_RAM`). This is useful for embedded systems with tight RAM constraints.
```
hanno-becker, line 17: The example needs to be added to the toplevel `Makefile`

mldsa/mldsa_native_config.h
hanno-becker: See https://github.com/pq-code-package/mldsa-native/pull/743/ for a more metrics-focused approach to the documentation.

examples/README.md
mkannwischer: Fixed.

mldsa/src/sign.c
mkannwischer: Okay. I also added a bit of commentary to explain the rationale.

mldsa/mldsa_native_config.h
mkannwischer: Okay.

PR #742, opened 2025-11-25
Introduce mld_polymat_get_row() helper function

hanno-becker:
- Add mld_polymat_get_row() to retrieve matrix row pointer
- Update mld_polyvec_matrix_pointwise_montgomery() to use helper
- If MLD_CONFIG_REDUCE_RAM is set, generate matrix ad-hoc.

* Resolves https://github.com/pq-code-package/mldsa-native/issues/738
* Resolves https://github.com/pq-code-package/mldsa-native/issues/739

PR #821, opened 2025-12-30
sign stack usage: compute z incrementally 

mkannwischer:
This commit reduces the stack usage of signing by computing z = y + s1*cp
incrementally (one polynomial at a time) allowing to eliminate the polyvecl
z (at to cost of a single poly z).
De-facto this saves L-1 KB irrespective of MLD_CONFIG_REDUCE_RAM.

Practically, the same buffer was used early in the function too. Here we
instead introduce a new polyvecl buffer tmp, but that can be placed in a union
together with w1.
Unfortuantely, with the current struct workaround for
https://github.com/diffblue/cbmc/issues/8813, this results in an increase in
stack space by L KB.
This gets eliminated when MLD_CONFIG_REDUCE_RAM is set.

- Depends on #742
- Alternative to https://github.com/pq-code-package/mldsa-native/pull/825 (which moves computation of z to a separate function).


hanno-becker:
I prefer #825 hoisting the computation of `z` into a separate function.

PR #822, opened 2025-12-31
Constant-time: Make signature declassifications explicit in verification

mkannwischer:
Previously, the signature would be declassified in the end of signing (which is the most common threat model).
However, for rare use cases the signature may be secret in which case it is useful to know which part of the code has to be adapted.

This commit removes the declassification from the end of signing, explicitly marks the signature as secret in the constant time tests (to also classify the previously declassified challenge c), and then adds necessary declassifications in verifiation.

We may consider eliminating some of those declassifications later. In particular, the data dependencies in use_hint are not hard to eliminate (in fact, the native implementations are already constant time). The final comparison of the challenges can also easily be turned into a constant-time comparison.

### DIFF
--- a/mldsa/src/sign.c
+++ b/mldsa/src/sign.c
@@ -616,10 +616,6 @@ __contract__(
   }
 
   /* All is well - write signature */
-  /* Constant time: At this point it is clear that the signature is valid - it
-   * can, hence, be considered public. */
-  MLD_CT_TESTING_DECLASSIFY(h, sizeof(*h));
-  MLD_CT_TESTING_DECLASSIFY(z, sizeof(*z));
   mld_pack_sig(sig, challenge_bytes, z, h, n);
 
   ret = 0; /* success */
@@ -874,6 +870,7 @@ int crypto_sign_verify_internal(const uint8_t *sig, size_t siglen,
                                 const uint8_t pk[MLDSA_CRYPTO_PUBLICKEYBYTES],
                                 int externalmu)
 {
+  uint32_t z_invalid;
   unsigned int i;
   int ret;
   MLD_ALLOC(buf, uint8_t, (MLDSA_K * MLDSA_POLYW1_PACKEDBYTES));
@@ -908,12 +905,24 @@ int crypto_sign_verify_internal(const uint8_t *sig, size_t siglen,
   /* mld_unpack_sig and mld_polyvecl_chknorm signal failure through a
    * single non-zero error code that's not yet aligned with MLD_ERR_XXX.
    * Map it to MLD_ERR_FAIL explicitly. */
+
+  /* Constant-time: The signature is commonly considered public. However, we
+   * mark it as secret to make explicit which parts of the code take time
+   * depending on the signature.
+   *
+   * mld_unpack_sig uses conditional branches to unpack the hint vector h.
+   * The h-component of the signature, hence, needs to be declassified.
+   */
   if (mld_unpack_sig(c, z, h, sig))
   {
     ret = MLD_ERR_FAIL;
     goto cleanup;
   }
-  if (mld_polyvecl_chknorm(z, MLDSA_GAMMA1 - MLDSA_BETA))
+
+  z_invalid = mld_polyvecl_chknorm(z, MLDSA_GAMMA1 - MLDSA_BETA);
+  /* Constant-time: We only leak if z is invalid or not. */
+  MLD_CT_TESTING_DECLASSIFY(&z_invalid, sizeof(uint32_t));
+  if (z_invalid)
   {
     ret = MLD_ERR_FAIL;
     goto cleanup;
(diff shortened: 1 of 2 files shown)


mldsa/src/sign.c
```diff
@@ -909,12 +906,24 @@ int crypto_sign_verify_internal(const uint8_t *sig, size_t siglen,
...
    * single non-zero error code that's not yet aligned with MLD_ERR_XXX.
    * Map it to MLD_ERR_FAIL explicitly. */
+
+  /* Constant-time: The signature is commonly considered public. However, we
+   * mark it as secret to make explicit which parts of the code take time
+   * depending on the signature. mld_unpack_sig uses conditional branches to
+   * unpack the hint vector h. */
+  MLD_CT_TESTING_DECLASSIFY(
+      sig + MLDSA_CTILDEBYTES + MLDSA_L * MLDSA_POLYZ_PACKEDBYTES,
+      MLDSA_POLYVECH_PACKEDBYTES);
```
hanno-becker: I'm not sure a public API should change the classification over user-provided data -- this can break or unexpectedly alter CT tests in integrations. Can we instead document that `verify` is _not_ CT for the (relevant parts of) the signature, and leave all classification to the tests?

hanno-becker (changes requested):
Toplevel Q first before checking the details: I don't think the public API should change the classification of user-provided buffers, or am I overlooking something?

mldsa/src/sign.c
mkannwischer: The downside of your proposal is that it does not make it apparent which operations leak.
With the current approach, it is very clear which function leaks what (and it is checked in CI, so you cannot introduce additional leaks).

mkannwischer: But, yes, we can move the declassification of the h-component to the tests insetead.
I do want to keep testing with the remainder of the signature marked as secret though.

mldsa/src/sign.c
```diff
@@ -908,12 +905,24 @@ int crypto_sign_verify_internal(const uint8_t *sig, size_t siglen,
   /* mld_unpack_sig and mld_polyvecl_chknorm signal failure through a
    * single non-zero error code that's not yet aligned with MLD_ERR_XXX.
    * Map it to MLD_ERR_FAIL explicitly. */
+
+  /* Constant-time: The signature is commonly considered public. However, we
+   * mark it as secret to make explicit which parts of the code take time
+   * depending on the signature.
```
hanno-becker, line 911: This comment seems outdated?

mldsa/src/sign.c
```diff
@@ -908,12 +905,24 @@ int crypto_sign_verify_internal(const uint8_t *sig, size_t siglen,
...
+   *
+   * mld_unpack_sig uses conditional branches to unpack the hint vector h.
+   * The h-component of the signature, hence, needs to be declassified.
```
hanno-becker, line 914:
```suggestion
   * mld_unpack_sig uses conditional branches to unpack the hint vector h.
   * It is the caller's responsibility to declassify the h-component of the signature.
```

mldsa/src/sign.c
```diff
@@ -937,6 +946,8 @@ int crypto_sign_verify_internal(const uint8_t *sig, size_t siglen,
   }
 
   /* Matrix-vector multiplication; compute Az - c2^dt1 */
+  /* Constant-time: poly_challenge uses conditional branches depending on c. */
+  MLD_CT_TESTING_DECLASSIFY(c, MLDSA_CTILDEBYTES);
```
hanno-becker, line 950: In the current version, the implementation is inconsistent regarding what is declassified by the function vs. expected to be declassified by the call-site: The h-component is expected to be declassified by the call-site, but the `c`-component is declassified here.

hanno-becker (changes requested):
We're still lacking a consistent story for how (de)classification is handled.

We rely on the caller of `sign` to classify the signature. When we verify, we expect the caller to again have declassified some parts of it, while others are declassified ad-hoc. 

Zooming out a little, I think we should strive for the following:
* For every `const` _input_ to a top-level API, the classification status of the bytes is not changed by the implementation. This is natural if you view `const` as extending to the `(value, secret_status)` presentation of data in the CT model. It also helps to avoid surprises when integrating with CT tests of consuming libraries.
* For every _output_ to a top-level API, the API should say explicitly what its classification status will be. 
 
At the moment, for example, keygen outputs a private key buffer parts of which are classified, and parts of which are declassified. This is not documented at the API-level, and it's not even obvious from the implementation. We should either say which parts are (de)classified and enforce it at the end of signing, or simply say that _all_ of the private key buffer is secret, enforce it in the implementation, and leave it to the caller to alter that if they wish to.

For signing, it's similar: We should say explicitly whether the signature is classified as secret or public, enforce it in the implementation, and adjust the call-sites (for us, the tests) accordingly.

Note that extending `const` to the security-state does _not_ imply that the implementation cannot declassify derived data from secret `const` inputs. Here, I'm not yet sure what to do. We can either a) declassify the relevant derived data, or b) where possible, expect the caller to have declassified the input data accordingly, and document that assumption. The latter is not always an option: For example, signing consumes the secret sk buffer, but safely declassifies derived data encountered during signature generation. For, for verification, we _could_ document that the relevant parts of the signature must already have been declassified, and not do any ad-hoc declassification in the implementation.

PR #825, opened 2025-12-31
sign stack usage: compute z incrementally

mkannwischer:
This commit reduces the stack usage of signing by computing z = y + s1*cp
incrementally (one polynomial at a time) allowing to eliminate the polyvecl
z (at to cost of a single poly z).
The computation of z is moved into a separate function (compute_pack_z) to
vastly speed up the CBMC proofs.

De-facto this saves L-1 KB irrespective of MLD_CONFIG_REDUCE_RAM.

Practically, the same buffer was used early in the function too. Here we
instead introduce a new polyvecl buffer tmp, but that can be placed in a union
together with w1.
Unfortuantely, with the current struct workaround for
https://github.com/diffblue/cbmc/issues/8813, this results in an increase in
stack space by L KB.
This gets eliminated when MLD_CONFIG_REDUCE_RAM is set.

- Depends on #742
- Alternative to #821 (which does not move computation of z to a separate function)

### DIFF
--- a/proofs/cbmc/crypto_sign_verify_pre_hash_internal/Makefile
+++ b/proofs/cbmc/crypto_sign_verify_pre_hash_internal/Makefile
@@ -20,14 +20,7 @@ PROOF_SOURCES += $(PROOFDIR)/$(HARNESS_FILE).c
 PROJECT_SOURCES += $(SRCDIR)/mldsa/src/sign.c
 
 CHECK_FUNCTION_CONTRACTS=$(MLD_NAMESPACE)verify_pre_hash_internal
-USE_FUNCTION_CONTRACTS=$(MLD_NAMESPACE)verify_internal \
-                       $(MLD_NAMESPACE)unpack_pk \
-                       $(MLD_NAMESPACE)unpack_sig \
-                       $(MLD_NAMESPACE)polyvec_matrix_expand \
-                       $(MLD_NAMESPACE)polyvecl_invntt_tomont \
-                       $(MLD_NAMESPACE)polyveck_invntt_tomont \
-                       $(MLD_NAMESPACE)polyveck_sub \
-                       $(MLD_NAMESPACE)polyveck_use_hint
+USE_FUNCTION_CONTRACTS=$(MLD_NAMESPACE)verify_internal
 USE_FUNCTION_CONTRACTS+=mld_zeroize
 
 APPLY_LOOP_CONTRACTS=on


mldsa/src/packing.c
```diff
@@ -100,15 +100,15 @@ void mld_unpack_sk(uint8_t rho[MLDSA_SEEDBYTES], uint8_t tr[MLDSA_TRBYTES],
 
 MLD_INTERNAL_API
 void mld_pack_sig(uint8_t sig[MLDSA_CRYPTO_BYTES],
```
hanno-becker: Nit: Perhaps make it explicit what is being packed? `mld_pack_sig_c_h`?

mldsa/src/sign.c
```diff
@@ -426,6 +427,90 @@ __contract__(
...
+  requires(forall(k1, 0, MLDSA_L, array_abs_bound(s1->vec[k1].coeffs, 0, MLDSA_N, MLD_NTT_BOUND)))
+  assigns(memory_slice(sig, MLDSA_CRYPTO_BYTES))
+  ensures(return_value == 0 || return_value == MLD_ERR_FAIL ||
+          return_value == MLD_ERR_OUT_OF_MEMORY)
+)
+{
+  unsigned int i;
+  uint32_t z_invalid;
+  int ret;
+  MLD_ALLOC(z, mld_poly, 1);
```
hanno-becker: If we can, we should rather avoid allocation in lower level functions. Can we pass `z` (perhaps renamed to `tmp`) as an argument?

hanno-becker: Not blocking, just a general thought.

mldsa/src/sign.c
```diff
@@ -426,6 +427,90 @@ __contract__(
...
+    invariant(i <= MLDSA_L)  
+  )
+  {
+    mld_poly_pointwise_montgomery(z, cp, &s1->vec[i]);
+    mld_poly_invntt_tomont(z);
+    mld_poly_add(z, &y->vec[i]);
+    mld_poly_reduce(z);
+
+    z_invalid = mld_poly_chknorm(z, MLDSA_GAMMA1 - MLDSA_BETA);
+    /* Constant time: It is fine (and prohibitively expensive to avoid)
```
hanno-becker, line 491: Note that this change has increased the granularity of the leakage: We now leak which part of `z` was invalid, not just whether any entry in the entire vector was invalid. As per the comment, this is OK, but the comment might still need adjusting.

mldsa/src/sign.c
```diff
@@ -426,6 +427,90 @@ __contract__(
...
+     * See Section 5.5 of @[Round3_Spec]. */
+    MLD_CT_TESTING_DECLASSIFY(&z_invalid, sizeof(uint32_t));
+    if (z_invalid)
+    {
+      ret = MLD_ERR_FAIL; /* reject */
+      goto cleanup;
+    }
+    /* If z is valid, then its coefficients are bounded by  */
+    /* MLDSA_GAMMA1 - MLDSA_BETA. This will be needed below */
+    /* to prove the pre-condition of pack_sig_z()             */
```
hanno-becker: Nit: formatting

mldsa/src/sign.c
```diff
@@ -426,6 +427,90 @@ __contract__(
...
+    mld_assert_abs_bound(z, MLDSA_N, (MLDSA_GAMMA1 - MLDSA_BETA));
+    mld_pack_sig_z(sig, z, i);
```
hanno-becker, line 510: Can you add a comment (ideally, a reference) why it is OK to write `z` to the public buffer `sig` here?

mldsa/src/sign.c
```diff
@@ -542,29 +640,12 @@ __contract__(
...
-   * as the candidate signature will be discarded anyway.
-   * See Section 5.5 of @[Round3_Spec]. */
-  MLD_CT_TESTING_DECLASSIFY(&z_invalid, sizeof(uint32_t));
-  if (z_invalid)
+  ret = mld_compute_pack_z(sig, cp, s1, y);
+  if (ret)
   {
-    ret = MLD_ERR_FAIL; /* reject */
     goto cleanup;
   }
```
hanno-becker, line 649: Nit: Double empty line

mkannwischer: Fixed.

mldsa/src/sign.c
mkannwischer: Agreed! Moved one level up.

mldsa/src/packing.c
mkannwischer: Good idea. Fixed.

mldsa/src/sign.c

mldsa/src/sign.c
mkannwischer: Updated.

mldsa/src/sign.c
mkannwischer: Added a paragraph. I do not have a good reference right now. WIll search a bit more later.

PR #835, opened 2026-01-03
CT: Declassify sig in test instead of at the end of signing

mkannwischer:
Alternative to https://github.com/pq-code-package/mldsa-native/pull/822 that I hope to be less controversial.
Currently the constant time tests for verification rely on the signature being declassified at the end of verification. This is not ideal. This commit moves this declassification to the constant-time test instead.

As suggested in
https://github.com/pq-code-package/mldsa-native/pull/822#pullrequestreview-3622033348, there is more work left to clean up the story around declassifications. This PR is a first step towards cleaning up that story to unblock https://github.com/pq-code-package/mldsa-native/pull/825 and https://github.com/pq-code-package/mldsa-native/pull/821, but there is more work left.

### DIFF
--- a/mldsa/src/sign.c
+++ b/mldsa/src/sign.c
@@ -113,6 +113,10 @@ static int mld_check_pct(uint8_t const pk[MLDSA_CRYPTO_PUBLICKEYBYTES],
   }
 #endif /* MLD_CONFIG_KEYGEN_PCT_BREAKAGE_TEST */
 
+  /* Constant time: Declassify signature as it is considered public in
+   * verification */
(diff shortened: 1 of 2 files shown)


hanno-becker:
@mkannwischer I'm just trying to understand if the dependency can be avoided. If the intent of the current (`main`) code is to have the signature declassified, I think that should just be done explicitly at the end of signing. Would that not be cleaner and remove the dependency? We can still continue discussing if the signature should be declassified or not, but at least you can progress the RAM optimizations independently.

mkannwischer:
> Would that not be cleaner and remove the dependency? We can still continue discussing if the signature should be declassified or not, but at least you can progress the RAM optimizations independently.

Yes, I have already done that - see the last two sentences of my previous comment.

PR #837, opened 2026-01-03
Align mld_ct_memcmp with mlkem-native

hanno-becker:
The previous version of mld_ct_memp (added in #714) had a misplaced value barrier allowing the compiler to elide part of the loop designed to prevent an early abort. We have not seen a compiler placing an early abort without this prevention -- and as such, this is not (yet) a security issue -- but as-is the code makes no sense.

This commit aligns mld_ct_memcmp with mlk_ct_memcmp from mlkem-native and adds the necessary cmask helpers and their CBMC proofs. We also slightly strengthen the spec to include that the return value is either 0x00 or 0xFF. Previously, we had only documented that it's 0 for equal data, and _some_ non-zero value otherwise. This needs to be backported to mlkem-native.

Note the minor difference with mlkem-native that ct_cmask_nonzero_u8 is built atop ct_cmask_nonzero_u32, rather than ct_cmask_nonzero_u16 as in mlkem-native. This is more natural in the context of mldsa-native where most arithmetic is 32-bit.

### DIFF
--- a/mldsa/src/ct.h
+++ b/mldsa/src/ct.h
@@ -216,6 +216,37 @@ __contract__(
 }
 
 /*************************************************
+ * Name:        mld_ct_cmask_nonzero_u32
+ *
+ * Description: Return 0 if input is zero, and -1 otherwise.
+ *
+ * Arguments:   uint32_t x: Value to be converted into a mask
+ *
+ **************************************************/
+static MLD_INLINE uint32_t mld_ct_cmask_nonzero_u32(uint32_t x)
+__contract__(ensures(return_value == ((x == 0) ? 0 : 0xFFFFFFFF)))
+{
+  int64_t tmp = mld_value_barrier_i64(-((int64_t)x));
+  tmp >>= 32;
+  return mld_cast_int64_to_uint32(tmp);
+}
+
+/*************************************************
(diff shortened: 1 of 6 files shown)


mldsa/src/ct.h
```diff
@@ -258,41 +289,43 @@ __contract__(
...
-static MLD_INLINE uint8_t mld_ct_memcmp(const void *a, const void *b,
+static MLD_INLINE uint8_t mld_ct_memcmp(const uint8_t *a, const uint8_t *b,
                                         const size_t len)
 __contract__(
   requires(len <= UINT16_MAX)
   requires(memory_no_alias(a, len))
   requires(memory_no_alias(b, len))
-  ensures((return_value == 0) == forall(i, 0, len, (((const uint8_t *)a)[i] == ((const uint8_t *)b)[i])))
-)
+  ensures((return_value == 0) == forall(i, 0, len, (a[i] == b[i]))))
```
mkannwischer, line 307: Is it intentional that the CBMC spec does not state that the return value is `0` or `0xFF`?

mkannwischer: I look the liberty to add the post-condition - proofs work just fine.

PR #838, opened 2026-01-03
verify: Switch to constant-time memcmp

hanno-becker:
* Based on #837 
* Taken over from #750 

This has negligible impact on performance while improving readability and provability.

For now, we still declassify the result of the verification, but this may be further improved in the future.


mkannwischer (approved):
Thanks Hanno for pushing this over the line! LGTM.

PR #841, opened 2026-01-05
`pk_from_sk`: Add validation of s1 and s2

mkannwischer:
Altenative to https://github.com/pq-code-package/mldsa-native/pull/807 (re-using `chknorm` instead of re-implementing a norm check, omitting tests for now). 

This commit adds validation of the s1 and s2 components of the secret key to the pk_from_sk function. It checks if coefficients are within the valid bound [-MLDSA_ETA, MLDSA_ETA] by using the chknorm function that is already present in the code.

Documentation and CBMC proofs are adjusted accordingly.


mldsa/mldsa_native.h
```diff
@@ -735,8 +735,12 @@ size_t MLD_API_NAMESPACE(prepare_domain_separation_prefix)(
 /*************************************************
  * Name:        crypto_sign_pk_from_sk
  *
- * Description: Derives public key from secret key with validation.
- *              Checks that t0 and tr stored in sk match recomputed values.
+ * Description: Derives public key from secret key.
+ *              Validity of the the secret key is checked with the following
+ *              tests:
```
hanno-becker:
```suggestion
 * Description: Performs basic validity checks on secret key, and derives public key.
 *
 *              Referring to the decoding `sk=(rho, K, tr, s1, s2, t0)` of the secret key
 *              (cf. [@FIPS204, Algorithm 25 skDecode]), the following checks are performed:
 *           ...
```

NB Why are we using `s0/s1` instead of `s1/s2` as in the spec?

mkannwischer: My bad. It is s1 and s2 everywhere except for this comment. Updated.

hanno-becker: Thanks! Can you also add the ref to the standard?

mkannwischer: Added. Sorry, missed the suggested change.

mldsa/src/sign.c
```diff
@@ -1333,7 +1337,7 @@ int crypto_sign_pk_from_sk(uint8_t pk[MLDSA_CRYPTO_PUBLICKEYBYTES],
                        sizeof(mld_polyveck));
   cmp1 = mld_ct_memcmp((const uint8_t *)tr, (const uint8_t *)tr_computed,
                        MLDSA_TRBYTES);
-  cmp = mld_value_barrier_u8(cmp0 | cmp1);
+  cmp = mld_value_barrier_u8(cmp0 | cmp1 | cmp2 | cmp3);
```
hanno-becker: Nit: `cmp` is no longer a good name. Perhaps `check`?

hanno-becker (approved):
LGTM, only one nit regarding variable naming

mldsa/src/sign.c
mkannwischer: Yes, `check` is better. Updated.

PR #850, opened 2026-01-07
mldsa_native.h: Introduce `MLD_TOTAL_ALLOC` constants

mkannwischer:
For consumers implementing custom allocations (in cases where the stack is
not large enough), it may be important to know how much space has to be
reserved for allocations of large structures.

This commit adds per-operation allocation constants:
 - MLD_TOTAL_ALLOC_{44,65,87}_{KEYPAIR,SIGN,VERIFY} for each parameter set
   and operation, indicating peak memory consumption plus 64 bytes headroom.
 - MLD_MAX_TOTAL_ALLOC_{KEYPAIR,SIGN,VERIFY} for maximum across parameter
   sets per operation.
 - MLD_TOTAL_ALLOC_{44,65,87} for maximum across operations per parameter set.
 - MLD_MAX_TOTAL_ALLOC for maximum across all parameter sets and operations.

Constants are provided for both default and MLD_CONFIG_REDUCE_RAM modes.

The alloc_test is extended to check that the constants indeed match the peak
memory consumption of the corresponding functions.
    
- Resolves https://github.com/pq-code-package/mldsa-native/issues/851

### DIFF
--- a/.github/workflows/baremetal.yml
+++ b/.github/workflows/baremetal.yml
@@ -35,5 +35,5 @@ jobs:
           acvp: true
           examples: false
           stack: false
-          alloc: false
+          alloc: true
           rng_fail: false


mkannwischer:
@gilles-peskine-arm: Is this approximately what you have in mind?

mldsa/mldsa_native.h
```diff
@@ -800,4 +800,27 @@ int MLD_API_NAMESPACE(pk_from_sk)(
...
+
+/***************************** Memory Usage **********************************/
+
+/*
+ * By default mldsa-native performs all memory allocations on the stack.
+ * Alternatively, mldsa-native supports custom allocation of large structures
+ * through the `MLD_CONFIG_CUSTOM_ALLOC_FREE` configuration option.
+ * See mldsa_native_config.h for details.
+ *
+ * `MLD_MAX_ALLOC` indicates the maximum (accumlative) allocations rounded up
```
hanno-becker: s/accumlative/accumulative

hanno-becker: The name `MLD_MAX_ALLOC` is (to me) suggestive of the maximum size of individual allocation requests, not the total amount of memory consumed.

How about `MLD_MAX_TOTAL_ALLOC` or `MLD_MAX_MEMORY_USAGE`?

test/src/test_alloc.c
```diff
@@ -266,6 +266,13 @@ void custom_free(void *p, size_t sz, const char *file, int line,
         return 1;                                                             \
       }                                                                       \
     }                                                                         \
+    if (((bump_high_mark + 1023) / 1024) > (MLD_MAX_ALLOC / 1024))            \
+    {                                                                         \
+      fprintf(stderr,                                                         \
+              "ERROR: max allocation %u in %s exceeded MLD_MAX_ALLOC=%d\n",   \
+              (unsigned)bump_high_mark, test_name, MLD_MAX_ALLOC);            \
+      return 1;                                                               \
+    }                                                                         \
```
hanno-becker: Should we keep track of the global high mark and assert that it's the same as `MLD_MAX_ALLOC`? So we're forced to update `MLD_MAX_ALLOC` when we implement further memory reductions?

mldsa/mldsa_native.h
mkannwischer: fixed.

mkannwischer: memory usage suggests it includes all memory usage. We still have considerable buffers not covered by MLD_ALLOC. Lets go for `MLD_MAX_TOTAL_ALLOC`. Fixed.

test/src/test_alloc.c
mkannwischer: Good idea. I implemented it.
This allows us to provide this for each parameter set separately. 
I added `MLD_TOTAL_ALLOC_44`, `MLD_TOTAL_ALLOC_65`, `MLD_TOTAL_ALLOC_87`.

`MLD_TOTAL_ALLOC` maps to the currently selected parameter set.
`MLD_MAX_TOTAL_ALLOC` corresponds to the maximum of those, i.e., `MLD_TOTAL_ALLOC_87`

mldsa/mldsa_native.h
```diff
@@ -807,4 +807,42 @@ int MLD_API_NAMESPACE(pk_from_sk)(
...
+#define MLD_TOTAL_ALLOC_65 (52 * 1024)
+#define MLD_TOTAL_ALLOC_87 (70 * 1024)
+#endif
+/* check-magic: on */
+
+#define MLD_MAX_TOTAL_ALLOC MLD_TOTAL_ALLOC_87
+
+#define MLD_TOTAL_ALLOC__(LVL) MLD_TOTAL_ALLOC_##LVL
+#define MLD_TOTAL_ALLOC_(LVL) MLD_TOTAL_ALLOC__(LVL)
+#define MLD_TOTAL_ALLOC MLD_TOTAL_ALLOC_(MLD_CONFIG_API_PARAMETER_SET)
```
gilles-peskine-arm: Thanks, this is the kind of things I wanted.

But ideally I'd like more fine tuning: I'd like to know the size for each operation (generate, sign, verify). An application that only does one of these operations shouldn't need the memory of the costliest operation.

test/src/test_alloc.c
```diff
@@ -266,10 +267,22 @@ void custom_free(void *p, size_t sz, const char *file, int line,
         return 1;                                                             \
       }                                                                       \
     }                                                                         \
+    if (((bump_high_mark + 1023) / 1024) > (MLD_TOTAL_ALLOC / 1024))          \
```
gilles-peskine-arm: A threshold-based check seems both wasteful (1kB isn't negligible on platforms that care) and fragile (adding and removing one byte could cause the mark to go up and down if you're unlucky). How about a check with a fixed tolerance? (Or a percentage but that seems more complicated and not that different in practice since we know the order of magnitude.) Something like
```
bump_high_mark <= MLD_TOTAL_ALLOC && bump_high_mark >= MLD_TOTAL_ALLOC - 100
```

mldsa/mldsa_native.h
mkannwischer: Makes sense. I split it up by operation.

test/src/test_alloc.c
mkannwischer: Yes, that's better. I changed it to 64 bytes of headroom. 

My main remaining question is if we do need that tolerance or if we can rely on the structs having exactly the same size with all compilers.

Our CI is testing gcc 4.8, 4.9, 7, 11, 13, 14, 15; clang 18, 19, 20, 21, zig 0.12, 0.13, 0.14, 0.15 on aarch64 and x86_64 and they all match the exact values. 
Our baremetal CI (Arm Cortex-M55, `arm-none-eabi-gcc 14.2.Rel1`) also results in the same values.

Given that, I'm inclined to remove the extra 64 byte entirely unless someone can think of a reason why we would want it.

/cc @hanno-becker

mldsa/mldsa_native.h
```diff
@@ -807,4 +807,84 @@ int MLD_API_NAMESPACE(pk_from_sk)(
...
+
+/*
+ * By default mldsa-native performs all memory allocations on the stack.
+ * Alternatively, mldsa-native supports custom allocation of large structures
+ * through the `MLD_CONFIG_CUSTOM_ALLOC_FREE` configuration option.
+ * See mldsa_native_config.h for details.
+ *
+ * `MLD_TOTAL_ALLOC_{44,65,87}_{KEYPAIR,SIGN,VERIFY}` indicates the maximum
+ * (accumulative) allocation plus 64 bytes of headroom for each parameter
+ * set and operation.
```
hanno-becker: It should perhaps be mentioned that this is specific to the places where `MLD_ALLOC` is used, and that places remain where stack allocation is used unconditionally?

With the default stack allocation, `MLD_TOTAL_XXX` will be lower by a few K than the actual stack usage.

mldsa/mldsa_native.h
```diff
@@ -807,4 +807,84 @@ int MLD_API_NAMESPACE(pk_from_sk)(
...
+ *
+ * This may be used to implement allocations using a fixed-sized buffer and a
+ * simple allocator (e.g., bump allocator).
+ */
+/* check-magic: off */
+#if defined(MLD_API_LEGACY_CONFIG) || !defined(MLD_CONFIG_REDUCE_RAM)
```
hanno-becker, line 829: Kudos for getting this right first go

mldsa/mldsa_native.h
```diff
@@ -807,4 +807,84 @@ int MLD_API_NAMESPACE(pk_from_sk)(
...
+#define MLD_TOTAL_ALLOC_KEYPAIR \
+  MLD_TOTAL_ALLOC_KEYPAIR_(MLD_CONFIG_API_PARAMETER_SET)
+
+#define MLD_TOTAL_ALLOC_SIGN__(LVL) MLD_TOTAL_ALLOC_##LVL##_SIGN
+#define MLD_TOTAL_ALLOC_SIGN_(LVL) MLD_TOTAL_ALLOC_SIGN__(LVL)
+#define MLD_TOTAL_ALLOC_SIGN MLD_TOTAL_ALLOC_SIGN_(MLD_CONFIG_API_PARAMETER_SET)
+
+#define MLD_TOTAL_ALLOC_VERIFY__(LVL) MLD_TOTAL_ALLOC_##LVL##_VERIFY
+#define MLD_TOTAL_ALLOC_VERIFY_(LVL) MLD_TOTAL_ALLOC_VERIFY__(LVL)
+#define MLD_TOTAL_ALLOC_VERIFY \
```
hanno-becker: How does this not break in multilevel build situations where we include the header 3 times?

hanno-becker (changes requested):
Looks overall good, but I am confused about level-dependent constants in the header -- this should not work in a multilevel monobuild?

mldsa/mldsa_native.h
mkannwischer: It's evaluated lazily, so if you don't use it you do not get an error.

mldsa/mldsa_native.h
mkannwischer: Added some clarifying text, but I think we should eliminate most of this asap.

mkannwischer: Opened https://github.com/pq-code-package/mldsa-native/issues/867 to get those eliminated

mldsa/mldsa_native.h
hanno-becker: It's still incompatible with multiple inclusions of the header? I think we should stick to not having any level-dependent definitions in the header.

mkannwischer: okay, I replaced it by namespaced variants `MLD_TOTAL_ALLOC_{44,65,87}`

test/src/test_alloc.c
hanno-becker: In the current version of the PR, don't have a tolerance window, but we check that the high watermark is exactly 64 bytes below the documented limit. Why? Sorry if I missed the point.

hanno-becker (commented):
I don't understand the purpose of forcing the documented limit to be 64 bytes above the actual limit. @mkannwischer can you clarify?

test/src/test_alloc.c
mkannwischer: Yes, that was to test that it is the same on all platforms. 
I think we should remove the extra 64 bytes.

hanno-becker: Agree

mkannwischer: Removed.

hanno-becker (approved):

test/src/test_alloc.c
gilles-peskine-arm: > do need that tolerance or if we can rely on the structs having exactly the same size with all compilers.

That depends on the layout of the structs. In theory, C compilers are allowed to insert arbitrary amounts of padding. But in practice, there isn't a lot of difference. You might need to test on slightly more exotic platforms though, as x86_64, aarch64 and arm are relatively tame.

mldsa/mldsa_native.h
```diff
@@ -807,4 +807,77 @@ int MLD_API_NAMESPACE(pk_from_sk)(
...
+ * Note that some stack allocation remains even
+ * when using custom allocators, so these values are lower than total stack
+ * usage with the default stack-only allocation.
+ *
+ * These constants may be used to implement custom allocations using a
+ * fixed-sized buffer and a simple allocator (e.g., bump allocator).
+ */
+/* check-magic: off */
+#if defined(MLD_API_LEGACY_CONFIG) || !defined(MLD_CONFIG_REDUCE_RAM)
+#define MLD_TOTAL_ALLOC_44_KEYPAIR 56640
```
gilles-peskine-arm, line 830: How come the values don't depend on the SHAKE implementation? Are SHAKE contexts never allocated through `MLD_ALLOC`?

test/src/test_alloc.c
gilles-peskine-arm: I browsed the `MLD_ALLOC` calls struct definitions a little. It looks like all `MLD_ALLOC` calls are either on byte arrays or on structs that contain arrays of `mld_poly` and possibly byte arrays (in the case of `mld_polymat` when `MLD_CONFIG_REDUCE_RAM` is enabled). `mld_poly` contains an array of `int32_t` with an attribute `MLD_ALIGN`. Depending on the compiler and options, in practice, this may mean a 4-byte or 8-byte alignment for `mld_poly`. This, in turn, could cause differences of 4 bytes for the size of some structures on some 32-bit platforms when `MLD_ALIGN` doesn't force a larger alignment.

mldsa/mldsa_native.h
mkannwischer: Keccak states are currently not allocated through MLD_ALLOC, and I'd prefer to keep it that way.
If you need to allocate it elsewhere, you can plug in a custom `fips202.h` (via `MLD_CONFIG_FIPS202_CUSTOM_HEADER`) where `mld_shake128/256ctx` is defined as a pointer of some sorts and actual allocation happens in the `_init` functions.

Probably we want to pull the context parameter introduced in #874 all the way through to the fips202 functions.
Let us know if someone needs that. 

/cc @davidchisnall

hanno-becker:
> ...and actual allocation happens in the _init functions.

The potential failure of those functions cannot currently be accommodated, though.

mldsa/mldsa_native.h
davidchisnall: It would be nice if the fips202 code could be factored out so that the current internal version is exposed as an implementation of the internal one.  Currently, the ML-KEM and ML-DSA implementations both expose one, but it's not the same.

mkannwischer: Consolidating those is on our roadmap: https://github.com/pq-code-package/mldsa-native/issues/604. I am not sure when we will get to that.
None of our consumers is using the internal implementation, so there is not much need for now.

Issue #783, opened 2025-12-10
Consolidate `MLD_CONFIG_CUSTOM_ZEROIZE` with mlkem-native


mkannwischer:
The `MLD _CONFIG_CUSTOM_ZEROIZE` works slightly different from how `MLK_CONFIG_CUSTOM_ZEROIZE` works:

https://github.com/pq-code-package/mldsa-native/blob/734e659ed3a8ec073ba4449c6c62c3d9531595b3/mldsa/src/ct.h#L314-L320

https://github.com/pq-code-package/mlkem-native/blob/b9bab5b74a612db84ee370ce1cffc5a3948b55e9/mlkem/src/verify.h#L434-L459

We should consolidate that.

PR #852, opened 2026-01-07
Consolidate MLD_CONFIG_CUSTOM_ZEROIZE with mlkem-native

willieyz:
- Resolves: #783

The MLD_CONFIG_CUSTOM_ZEROIZE macro in mldsa/src/ct.h behaves slightly differently from MLK_CONFIG_CUSTOM_ZEROIZE in mlkem/src/verify.h.

This commit aligns their behavior.

### DIFF
--- a/test/configs/custom_zeroize_config.h
+++ b/test/configs/custom_zeroize_config.h
@@ -363,13 +363,13 @@
  *              zeroizes intermediate stack buffers before returning from
  *              function calls.
  *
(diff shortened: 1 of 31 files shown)


test/configs/custom_zeroize_config.h
mkannwischer: The documentation also needs to be updated - it still says `mld_zeroize_native`

mkannwischer (approved):
Thanks @willieyz. LGTM.

PR #854, opened 2026-01-07
WIP: Reduce memory footprint of key generation

mkannwischer:
This PR works towards reducing memory consumption of key generation.


mkannwischer:
I'm running out of ideas for this one.
@rod-chapman, do you happen to have some spare cycles to take a look at it, please?

PR #861, opened 2026-01-07
feat(baremetal): Add m33-an524 platform support for bare-metal building and  testing

jricardini:
- Add comprehensive bare-metal testing platform for Cortex-M33 on MPS2-AN521
- Include platform-specific linker script (m33-an521.ld) for memory layout
- Add startup code (startup.c) for ARM Cortex-M33 initialization
- Implement functional test suite (comprehensive_func_test.c) for all ML-DSA parameter sets
- Implement performance benchmark suite (bench_baseline.c) with cycle counting
- Add Python execution wrapper (exec_wrapper.py) for QEMU integration
- Include platform makefile (platform.mk) for cross-compilation configuration
- Add comprehensive documentation (CROSS-BUILDING-AND-TESTING.md) with setup and troubleshooting
- Support testing all three security levels (ML-DSA-44, ML-DSA-65, ML-DSA-87)
- Enable automated testing via QEMU with semihosting support

### DIFF
--- a/nix/m33-an524-cmsis/default.nix
+++ b/nix/m33-an524-cmsis/default.nix
@@ -1,0 +1,97 @@
+# Copyright (c) The mldsa-native project authors
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+
+{ stdenvNoCC
+, fetchFromGitHub
+}:
+
+let
+  # CMSIS_5 repository from ARM
+  # Commit: 55b19837f5703e418ca37894d5745b1dc05e4c91 (2024-09-03)
+  # This commit corresponds to CMSIS v5.3.0 on the develop branch
+  # 
+  # Rationale for using fetchFromGitHub:
+  # - Fetches entire repository once instead of 12 individual files
+  # - More resilient to upstream URL changes
+  # - Easier to add new files from the same repository
+  # - Only requires maintaining one SHA256 hash
+  #
+  # To update to a newer CMSIS_5 version:
+  # 1. Find the desired commit hash from https://github.com/ARM-software/CMSIS_5
+  # 2. Run: nix-prefetch-github ARM-software CMSIS_5 --rev <commit-hash>
+  # 3. Update the rev and sha256 below
+  # 4. Verify all file paths still exist in the new commit
+  # 5. Test that patches still apply successfully
+  cmsis5_src = fetchFromGitHub {
+    owner = "ARM-software";
+    repo = "CMSIS_5";
+    rev = "55b19837f5703e418ca37894d5745b1dc05e4c91";
+    sha256 = "07hdcpd90r7g1wy1km2k69q2ny43bksdy8d3hpb32y1fgl83xra5";
+  };
+
+in
+stdenvNoCC.mkDerivation {
+  pname = "m33-an524-cmsis";
+  version = "2026-01-29";
+
+  dontUnpack = true;
+
+  buildPhase = ''
+    runHook preBuild
+    
+    # Create directory structure matching patch expectations
+    mkdir -p envs/m33-an524/src/platform/m-profile
+
+    # Copy CMSIS_5 files from repository to match patch paths
+    # All files are sourced from the CMSIS_5 repository at commit 55b19837
+    cp ${cmsis5_src}/CMSIS/Core/Include/core_cm33.h envs/m33-an524/src/platform/core_cm33.h
+    cp ${cmsis5_src}/CMSIS/Core/Include/cmsis_compiler.h envs/m33-an524/src/platform/cmsis_compiler.h
+    cp ${cmsis5_src}/CMSIS/Core/Include/cmsis_gcc.h envs/m33-an524/src/platform/cmsis_gcc.h
+    cp ${cmsis5_src}/CMSIS/Core/Include/cmsis_version.h envs/m33-an524/src/platform/cmsis_version.h
+    cp ${cmsis5_src}/Device/ARM/ARMCM33/Include/ARMCM33.h envs/m33-an524/src/platform/ARMCM33.h
+    cp ${cmsis5_src}/Device/ARM/ARMCM33/Source/startup_ARMCM33.c envs/m33-an524/src/platform/startup_ARMCM33.c
+    cp ${cmsis5_src}/Device/ARM/ARMCM33/Source/system_ARMCM33.c envs/m33-an524/src/platform/system_ARMCM33.c
+    cp ${cmsis5_src}/Device/ARM/ARMCM33/Include/system_ARMCM33.h envs/m33-an524/src/platform/system_ARMCM33.h
+    cp ${cmsis5_src}/Device/ARM/ARMCM33/Source/GCC/gcc_arm.ld envs/m33-an524/src/platform/m33-an524.ld
+    cp ${cmsis5_src}/CMSIS/Core/Include/mpu_armv8.h envs/m33-an524/src/platform/m-profile/armv8m_mpu.h
+    cp ${cmsis5_src}/CMSIS/Core/Include/cachel1_armv7.h envs/m33-an524/src/platform/m-profile/armv7m_cachel1.h
+
+    # Make files writable for patching
+    chmod -R u+w envs/
+
+    # Copy patches from repository
+    cp -r ${../../test/baremetal/platform/m33-an524/patches} patches
+    chmod -R u+w patches
+
+    # Apply patches to customize files for mldsa-native
+    patch -p0 < patches/cmsis5/core_cm33.h.patch || true
+    patch -p0 < patches/cmsis5/ARMCM33.patch || true
+    patch -p0 < patches/cmsis5/startup_ARMCM33.patch || true
+    patch -p0 < patches/cmsis5/system_ARMCM33.patch || true
+    patch -p0 < patches/cmsis5/m33-an524.ld.patch || true
+    
+    runHook postBuild
+  '';
+
+  installPhase = ''
+    mkdir -p $out/CMSIS/m-profile
+    
+    # Copy patched files to output in CMSIS directory structure
+    cp envs/m33-an524/src/platform/core_cm33.h $out/CMSIS/
+    cp envs/m33-an524/src/platform/cmsis_compiler.h $out/CMSIS/
+    cp envs/m33-an524/src/platform/cmsis_gcc.h $out/CMSIS/
+    cp envs/m33-an524/src/platform/cmsis_version.h $out/CMSIS/
+    cp envs/m33-an524/src/platform/ARMCM33.h $out/CMSIS/
+    cp envs/m33-an524/src/platform/startup_ARMCM33.c $out/CMSIS/
+    cp envs/m33-an524/src/platform/system_ARMCM33.c $out/CMSIS/
+    cp envs/m33-an524/src/platform/system_ARMCM33.h $out/CMSIS/
+    cp envs/m33-an524/src/platform/m33-an524.ld $out/CMSIS/
+    cp envs/m33-an524/src/platform/m-profile/armv8m_mpu.h $out/CMSIS/m-profile/
+    cp envs/m33-an524/src/platform/m-profile/armv7m_cachel1.h $out/CMSIS/m-profile/
+  '';
+
+  meta = {
+    description = "Patched CMSIS files for Cortex-M33 (AN524)";
+    homepage = "https://github.com/ARM-software/CMSIS_5";
+  };
+}
--- a/test/baremetal/platform/m33-an524/README.md
+++ b/test/baremetal/platform/m33-an524/README.md
@@ -1,0 +1,38 @@
+
+[//]: # (SPDX-License-Identifier: CC-BY-4.0)
+
+# Cortex-M33 MPS3-AN524 Platform
+
+CMSIS platform files for ARM Cortex-M33 on MPS3-AN524 FPGA (SSE-200 subsystem). Runs on QEMU `qemu-system-arm -M mps3-an524` and real hardware.
+
+**Target:** ARM Cortex-M33 r0p4 with FPU, DSP, TrustZone, caches enabled  
+**Memory:** 512KB Flash @ 0x10000000, 128KB SRAM @ 0x20000000  
+
+## Build
+
+```bash
+# Nix shell (provides toolchain and QEMU)
+nix develop --experimental-features 'nix-command flakes' .#arm-embedded
+
+# Build library
+make lib EXTRA_MAKEFILE=test/baremetal/platform/m33-an524/platform.mk OPT=0
+
+# Run functional tests
+make run_func_44 EXTRA_MAKEFILE=test/baremetal/platform/m33-an524/platform.mk OPT=0
+
+# Run benchmarks
+make run_bench_44 EXTRA_MAKEFILE=test/baremetal/platform/m33-an524/platform.mk OPT=0 CYCLES=NO
+```
+
+## Patches
+
+Adapted from ARM CMSIS_5 and pqmx. Key changes:
+
+- **ARMCM33.h**: Generic Cortex-M33 -> AN524 board config with 95 interrupts (UART, GPIO, SPI, Ethernet, I2S), enabled FPU/DSP/SAU/caches
+- **m33-an524.ld**: Memory map for AN524 (Flash 0x10000000, SRAM 0x20000000), 96KB stack, single PHDRS load segment, `.ARM.exidx` inside `.text` to prevent auto-generated ARM_EXIDX segment
+- **startup_ARMCM33.c**: Vector table with 95 AN524 interrupt handlers
+- **system_ARMCM33.c**: UART initialization at 32 MHz, `uart_putc()` for stdio output
+- **uart.c**: UART driver at 0x41303000 with 32 MHz clock
+- **core_cm33.h**: Corrected ARMv8-M MPU include path
+
+Per ARM DAI0524F Application Note.
--- a/test/baremetal/platform/m33-an524/patches/cmsis5/ARMCM33.patch
+++ b/test/baremetal/platform/m33-an524/patches/cmsis5/ARMCM33.patch
@@ -1,0 +1,131 @@
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+--- /tmp/ARMCM33.h.orig	2026-01-13 17:07:21.659694888 +0000
++++ envs/m33-an524/src/platform/ARMCM33.h	2026-01-13 15:11:08.445799847 +0000
+@@ -47,19 +47,91 @@ typedef enum IRQn
+   PendSV_IRQn                   =  -2,     /* 14 Pend SV Interrupt */
+   SysTick_IRQn                  =  -1,     /* 15 System Tick Interrupt */
+ 
+-/* -------------------  Processor Interrupt Numbers  ------------------------------ */
+-  Interrupt0_IRQn               =   0,
+-  Interrupt1_IRQn               =   1,
+-  Interrupt2_IRQn               =   2,
+-  Interrupt3_IRQn               =   3,
+-  Interrupt4_IRQn               =   4,
+-  Interrupt5_IRQn               =   5,
+-  Interrupt6_IRQn               =   6,
+-  Interrupt7_IRQn               =   7,
+-  Interrupt8_IRQn               =   8,
+-  Interrupt9_IRQn               =   9,
+-  /* Interrupts 10 .. 479 are left out */
+-  Interrupt480_IRQn             =   480
++/* -------------------  MPS3 AN524 Specific Interrupt Numbers ------------------------------ */
++  UART0_RX_IRQn               = 32,     /*  32 UART 0 receive interrupt         */
++  UART0_TX_IRQn               = 33,     /*  33 UART 0 transmit interrupt        */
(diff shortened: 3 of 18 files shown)


test/baremetal/platform/m33-an521/CROSS-BUILDING-AND-TESTING.md
```diff
@@ -0,0 +1,59 @@
...
+
+The build system automatically selects the optimal backend:
+
+- **Cortex-M33/M35 with OPT=1**: Uses DSP-optimized backend
+- **Other ARM targets with OPT=1**: Falls back to C backend (with warning)
+- **Any target with OPT=0**: Uses portable C backend
+
+#### Testing on QEMU
+
+A complete testing platform is available for m33-an521
```
hanno-becker: @jricardini It would be helpful to use AN524 instead as this targets the MPS3 FPGA prototyping board which @mkannwischer has available. This way, one could a) validate the setup on real HW, b) ad-hoc obtain performance numbers (without permission we cannot automatically publish them here).

AN524 appears to be supported by QEMU (https://www.qemu.org/docs/master/system/arm/mps2.html) as well.

test/baremetal/platform/m33-an521/startup.c
```diff
@@ -0,0 +1,666 @@
+/*
+ * Copyright (c) The mldsa-native project authors
+ * SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
```
hanno-becker: How was the startup code obtained? If it is derived / copied from somewhere, we need to adhere to and retain the license.

Generally, I'd prefer if we could stick to official CMSIS code as much as possible. See [here](https://github.com/slothy-optimizer/pqmx/tree/main/envs/m55-an547/src) for how it was done in pqmx for AN547.

test/baremetal/platform/m33-an521/comprehensive_func_test.c
```diff
@@ -0,0 +1,310 @@
+/*
+ * Copyright (c) The mldsa-native project authors
+ * SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+ *
+ * ML-DSA Comprehensive Functional Test for Cortex-M33
+ *
+ * Tests: sign/verify, unaligned access, invalid inputs, external mu,
+ *        pre-hash, and public key derivation
+ */
```
hanno-becker: @jricardini Can we work with the existing tests rather than introducing new ones? I assume the primary reason to have a separate file here is the custom `printf`, but can that not be supplied at link-time through an additional source file, rather than adjusting the test?

cf how this is done for the existing AN547 baremetal environment: If you go into the `arm-embedded` shell, `ls $M55_AN547_PATH` will show you the various startup files, including semihosting and uart setup.

hanno-becker (changes requested):
Thank you @jricardini, I support the addition of a baremetal environment for Cortex-M33.

The primary areas of work, for me (@mkannwischer please chime in), are:
* Switching to AN524 so we can ensure that this works both on QEMU and on real HW (@mkannwischer has an MPS3, and we could even hook it into CI)
* Generalizing and reusing the existing test files rather than copy-adjusting them.

Probably optional, but strongly desired:
* Add CI job exercising this baremetal environment.
* Consider using CMSIS code as in pqmx (_ideally_, the environment moves into pqmx as for AN547, but I think this could be done in a follow-up)

test/baremetal/platform/m33-an521/comprehensive_func_test.c
jricardini: Sure. I'll do it. It as in my plans to check how all the startup files are setup for AN547 in nix and do the same for m33 anyway!

test/baremetal/platform/m33-an521/CROSS-BUILDING-AND-TESTING.md
jricardini: done.

test/baremetal/platform/m33-an521/startup.c
jricardini: Using official CMSIS now, with proper license adherence

test/baremetal/platform/m33-an521/comprehensive_func_test.c
jricardini: done.

nix/m33-an524-arm-none-eabi/default.nix
```diff
@@ -0,0 +1,27 @@
...
+
+  setupHook = writeText "setup-hook.sh" ''
+    export M33_AN524_PATH="$1/src/platform/"
+  '';
+
+  meta = {
+    description = "Platform files for the Cortex-M33 (AN524)";
+    homepage = "https://github.com/mldsa-native/mldsa-native";
+  };
+}
```
hanno-becker: I am a little confused about this. Why do we need this if we check in the platform files into the repository? For AN547, we don't check them in but import them from PQMX, which is a bit different.

jricardini: Good catch! I mimicked the AN547 setup without fully understanding Nix ; didn't realize the package is only needed for external files. 
Should I remove this and set M33_AN524_PATH directly in the shellHook, or would you prefer to handle that? 
Thanks for the feedback!

envs/m33-an524/README.md
```diff
@@ -0,0 +1,86 @@
...
+### Hardware Support
+- **Target**: ARM Cortex-M33 on MPS3-AN524 FPGA board
+- **QEMU**: Full support for `qemu-system-arm -M mps3-an524`
+- **Memory Map**: 
+  - Flash: 0x10000000 (Secure BRAM, 512KB)
+  - RAM: 0x70000000 (Secure DDR4, 2MB)
+- **Clock**: 250 MHz (matches STM32H5 series)
+
+### CMSIS Compliance
+- **Core CMSIS**: Complete ARM CMSIS v5 headers
```
hanno-becker: Could you indicate if you had to modify the Arm-originated files at all, and if so, how?

nix/m33-an524-cmsis/default.nix
```diff
@@ -0,0 +1,152 @@
+# Copyright (c) The mldsa-native project authors
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+
+{ stdenvNoCC
+, fetchurl
+}:
+
+let
```
hanno-becker, line 8: @mkannwischer Could you review the nix files?

nix/m33-an524-cmsis/default.nix
```diff
@@ -0,0 +1,152 @@
...
+  # CMSIS_5 files from ARM repository (using develop branch commit with V5.3.0)
+  cmsis5_core_cm33 = fetchurl {
+    url = "https://raw.githubusercontent.com/ARM-software/CMSIS_5/55b19837f5703e418ca37894d5745b1dc05e4c91/CMSIS/Core/Include/core_cm33.h";
+    sha256 = "0wjb20lr31kbfqq5hc6bx7pvl13crs12xvjzv6sz8q11wm83qqvm";
+  };
+
+  cmsis5_cmsis_compiler = fetchurl {
```
hanno-becker: Why are we fetching all files separately rather than fetching once and copying the relevant files?

envs/m33-an524/patches/cmsis5/core_cm33.h.patch
```diff
@@ -0,0 +1,1193 @@
...
++  __IM uint32_t DEVTYPE;                /*!< Offset: 0xFCC (R/ )  Device Type Identifier Register */
+ } DWT_Type;
+ 
+ /* DWT Control Register Definitions */
+@@ -1221,49 +1185,49 @@ typedef struct
+ #define DWT_CTRL_NUMCOMP_Msk               (0xFUL << DWT_CTRL_NUMCOMP_Pos)             /*!< DWT CTRL: NUMCOMP Mask */
+ 
+ #define DWT_CTRL_NOTRCPKT_Pos              27U                                         /*!< DWT CTRL: NOTRCPKT Position */
+-#define DWT_CTRL_NOTRCPKT_Msk              (0x1UL << DWT_CTRL_NOTRCPKT_Pos)            /*!< DWT CTRL: NOTRCPKT Mask */
++#define DWT_CTRL_NOTRCPKT_Msk              (1UL << DWT_CTRL_NOTRCPKT_Pos)              /*!< DWT CTRL: NOTRCPKT Mask */
```
hanno-becker: A lot of changes are formatting and unnecessarily increase the patch. Could you remove them? I assume this stems from `format` mangling them while they were in the mldsa-native source tree?

envs/m33-an524/patches/cmsis5/cmsis_version.h.patch
```diff
@@ -0,0 +1,62 @@
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+--- /tmp/cmsis_version.h.orig	2026-01-13 17:40:28.872287080 +0000
++++ envs/m33-an524/src/platform/cmsis_version.h	2026-01-13 15:11:08.455800010 +0000
+@@ -1,11 +1,5 @@
+-/**************************************************************************//**
+- * @file     cmsis_version.h
+- * @brief    CMSIS Core(M) Version definitions
+- * @version  V5.0.5
+- * @date     02. February 2022
+- ******************************************************************************/
```
hanno-becker: Why was this header removed?

envs/m33-an524/patches/cmsis5/core_cm33.h.patch
```diff
@@ -0,0 +1,1193 @@
...
++#define SCnSCB_CPPWR_SUS11_Msk             (1UL << SCnSCB_CPPWR_SUS11_Pos)             /*!< CPPWR: SUS11 Mask */
++
++#define SCnSCB_CPPWR_SU11_Pos              22U                                         /*!< CPPWR: SU11 Position */
++#define SCnSCB_CPPWR_SU11_Msk              (1UL << SCnSCB_CPPWR_SU11_Pos)              /*!< CPPWR: SU11 Mask */
++
++#define SCnSCB_CPPWR_SUS10_Pos             21U                                         /*!< CPPWR: SUS10 Position */
++#define SCnSCB_CPPWR_SUS10_Msk             (1UL << SCnSCB_CPPWR_SUS10_Pos)             /*!< CPPWR: SUS10 Mask */
++
++#define SCnSCB_CPPWR_SU10_Pos              20U                                         /*!< CPPWR: SU10 Position */
++#define SCnSCB_CPPWR_SU10_Msk             (1UL << SCnSCB_CPPWR_SU10_Pos)               /*!< CPPWR: SU10 Mask */
```
hanno-becker: I assume this is copied from another CMSIS file? Why was it moved here, and why do we need it?

envs/m33-an524/patches/cmsis5/core_cm33.h.patch
```diff
@@ -0,0 +1,1193 @@
...
+-  __IM  uint32_t PID3;                   /*!< Offset: 0xFEC (R/ )  ITM Peripheral Identification Register #3 */
+-  __IM  uint32_t CID0;                   /*!< Offset: 0xFF0 (R/ )  ITM Component  Identification Register #0 */
+-  __IM  uint32_t CID1;                   /*!< Offset: 0xFF4 (R/ )  ITM Component  Identification Register #1 */
+-  __IM  uint32_t CID2;                   /*!< Offset: 0xFF8 (R/ )  ITM Component  Identification Register #2 */
+-  __IM  uint32_t CID3;                   /*!< Offset: 0xFFC (R/ )  ITM Component  Identification Register #3 */
++  __IOM uint32_t ITCTRL;                  /*!< Offset: 0xF00 (R/W)  Integration Mode Control Register */
++        uint32_t RESERVED6[46U];
++  __IM  uint32_t DEVARCH;                 /*!< Offset: 0xFBC (R/ )  Device Architecture Register */
++        uint32_t RESERVED7[3U];
++  __IM  uint32_t DEVTYPE;                 /*!< Offset: 0xFCC (R/ )  Device Type Register */
```
hanno-becker: Why was this changed?

envs/m33-an524/patches/cmsis5/core_cm33.h.patch
```diff
@@ -0,0 +1,1193 @@
...
++    __IOM uint32_t DCRDR;
++    __IOM uint32_t DEMCR;
++          uint32_t RESERVED0[1U];
++    __IOM uint32_t DAUTHCTRL;
++    __IOM uint32_t DSCSR;
++  } CoreDebug_Type;
++
++/* Debug Halting Control and Status Register Definitions */
++#define CoreDebug_DHCSR_DBGKEY_Pos DCB_DHCSR_DBGKEY_Pos
++#define CoreDebug_DHCSR_DBGKEY_Msk DCB_DHCSR_DBGKEY_Msk
```
hanno-becker: Similar question as above: This looks like it was copied from somewhere? Why was this necessary?

envs/m33-an524/README.md
```diff
@@ -0,0 +1,213 @@
+[//]: # (SPDX-License-Identifier: CC-BY-4.0)
+
+# Cortex-M33 MPS3-AN524 Platform Environment
+
+This directory contains the complete CMSIS-compliant platform environment for Cortex-M33 on MPS3-AN524, providing 100% feature parity with the M55-AN547 reference implementation.
```
hanno-becker: This file seems AI generated. Can we vastly trim it down to the essence? To me, that's a) short description of target, b) build instructions (incl mention of nix shell), c) patch explanation.

nix/m33-an524-cmsis/default.nix
jricardini: No special reason, really. I just looked for the files I needed in the ARM repo and got the raw URL for each one. The files are scattered across different folders in the CMSIS repository, so fetching the entire root folder just to copy a few seemed like a waste. Happy to change the approach if you think it's better though!

envs/m33-an524/patches/cmsis5/core_cm33.h.patch
jricardini: cleaned that up.

envs/m33-an524/patches/cmsis5/core_cm33.h.patch
jricardini: cleaned that up.

envs/m33-an524/README.md
jricardini: Done.

envs/m33-an524/patches/cmsis5/core_cm33.h.patch
jricardini: cleaned that up.

envs/m33-an524/patches/cmsis5/cmsis_version.h.patch
jricardini: added back. Removed when trying to fix ./scripts/lint error when the files were checked in the repo.

envs/m33-an524/patches/cmsis5/core_cm33.h.patch
jricardini: yes. All of those were from format mangling. Fixed.

nix/m33-an524-cmsis/default.nix
jricardini: Changed the approach  fetching once and copying the relevant files.

test/baremetal/platform/m33-an524/patches/cmsis5/ARMCM33.patch
```diff
@@ -0,0 +1,131 @@
...
+ 
+@@ -91,14 +163,18 @@ typedef enum IRQn
+ 
+ 
+ /* --------  Configuration of Core Peripherals  ----------------------------------- */
+-#define __CM33_REV                0x0000U   /* Core revision r0p1 */
+-#define __SAUREGION_PRESENT       0U        /* SAU regions present */
++/* Configuration of the Cortex-M33 Processor and Core Peripherals */
++#define __CM33_REV                0x0004U   /* Core revision r0p4 */
++#define __SAUREGION_PRESENT       1U        /* SAU regions present */
```
hanno-becker, line 117: One of the comments must be wrong, because both the old and the new comment say `/* SAU regions present */`.

Why do we need to change this?

envs/m33-an524/patches/cmsis5/startup_ARMCM33.patch
```diff
@@ -0,0 +1,235 @@
...
+ 
+ #if defined ( __GNUC__ )
+@@ -129,9 +280,9 @@ extern const VECTOR_TABLE_Type __VECTOR_TABLE[496];
+  *----------------------------------------------------------------------------*/
+ __NO_RETURN void Reset_Handler(void)
+ {
+-  __set_PSP((uint32_t)(&__INITIAL_SP));
+-
+-  __set_MSPLIM((uint32_t)(&__STACK_LIMIT));
++  // Remove these lines - they're causing the UsageFault
```
hanno-becker: If they are causing a usage fault, are we overflowing the stack? The stack guard is a useful feature for us, we should keep it enabled if possible.

envs/m33-an524/patches/cmsis5/m33-an524.ld.patch
```diff
@@ -0,0 +1,76 @@
...
+  ******************************************************************************/
+@@ -20,38 +20,37 @@
+  * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  * See the License for the specific language governing permissions and
+  * limitations under the License.
++ *
++ * Modifications for MPS3 AN524:
++ * - Flash Base: 0x00000000 (Non-Secure BRAM, 512KB - QEMU kernel load address)
++ * - RAM Base: 0x60000000 (Non-Secure DDR4, using 2MB of 256MB available)
++ * - Stack Size: 64KB (increased for ML-DSA-87)
```
hanno-becker, line 21: Unless you use the reduced RAM config, 64K won't be enough

envs/m33-an524/patches/cmsis5/m33-an524.ld.patch
```diff
@@ -0,0 +1,76 @@
...
+-    <o0> Stack Size (in Bytes) <0x0-0xFFFFFFFF:8>
+-    <o1> Heap Size (in Bytes) <0x0-0xFFFFFFFF:8>
+-  </h>
++  Stack Size: 64KB (for ML-DSA-87)
++  Heap Size: 512KB (for cryptographic operations)
+   -----------------------------------------------------------------------------*/
+-__STACK_SIZE = 0x00000400;
+-__HEAP_SIZE  = 0x00000C00;
++__STACK_SIZE = 0x00010000;
++__HEAP_SIZE  = 0x00080000;
```
hanno-becker: We don't need any heap

envs/m33-an524/patches/cmsis5/system_ARMCM33.patch
```diff
@@ -0,0 +1,77 @@
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+--- /tmp/system_ARMCM33.c.orig	2026-01-13 17:06:53.809238718 +0000
++++ envs/m33-an524/src/platform/system_ARMCM33.c	2026-01-13 15:11:08.435799683 +0000
+@@ -46,7 +46,7 @@
+ /*----------------------------------------------------------------------------
+   Define clocks
+  *----------------------------------------------------------------------------*/
+-#define  XTAL            (50000000UL)     /* Oscillator frequency */
++#define  XTAL            (500000000UL)     /* Oscillator frequency - 250 MHz Matches STM32H5 series           */
```
hanno-becker: Why was this change made?

envs/m33-an524/patches/cmsis5/system_ARMCM33.patch
```diff
@@ -0,0 +1,77 @@
...
+ #endif
+ 
+-#if defined (__ARM_FEATURE_CMSE) && (__ARM_FEATURE_CMSE == 3U)
+-  TZ_SAU_Setup();
++  /* Enable fault handlers */
++  SCB->SHCSR |= (SCB_SHCSR_MEMFAULTENA_Msk | SCB_SHCSR_BUSFAULTENA_Msk |
++                 SCB_SHCSR_USGFAULTENA_Msk);
++
++#ifdef __ARM_FEATURE_CMSE
++  SCB->SHCSR |= SCB_SHCSR_SECUREFAULTENA_Msk; /* Enable SecureFault handler (TrustZone) */
```
hanno-becker: Why was this change made?

envs/m33-an524/patches/cmsis5/m33-an524.ld.patch
mkannwischer: Even with the reduced RAM config, it won't be enough for now. Don't forget that you need to store keys and signature in addition to the local stack.

test/baremetal/platform/m33-an524/README.md
```diff
@@ -0,0 +1,34 @@
+[//]: # (SPDX-License-Identifier: CC-BY-4.0)
```
hanno-becker, line 2: Can you move this under `test/baremetal/platform/m33-an524`? I don't think we want a new toplevel `envs`.

envs/m33-an524/patches/cmsis5/system_ARMCM33.patch
jricardini: I change it back, to be like closer to the original CMSIS files.

PR #864, opened 2026-01-08
testing: do not generate custom build for rng failure test

L-series:
As a follow up to e15b34f, we modify the rng failure test so that it
does not generate a new build. To achieve this we do the following:

* Modify compontents and rules to not generate a cusom build.
* Remove the custom test_rng_fail config.
* Do not link the rng failure test against notrandombytes.c and instead
  implement our own randombytes.
* Updated test function names to use standard crypto_sign_* API
  instead of mld_* aliases as the mld_* namespace is no longer available
  as we are now using the default build with
  MLD_DEFAULT_NAMESPACE_PREFIX. Those functions which are not exposed
  in this way by default are wrapped with MLD_API_NAMESPACE.


test/notrandombytes/notrandombytes.c
```diff
@@ -37,6 +37,9 @@ static uint32_t in[12];
 static uint32_t out[8];
 static int32_t outleft = 0;
 
+int randombytes_counter __attribute__((weak)) = 0;
```
hanno-becker: Why do we need to change `notrandombytes`? I was hoping we just don't link that when we build the RNG failure test.

L-series: Sorry, I misunderstood what you meant in the comments on the previous PR. I've reimplemented it so that we do not link against notrandombytes and instead implement a custom randombytes in the rng_fail test. I've also removed the custom config as it is not actually needed. Please let me know if this is closer to what you intended!

Also (off topic), this first implementation made me realise that `ld` on linux and macos are quite different.. The mac version doesn't have the equivalent of gnu ld's `--wrap` flag.

test/src/test_rng_fail.c
```diff
@@ -25,26 +24,34 @@
...
   {
     return -1;
   }
 
-  return randombytes(out, outlen);
+  /* Fill the output buffer with random bytes */
+  for (i = 0; i < outlen; i++)
+  {
+    out[i] = (uint8_t)(((unsigned int)rand() * outlen) % 256);
+  }
```
hanno-becker: I would prefer `notrandombytes` at the core to remain consistent in terms of the underlying test-PRNG. I see that this leads to a name-clash, though.

One way one would hack this is to `#include "notrandombytes.c"` and just bracket it with `#define randombytes notrandombytes ... #undef randombytes`, and then call `notrandombytes` here.

But perhaps I'm being too pedantic, @mkannwischer WDYT?

test/src/test_rng_fail.c
```diff
@@ -126,15 +133,15 @@ static int test_sign_rng_failure(void)
 
   /* Generate valid keypair first */
   reset_all();
-  if (mld_keypair(pk, sk) != 0)
+  if (crypto_sign_keypair(pk, sk) != 0)
```
hanno-becker, line 132: Why those changes?

hanno-becker (changes requested):
Thank you @L-series, in principle I agree with the approach taken in this PR, which will help to reduce CI runtime.

There are a number of SUPERCOP related changes which don't seem to belong in this PR.

Beyond that, I wonder if there is a way we can use `notrandombytes` at the core of the instrumented PRNG still.

test/src/test_rng_fail.c
L-series: Isn't the default library in the tests compiled without a custom config and hence without `MLD_CONFIG_NAMESPACE_PREFIX: mld`? When i was using the mld_ prefixed functions I was seeing implicit declaration errors. The supercop naming convensions and how the libraries concat custom prefixes when building is still slightly fuzzy to me.

hanno-becker: You are right. this is also in line with the other tests so far.

I prefer to get rid of the SUPERCOP API ultimately, but for now this change is correct.

test/src/test_rng_fail.c
mkannwischer: I don't think we want the stdlib dependency here. 
I'm fine with the approach @hanno-becker proposes.
I'd also be fine with just setting the output to zero here - it shouldn't matter.

hanno-becker: Just using 0 is OK by me too.

hanno-becker: I took the liberty to rebase the PR. I also changed my mind regarding the 0'ing: I prefer consistency, and so switched to `notrandombytes` as suggested in my earlier comment.

Issue #857, opened 2026-01-07
Port: Use consistent architecture identifier in `autogen`


mkannwischer:
https://github.com/pq-code-package/mlkem-native/pull/1456

PR #865, opened 2026-01-09
Port: use consistent architecture identifier

willieyz:
- Resolves: #857 
- This PR ports the change from https://github.com/pq-code-package/mlkem-native/pull/1456 to mldsa-native.

Currently, mldsa-native already uses x86_64 as the architecture identifier for x86 in its existing HOL-Light proofs, and no changes are required in that regard. This commit simply aligns the surrounding scripts and infrastructure to consistently use x86_64 as the architecture name.

- Adds the related scripts to support dual architectures by introducing an explicit arch parameter in the autogen process.

- Replaces the arm_or_x86 flag with a unified arch flag in test scripts.


mkannwischer (approved):
Thanks @willieyz. LGTM!

PR #869, opened 2026-01-10
mldsa_native.h: Add PCT-aware MLD_TOTAL_ALLOC constants

mkannwischer:
Split KEYPAIR allocation constants into _NO_PCT and _PCT variants to
accurately reflect memory usage with and without pairwise consistency
testing (PCT). MLD_TOTAL_ALLOC_*_KEYPAIR now automatically selects
the appropriate value based on MLD_CONFIG_KEYGEN_PCT.

For legacy configs, assume PCT is enabled (conservative).

Also correct REDUCE_RAM KEYPAIR values which were previously using
PCT values unconditionally.

The default alloc test config no longer enables PCT; PCT is tested
via config-variations.


mldsa/mldsa_native.h
```diff
@@ -827,33 +827,52 @@ int MLD_API_NAMESPACE(pk_from_sk)(
  */
 /* check-magic: off */
 #if defined(MLD_API_LEGACY_CONFIG) || !defined(MLD_CONFIG_REDUCE_RAM)
-#define MLD_TOTAL_ALLOC_44_KEYPAIR 56640
+#define MLD_TOTAL_ALLOC_44_KEYPAIR_NO_PCT 45248
```
hanno-becker, line 905: Nit: I would suggest `MLD_TOTAL_XXX` (no PCT) as the default, and `MLD_TOTAL_XXX_PCT` for configurations _with_ PCT. The `MLK_CONFIG_KEYGEN_PCT` is also a _positive_ option that's disabled by default.

mkannwischer: hmm, lower in the file `MLD_TOTAL_XXX` is defined depending on the actual value of `MLK_CONFIG_KEYGEN_PCT`.

hanno-becker: Nevermind, I hadn't yet seen the wrapper below.

hanno-becker (approved):
LGTM

PR #875, opened 2026-01-11
autogen: fix print issue with narrow terminals

L-series:
This commit replaces the hardcoded width of 160 in `clear_status_line`
with the width of ones terminal window. This resolves the issue in which
we see autogen generate large amounts of new lines for narrow terminals
(width <160).


mkannwischer (approved):
Thansk @L-series! LGTM.

PR #877, opened 2026-01-12
autogen: Use `rich` for progress bar

hanno-becker:
Port of https://github.com/pq-code-package/mlkem-native/pull/1474


mkannwischer (approved):
Thanks @hanno-becker. Works great.

There are a few minor differences to the mlkem-native `autogen`, but the way I see it, this should be changed in mlkem-native.

Issue #853, opened 2026-01-07
Remove unnecessary call to `mld_polyveck_reduce`


hanno-becker:
https://github.com/pq-code-package/mldsa-native/blob/ed79e3f8303bae1e633259c7d6be7bd48b161941/mldsa/src/sign.c#L524-L527 The call `mld_polyveck_reduce(w0)` seems unnecessary since the post-condition of `mld_polyvec_matrix_pointwise_montgomery(w0, mat, z)` and the pre-condition of `mld_polyveck_invntt_tomont(w0)` already line up.

mkannwischer:
Same here: 
https://github.com/pq-code-package/mldsa-native/blob/ed79e3f8303bae1e633259c7d6be7bd48b161941/mldsa/src/sign.c#L248

PR #879, opened 2026-01-13
Remove extra reduction from keygen and signing

mkannwischer:
The post-condition of `mld_polyvec_matrix_pointwise_montgomery` guarantees coefficients to be of absolute value less than q which lines up with the pre-condition of `mld_polyveck_invntt_tomont`.

This commit removes an extra reduction between
`polyvec_matrix_pointwise_montgomery` and `mld_polyveck_invntt_tomont` in two places: Once in key generation and once in signing.

- Resolves https://github.com/pq-code-package/mldsa-native/issues/853


hanno-becker (approved):
Those changes are the best :smile: Make the CBMC method shine.

PR #883, opened 2026-01-14
Add hol-server for programmatic HOL Light communication

mkannwischer:
Integrates hol_server (https://github.com/monadius/hol_server) to enable TCP-based communication with HOL Light. This allows sending commands programmatically via netcat or the VS Code extension.

Usage: hol-server [port]  # default port is 2012

Update documentation accordingly.

### DIFF
--- a/nix/hol_light/hol-server.sh
+++ b/nix/hol_light/hol-server.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+# Copyright (c) The mldsa-native project authors
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+#
+# HOL Light server for programmatic communication
+# Based on https://github.com/monadius/hol_server
+#
+# Usage: hol-server [port]
+
+set -e
+
+PORT=${1:-2012}
+
+# These are replaced by nix
+HOL_LIGHT_DIR="@hol_light@/lib/hol_light"
+HOL_SERVER_SRC="@hol_server_src@"
+
+# cd to x86_64 proofs directory if in mldsa-native repo
+PROOF_DIR="$(git rev-parse --show-toplevel 2>/dev/null)/proofs/hol_light/x86_64"
+[ -d "$PROOF_DIR" ] && cd "$PROOF_DIR"
+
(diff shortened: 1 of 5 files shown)


nix/hol_light/hol-server.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+# Copyright (c) The mldsa-native project authors
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+#
+# HOL Light server for programmatic communication
+# Based on https://github.com/monadius/hol_server
```
hanno-becker, line 6: If this is based on https://github.com/monadius/hol_server, is the license correct?

proofs/hol_light/README.md
```diff
@@ -0,0 +1,52 @@
+[//]: # (SPDX-License-Identifier: CC-BY-4.0)
+
+# HOL Light functional correctness proofs
+
+This directory contains functional correctness proofs for AArch64 and x86_64 assembly routines
+used in mldsa-native. The proofs are written in the [HOL Light](https://hol-light.github.io/) theorem
+prover, utilizing the assembly verification infrastructure from [s2n-bignum](https://github.com/awslabs/s2n-bignum).
+
+Each function is proved in a separate `.ml` file in [x86_64/proofs/](x86_64/proofs). Each file
+contains the byte code being verified, as well as the specification that is being proved.
```
hanno-becker: A bit confusing that the first paragraph mentions Arm and x86, and the second only x86.

nix/hol_light/hol-server.sh
mkannwischer: "based" as in, it's using it as a nix dependency. 

The only thing this file has in common with the hol_server README is
```
#directory "+threads";;
#load "unix.cma";;
#load "threads.cma";;
```

which should be sufficiently trivial?

hanno-becker: Agree

proofs/hol_light/README.md
mkannwischer: Fixed.

PR #884, opened 2026-01-15
sys.h: Fix __attribute__ usage for non-GCC compilers

mkannwischer:
Restructure `MLD_INLINE` and `MLD_ALWAYS_INLINE` into separate blocks,
each independently guarded:

`MLD_INLINE`:
- MSVC: `__inline`
- C99 or `inline` macro defined: `inline`
- GCC/Clang C90: `__attribute__((unused))` to silence warnings
- Other C90: empty

`MLD_ALWAYS_INLINE`:
- MSVC: `__forceinline`
- GCC/Clang C99+: `MLD_INLINE __attribute__((always_inline))`
- Other: `MLD_INLINE` (no forced inlining)

This fixes two issues:
1. If `inline` was defined as a macro before including sys.h, the MSVC
   check was bypassed and `__attribute__` was used, causing MSVC failures.
2. For non-MSVC/GCC/Clang compilers like IAR, `__attribute__` was used
   unconditionally, causing build failures.

Also fix typo: `defined(clang)` -> `defined(__clang__)`

### DIFF
--- a/mldsa/src/sys.h
+++ b/mldsa/src/sys.h
@@ -106,34 +106,42 @@
 #endif
 
 /*
- * C90 does not have the inline compiler directive yet.
- * We don't use it in C90 builds.
- * However, in that case the compiler warns about some inline functions in
- * header files not being used in every compilation unit that includes that
- * header. To work around it we silence that warning in that case using
- * __attribute__((unused)).
+ * MLD_INLINE: Hint for inlining.
+ * - MSVC: __inline
+ * - C99+: inline
+ * - GCC/Clang C90: __attribute__((unused)) to silence warnings
+ * - Other C90: empty
  */
-
-/* Do not use inline for C90 builds*/
 #if !defined(MLD_INLINE)
-#if !defined(inline)
 #if defined(_MSC_VER)
 #define MLD_INLINE __inline
-/* Don't combine __inline and __forceinline */
-#define MLD_ALWAYS_INLINE __forceinline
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm (approved):
Thank you very much! I can confirm that 33cac9eb1f728ffbecef8570d284e0a4a3774f37 fixes the build¹ with IAR 9 (which TF-PSA-Crypto officially supports due to user demand), and also ARM Compiler 5 (which neither Arm not TF-PSA-Crypto officially supports any longer). I also tried other compilers I could find on my Ubuntu machine (and that we don't care about): [tcc](https://bellard.org/tcc/) implements enough GCC extensions that it was happy before, [bcc](https://packages.debian.org/trixie/bcc) doesn't provide `stdint.h`², [pcc](https://sdcc.sourceforge.net/)'s preprocessor seems to get confused and I don't care enough to see if that's fixable, and [sdcc](https://sdcc.sourceforge.net/) now compiles with warnings that suggest that the library won't work on a 16-bit platform³.

¹ I only tried `${CC} -c -I. -DMLD_CONFIG_PARAMETER_SET=87 mldsa_native.c` with a toy definition of `mld_zeroize` in `mldsa_native_config.h`. I'll see if I can run tests when I've made progress with our integration.

² I think that means your claim of “portable C90” implementation in `README.md` is wrong: `stdint.h` was a popular extension to C90, but it only became standard in C99.

³ I'm not sure whether it's realistic that a platform with 16-bit `int` could run ML-DSA, but this should be documented as well.

PR #891, opened 2026-01-16
Eliminate SUPERCOP API from tests

### LINKED ISSUE #878: Eliminate SUPERCOP API from tests
Our tests currently use the SUPERCOP API to call in mldsa-native and have to define a couple of additional SUPERCOP-style symbols:
https://github.com/pq-code-package/mldsa-native/blob/68490efc3389e21f172e46c3d12f328505b8f0f9/test/acvp/acvp_mldsa.c#L15-L25

We should just switch this to the regular API.

willieyz:
- Resolves: #878

- The SUPERCOP API (crypto_sign_*) is a compatibility layer intended for SUPERCOP benchmarking.
Previously, the tests defined a series of SUPERCOP-style names for internal functions (the _internal variants), even though these tests are not part of the standard SUPERCOP API.

- Instead of defining SUPERCOP-related macros in every test source file, this commit updates those macros to target the regular mldsa-native API rather than the SUPERCOP compatibility layer.
As a result, the tests invoke the standard API directly while still using macros for name mapping.

Internal functions are now referenced via:
```
#define keypair_internal MLD_API_NAMESPACE(keypair_internal)
```

### DIFF
--- a/test/acvp/acvp_mldsa.c
+++ b/test/acvp/acvp_mldsa.c
@@ -11,18 +11,18 @@
 
 #include "mldsa_native.h"
 
-/* Additional SUPERCOP-style macros for functions not in the standard set */
-#define crypto_sign_keypair_internal MLD_API_NAMESPACE(keypair_internal)
-#define crypto_sign_signature_internal MLD_API_NAMESPACE(signature_internal)
-#define crypto_sign_verify_internal MLD_API_NAMESPACE(verify_internal)
-#define crypto_sign_verify_extmu MLD_API_NAMESPACE(verify_extmu)
-#define crypto_sign_signature_pre_hash_internal \
+#define mld_sign_keypair_internal MLD_API_NAMESPACE(keypair_internal)
+#define mld_sign_signature_internal MLD_API_NAMESPACE(signature_internal)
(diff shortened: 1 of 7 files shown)


test/acvp/acvp_mldsa.c
mkannwischer: Still using the SUPERCOP API.

test/bench/bench_mldsa.c
mkannwischer: still using the SUPERCOP API.

mkannwischer: same here.

test/src/gen_KAT.c
mkannwischer: still using the SUPERCOP API

test/src/test_mldsa.c
mkannwischer, line 1: Using the SUPERCOP API in 25 places....

mkannwischer (changes requested):
Thanks @willieyz. 

You forgot to modify `test_stack.c` and `test_rng_fail.c`. In the files you changes you forgot a lot of functions.

mkannwischer:
- Superseded by https://github.com/pq-code-package/mldsa-native/pull/1248.

PR #892, opened 2026-01-17
Add HOL Light proof for aarch64 poly_caddq

mkannwischer:
Add the first HOL Light functional correctness proof for an aarch64 ML-DSA function: poly_caddq, which performs conditional addition of q to reduce polynomial coefficients from (-q, q) to [0, q).

This commit includes:
- HOL Light proof infrastructure for aarch64 (Makefile, utilities)
- Functional correctness proof showing the assembly computes `ival(x i) rem &8380417` for each coefficient
- autogen support for generating aarch64 HOL Light assembly

- Resolves https://github.com/pq-code-package/mldsa-native/issues/493

TODO before we merge
- [x] Merge addition of `mla` back to s2n-bignum
- [x] Autogenerate embedded bytecode in proof
- [x] Add according CBMC spec and proof

PR #893, opened 2026-01-17
[TEST] Update s2n-bignum

mkannwischer:
Debugging some issues I am getting with the x86_64 dumpbyte code after updating s2n-bignum in #892.


mkannwischer:
Updating HOL-Light fixes the problem.

PR #894, opened 2026-01-18
CI/CBMC: Report CBMC runtime

hanno-becker:
This commit adds to the CBMC CI action another step which, for PRs,
- emits one comment per security level showing the CBMC runtimes and the relative slowdown/speedup compares to the last run on `main`
- warns about severe proof regressions (>50% for now) and failures.

Comments are updated when the CI is run multiple times on the same PR.


mkannwischer (approved):
Thanks @hanno-becker. I validated that the changes are aligned with the changes in mlkem-native.

Issue #904, opened 2026-01-21
Port: Unit tests: Add native Keccak x1 and x4 tests


mkannwischer:
https://github.com/pq-code-package/mlkem-native/pull/1516

This should be based on top of https://github.com/pq-code-package/mldsa-native/pull/855.

PR #855, opened 2026-01-07
Unit-test: add consistency tests for mld_poly_uniform*_x1/x4

willieyz:
- Resolves: #658 

This PR adds consistency tests to test_unit.c to verify that the scalar and 4-way batched implementations produce identical results.

In mldsa-native, there are three functions that provide both scalar and 4-way batched variants:
```
mld_poly_uniform_gamma1 / mld_poly_uniform_gamma1_4x
mld_poly_uniform_eta    / mld_poly_uniform_eta_4x
mld_poly_uniform        / mld_poly_uniform_4x
```
For each of the above pairs, this PR implements consistency test functions in test_unit.c that compare the outputs of the scalar and 4-way batched variants.

For mld_poly_uniform_eta and mld_poly_uniform_eta_4x, the two variants are not defined under the same compilation conditions in poly_kl.h. To enable testing both variants, this PR introduces a new macro, `MLD_UNIT_TEST`, which is made available in `test_unit.c` to override the conditional compilation and allow both implementations to be exercised in the unit tests.

PR #910, opened 2026-01-23
Port: Unit tests: Add native Keccak x1 and x4 tests

willieyz:
- Resolves: #904 
- Based on: #855 
- Ported from: https://github.com/pq-code-package/mlkem-native/pull/1516

Add unit tests for keccakf1600 x1 and x4 permutation functions that
compare native implementations against the C reference implementation.

- test_keccakf1600_permute: tests x1 permutation when
  MLD_USE_FIPS202_X1_NATIVE
- test_keccakf1600x4_permute: tests x4 permutation when
  MLD_USE_FIPS202_X4_NATIVE
- Expose mld_keccakf1600_permute_c via MLD_STATIC_TESTABLE for testing

### DIFF
--- a/test/src/test_unit.c
+++ b/test/src/test_unit.c
@@ -8,15 +8,16 @@
 #include <string.h>
 #include "../notrandombytes/notrandombytes.h"
 
+#include "../../mldsa/src/fips202/keccakf1600.h"
 #include "../../mldsa/src/poly.h"
 #include "../../mldsa/src/poly_kl.h"
 #include "../../mldsa/src/polyvec.h"
 
 #ifndef NUM_RANDOM_TESTS
 #ifdef MLDSA_DEBUG
-#define NUM_RANDOM_TESTS 1000
+#define NUM_RANDOM_TESTS 800
 #else
-#define NUM_RANDOM_TESTS 5000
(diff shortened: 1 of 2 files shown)


test/src/test_unit.c
```diff
@@ -8,15 +8,16 @@
...
 #include "../../mldsa/src/poly_kl.h"
 #include "../../mldsa/src/polyvec.h"
 
 #ifndef NUM_RANDOM_TESTS
 #ifdef MLDSA_DEBUG
-#define NUM_RANDOM_TESTS 1000
+#define NUM_RANDOM_TESTS 800
 #else
-#define NUM_RANDOM_TESTS 5000
+#define NUM_RANDOM_TESTS 8000
```
mkannwischer: This change does not make sense. It should be 5000 or 4000

test/src/test_unit.c
mkannwischer, line 1: You forgot to port the changes to the `main` function.

test/src/test_unit.c
mkannwischer: Fixed.

test/src/test_unit.c
mkannwischer: Fixed.

Issue #886, opened 2026-01-15
CI: Test OpenTitan integration


mkannwischer:
With https://github.com/zerorisc/expo/pull/184 merged, we should make sure that we don't break the integration with future changes.

https://github.com/pq-code-package/mlkem-native/blob/main/.github/workflows/integration-opentitan.yml can be an inspiration, but it has to be adapted as we cannot afford to run verilator-based tests in CI (they run 1.5 hours on my laptop). We can try only testing one parameter set. If that is still to slow, we may only be able to check that it compiles.

PR #928, opened 2026-01-31
CI: Add OpenTitan build test

mkannwischer:
Add CI workflow to verify that mldsa-native builds correctly
within the OpenTitan (expo) build system. Only builds the test
binary without running the Verilator simulation, as ML-DSA is
too slow for simulation in CI.

- Resolves https://github.com/pq-code-package/mldsa-native/issues/886

### DIFF
--- a/.github/actions/setup-opentitan/action.yml
+++ b/.github/actions/setup-opentitan/action.yml
@@ -1,0 +1,79 @@
+# Copyright (c) The mldsa-native project authors
+# Copyright (c) The mlkem-native project authors
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+
+name: Setup OpenTitan
+description: Setup OpenTitan for build testing
+
+inputs:
+  expo-repository:
+    description: 'Expo repository to clone'
+    required: false
+    default: 'https://github.com/zerorisc/expo'
+  expo-commit:
+    description: 'Expo commit to checkout'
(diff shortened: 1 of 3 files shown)


.github/actions/setup-opentitan/action.yml
```diff
@@ -0,0 +1,79 @@
...
+        cd verilator
+        git checkout v$VERILATOR_VERSION
+        autoconf
+        CC=gcc-11 CXX=g++-11 ./configure --prefix=/tools/verilator/$VERILATOR_VERSION
+        CC=gcc-11 CXX=g++-11 make -j$(nproc)
+        sudo CC=gcc-11 CXX=g++-11 make install
+
+    - name: Add Verilator to PATH
+      shell: bash
+      run: |
```
hanno-becker, line 71: Not needed I suppose? Are we keeping it in case we want to simulate in CI at some point, after all?

hanno-becker (approved):
The Verilator setup seems unnecessary given that we only build, but it allows us to keep the setup action synchronized with mlkem-native, which is worthwhile.

.github/actions/setup-opentitan/action.yml
mkannwischer: That is what I hoped too. Unfortunately, just building the hardware requires to run verilator.
I have not found a way yet to run the build without it.

Issue #856, opened 2026-01-07
Use MLD_MUST_CHECK_RETURN_VALUE for all functions with return values


mkannwischer:
Enforcing internal error checking discipline. In particular, this applies to the backend APIs that can now signal fallback -- this must always be checked.
https://github.com/pq-code-package/mlkem-native/issues/1399

PR #931, opened 2026-02-01
Use MLD_MUST_CHECK_RETURN_VALUE for all functions with return values

f15hr:
Resolves https://github.com/pq-code-package/mldsa-native/issues/856

"Enforcing internal error checking discipline. In particular, this applies to the backend APIs that can now signal fallback -- this must always be checked."

### DIFF
--- a/dev/aarch64_clean/meta.h
+++ b/dev/aarch64_clean/meta.h
@@ -35,6 +35,7 @@
 #include "../api.h"
 #include "src/arith_native_aarch64.h"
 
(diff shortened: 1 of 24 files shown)


f15hr:
Changed `CBMC_OBJECT_BITS` to 12 in `proofs/cbmc/polyvec_matrix_expand/Makefile` to improve benchmark result for `polyvec_matrix_expand`.

mkannwischer (approved):
Thanks @f15hr. Everything is looking good. I double checked that you did not miss any function.

PR #934, opened 2026-02-02
aarch64: Don't perform unaligned memory accesses

flynd:
On a bare metal system with the MMU disabled or alignment checking enabled, accessing memory using an unaligned address causes an alignment fault exception.  By changing the memory accesses to smaller sizes that matches the alignment of the addresses, the code can run without depending on the CPU handling and accepting unaligned addresses.

Signed-off-by: Anders Sonmark <Anders.Sonmark@axis.com>

### DIFF
--- a/mldsa/src/fips202/native/aarch64/src/keccak_f1600_x4_v8a_scalar_hybrid_asm.S
+++ b/mldsa/src/fips202/native/aarch64/src/keccak_f1600_x4_v8a_scalar_hybrid_asm.S
@@ -68,43 +68,43 @@ MLD_ASM_FN_SYMBOL(keccak_f1600_x4_scalar_v8a_hybrid_asm)
         str	x0, [sp]
         add	x4, x0, #0xc8
         ldp	q25, q26, [x0]
-        ldp	q27, q28, [x4]
+        ld1	{v27.2d, v28.2d}, [x4], #0x20
         trn1	v0.2d, v25.2d, v27.2d
         trn2	v1.2d, v25.2d, v27.2d
         trn1	v2.2d, v26.2d, v28.2d
         trn2	v3.2d, v26.2d, v28.2d
         ldp	q25, q26, [x0, #0x20]
-        ldp	q27, q28, [x4, #0x20]
+        ld1	{v27.2d, v28.2d}, [x4], #0x20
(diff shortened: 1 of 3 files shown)


flynd:
> Thanks for your PR! Can you elaborate where exactly you got a fault due to an unaligned memory access? I took a quick glance at the changes, but I believe all of those addresses should be aligned as we are using `MLD_ALIGN` for the corresponding allocations. If you have expierenced an alignment fault, than that would hint that `MLD_ALIGN` is missing somewhere.

In keccak_f1600_x4_v8a_scalar_hybrid_asm.S, function keccak_f1600_x4_scalar_v8a_hybrid_asm is called with X0 aligned to 128 bits, but it then sets X4=X0+0xC8 at line 69, making X4 only 64 bit aligned.  The following "ldp q25, q26, [x0]" is thus fine, but "ldp q27, q28, [x4]" fails as it tries to read 2x128 bits from a 64 bit aligned pointer.  The same happens again when X4 is set up for writing at line 976.  My patch changes the reads and writes with X4 to 4x64 bits instead to fix this.

In polyz_unpack_19_asm.S, the loop increases X1 by 0x28, so even if it was 128 bit aligned the first time, it will be only 64 bit aligned in the second iteration.  My patch changes the reads from 128+128+64 bits to 5x64 bits.

In rej_uniform_asm.S, X7 starts out 128 bit aligned but the increase inside the loop at lines 108, 110, 112, and 114 only guarantees 32 bit alignment so my patch changes the writes from 128 bit to 4x32 bits.

It is possible that there are other unaligned accesses that I haven't noticed.  These occurrences are the ones that caused exceptions when I tried to verify ML-DSA-87 signatures on a Cortex A55 with the MMU disabled.  With these fixes, I can successfully verify signatures with native aarch64 backends for both arithmetic and fips202.

mkannwischer:
Thanks again @flynd for bringing this to our attention.

I have reproduced the issue in #939. It also revealed that you had missed a couple of other places (outside of verification, for other parameter sets, for other Keccak implementations) - #939 includes patches for all of them.

We will still need to properly integrate those into the `dev/` directory and adjust the HOL-Light proofs in mlkem-native.
Unfortunately, I am out of time for today, but we will definitely make sure this gets merged in the next days.

PR #939, opened 2026-02-03
AArch64: Use alignment-safe Neon loads/stores and test in CI

mkannwischer:
This PR fixes alignment issues in AArch64 Neon assembly that cause faults on bare-metal systems without an MMU or when alignment checking is enabled.

On AArch64, certain Neon load/store instructions like ldp/stp with Q-registers require 16-byte alignment. While Linux and most operating systems handle unaligned accesses transparently via the MMU, bare-metal environments running on Device memory fault immediately on misaligned accesses.

The batched Keccak implementations process two sequential 200-byte states. The first state is 16-byte aligned, but the second starts at offset 200 bytes, which is only 8-byte aligned. The ldp/stp instructions for Q-registers on this second pointer caused alignment faults. Similar issues existed in other assembly routines using unaligned addresses.

The fix replaces alignment-requiring instructions with their alignment-safe equivalents (ld1/st1), which have no alignment requirement and work correctly on both normal and Device memory.

This PR also adds a bare-metal AArch64-virt QEMU target that runs tests without an MMU, catching any remaining unaligned Neon accesses.

Thanks to @flynd who has reported this issue in https://github.com/pq-code-package/mldsa-native/pull/934.

### DIFF
--- a/dev/aarch64_clean/src/polyz_unpack_17_asm.S
+++ b/dev/aarch64_clean/src/polyz_unpack_17_asm.S
@@ -64,9 +64,8 @@ MLD_ASM_FN_SYMBOL(polyz_unpack_17_asm)
         mov count, #(64/4)
 
 polyz_unpack_17_loop:
-        ldr q1, [buf, #16]
-        ldr s2, [buf, #32]
-        ldr q0, [buf], #36
+        ld1	{v0.4s, v1.4s}, [buf], #0x20
+        ldr	s2, [buf], #0x04
 
         tbl v4.16b, {v0.16b}, idx0.16b
         tbl v5.16b, {v0.16b - v1.16b}, idx1.16b
--- a/dev/aarch64_clean/src/polyz_unpack_19_asm.S
+++ b/dev/aarch64_clean/src/polyz_unpack_19_asm.S
@@ -61,9 +61,8 @@ MLD_ASM_FN_SYMBOL(polyz_unpack_19_asm)
         mov count, #(64/4)
 
 polyz_unpack_19_loop:
-        ldr q1, [buf, #16]
-        ldr d2, [buf, #32]
-        ldr q0, [buf], #40
+        ld1	{v0.2d, v1.2d}, [buf], #0x20
+        ldr	d2, [buf], #0x08
 
         tbl v4.16b, {v0.16b}, idx0.16b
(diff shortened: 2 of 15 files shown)


dev/fips202/aarch64/src/keccak_f1600_x2_v84a_asm.S
```diff
@@ -159,46 +159,47 @@ keccak_f1600_x2_v84a_loop:
         eor	v0.16b, v0.16b, v31.16b
         sub	x2, x2, #0x1
         cbnz	x2, keccak_f1600_x2_v84a_loop
+       sub	x0, x0, #0xc0
         add	x2, x0, #0xc8
```
hanno-becker, line 163:
```suggestion
       add	x2, x0, #0x08
       sub	x0, x0, #0xc0
```

mkannwischer: Can we not do that microoptimization now, please? Let's try this out in mlkem-native first. If it makes a difference, we can port it to mldsa-native. Okay?

dev/fips202/aarch64/src/keccak_f1600_x2_v84a_asm.S
```diff
@@ -48,44 +48,44 @@ MLD_ASM_FN_SYMBOL(keccak_f1600_x2_v84a_asm)
         stp	d12, d13, [sp, #0x20]
         stp	d14, d15, [sp, #0x30]
         add	x2, x0, #0xc8
-        ldp	q25, q26, [x0]
-        ldp	q27, q28, [x2]
+        ldp	q25, q26, [x0], #32
+        ld1	{ v27.2d, v28.2d }, [x2], #32
```
hanno-becker, line 52: Not a blocker: It's a bit confusing that we mix `ldp` and `ld1` here, esp. given that there's no comment indicating that this is becuase of alignment.

dev/fips202/aarch64/src/keccak_f1600_x2_v84a_asm.S
hanno-becker: ack

hanno-becker (approved):
A nit on the mix of `ldp` and `ld1` being confusing, but it's not a blocker.

I tested the PR locally and it works as expected.

Issue #930, opened 2026-02-01
Port: Add actionlint to lint Github actions


mkannwischer:
https://github.com/pq-code-package/mlkem-native/pull/1537

PR #945, opened 2026-02-04
Port-Lint: Add actionlint to lint Github actions

willieyz:
- Resolves: #930

- This PR port: https://github.com/pq-code-package/mlkem-native/pull/1537 to mldsa-native, add actionlint to mldsa-native, and fix related syntax bug and style.

### DIFF
--- a/.github/actions/setup-nix/action.yml
+++ b/.github/actions/setup-nix/action.yml
@@ -62,8 +62,10 @@ runs:
       shell: bash
       run: |
         if [[ -f /.dockerenv ]]; then
+          apt update
           apt install nix -y
         else
+          sudo apt update
           sudo apt install nix -y
         fi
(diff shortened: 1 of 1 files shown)


.github/actionlint.yaml
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) The mlkem-native project authors
+# Copyright (c) The mlkem-native project authors
```
mkannwischer: This says mlkem twice.

mkannwischer (changes requested):
Thanks @willieyz. Mostly good, one nit, but I will change it myself shortly.

.github/actionlint.yaml
mkannwischer: fixed.

Issue #941, opened 2026-02-03
Config flag to only build for signature verification


flynd:
For an application that only needs to verify signatures, not create them (for example a boot stage implementing secure boot), it would be convenient to have a config option removes everything that isn't needed for verification.

Although making everything static (with MLD_CONFIG_EXTERNAL_API_QUALIFIER and MLD_CONFIG_INTERNAL_API_QUALIFIER) allows the compiler to drop C functions and tables that aren't used (but requires -Wno-unused-function to remove warnings), this is only a partial solution as it won't exclude unused functions from the native assembly backends such as poly_decompose_XX_asm and rej_uniform_etaX_asm.

mkannwischer:
Thanks @flynd! Yes, I support such a configuration option.

Probably the easiest is in the form of 
- `MLD_CONFIG_NO_KEYPAIR_API`
- `MLD_CONFIG_NO_SIGN_API`
- `MLD_CONFIG_NO_VERIFY_API`

Aligned with the pre-existing `MLD_CONFIG_NO_RANDOMIZED_API`.

That said, I don't think I will find the time to implement this this month.
If you need it urgently, maybe you could open a PR @flynd? That would be appreciated!

flynd:
> Thanks [@flynd](https://github.com/flynd)! Yes, I support such a configuration option.
> 
> Probably the easiest is in the form of
> 
> * `MLD_CONFIG_NO_KEYPAIR_API`
> * `MLD_CONFIG_NO_SIGN_API`
> * `MLD_CONFIG_NO_VERIFY_API`
> 
> Aligned with the pre-existing `MLD_CONFIG_NO_RANDOMIZED_API`.
> 
> That said, I don't think I will find the time to implement this this month. If you need it urgently, maybe you could open a PR [@flynd](https://github.com/flynd)? That would be appreciated!

This is not urgent for me as it is manageable with only a few local changes right now, with most of the work being done by the compiler to just discard everything it doesn't need when everything is static.  When #936 has been fixed, I'll consider whether to try and do this myself.

flynd:
Fixed with #1000 .

PR #953, opened 2026-02-06
native: Add .size directive to all functions

flynd:
Add size information to function symbols.  The makes the assembly functions show up with their correct size in the elf instead of always having a zero size, making it easier to see how much space each function takes up.


mkannwischer:
Closing as it has been solved in #956. Thanks again for the report @flynd.

flynd:
> Btw, @flynd, if you want to contribute more PRs, we should add you as a contributor, so you can open PRs from a local branch to run the full CI. Let us know if you would be interested in that. The way you currently do it is also fine though as you are suggesting very useful improvements. I'm happy to keep polishing your PRs and get them through CI.

Thank you.  I only expect to work with this for a limited time right now to finish the current project, then I don't know when I'll have a reason to work with mldsa-native again.  So I will probably contribute a few more things short term (I'm currently looking into #941) but after that I don't know if I'll be active.  It is not unlikely that we will need it for more things so it could definitely happen.  If you think it would be better for me to work directly as a contributor then we can do that, but I don't expect to understand the project well enough to not need polishing just yet though.

PR #954, opened 2026-02-06
native: Give single assembly unit a unique name

flynd:
Give the single compilation units unique names by renaming the assembly file.  This removes the need for having unique names for object files based on the source language and allows simpler rules, for example "%.o: %.c" and "%.o: %.S".


hanno-becker:
No concerns, this is a simple patch for AWS-LC (which we will be forced to provide already when we merge this change here, to appease integration-CI). I support this change.

Issue #789, opened 2025-12-11
`siglen` parameter in `crypto_sign_signature` is prone to API misuse


davidchisnall:
The `crypto_sign_signature` function takes a pointer that it will set to the length of the output buffer.  This is less than ideal for good ergonomics:

 - The parameter makes it appear as if the length of the signature is dynamic, but it is not, it is always `MLDSA_CRYPTO_BYTES`.
 - The shape of the parameter pair (buffer followed by length) is sufficiently close to other uses where it is an input parameter that callers may expect it to be set on input.
 - It isn't checked internally, so if a caller *does* set it and it has the wrong value, then this is ignored.
 - A check of the size after the call is *almost* never useful.  There are two cases:
   - The caller provided a buffer that is too large.  The check *may* enable a call to `realloc` or some other approach to trim the buffer (though given that the size is invariant, this is almost always an error).
   - The caller provided a buffer that is too small.  In this case, the check of the returned size happens *after* the library has written past the end of a buffer and therefore the state of the value is undefined.

All of this adds up to a dangerous API pattern, where people write code that appears defensive but will actually corrupt memory.

mkannwischer:
Thanks @davidchisnall for bringing up this important issue - this also had bothered me before.

I'm torn, however:

One one hand I agree with you that there is no good reason to have this parameter in ML-DSA in the first place. It serves no techincal purpose and as @davidchisnall outlines it causes at least confusion and possibly API misuse.

On the other hand, this API his inherited from the reference implementations which implements it to fit into libraries like libOQS^ and PQClean (which do indeed support signature schemes with variable-length signatures). I consider keeping mldsa-native compatible with these implementations important to make it easy to switch to mldsa-native.

^ More context: The `crypto_sign` (which has the same problem), and `crypto_sign_open` APIs are actuall inherited from the SUPERCOP project and were mandated by NIST for reference implementations. This is not the case for the `crypto_sign_signature` and the `crypto_sign_verify` APIs - those were primarily required by libOQS if I remember correctly. 

That said, even if we want to maintain compatibility, this does not have to be the default API. We could consider switchting to simpler, less confusing/dangerous API (i.e., just dropping the siglen/smlen from signature and sign) and adding a configuration option to enable the SUPERCOP/libOQS-compatible APIs. If we do that, however, we should move away from the `crypto_sign`-prefixes as that signals compatibility.

@hanno-becker, WDYT? Maybe @dstebila also has some opinion on this.

[...]

PR #959, opened 2026-02-10
Sign: Set smlen to 0 in case of failure

### LINKED ISSUE #974: Inconsistent and possibly confusing mld_sign() API and contracts
A pedantic reviewer points out:
```
File: mldsa/src/sign.c, in mld_sign:

c
ret = mld_sign_signature(sm, smlen, sm + MLDSA_CRYPTO_BYTES, mlen, ctx,
                         ctxlen, sk, context);
*smlen += mlen;
return ret;


If mld_sign_signature fails, it sets *siglen = 0. Then *smlen += mlen makes *smlen = mlen on failure. The CBMC contract says the function
should return *smlen == MLDSA_CRYPTO_BYTES + mlen on success, but doesn't constrain *smlen on failure. However, returning a non-zero
*smlen on failure could confuse callers into thinking there's valid data in the buffer. The mld_sign_open counterpart does set *mlen = 0
on failure.

Severity: Low (robustness). Consider setting *smlen = 0 on error for defensive consistency.
```

I tend to agree... we should set *smlen==0 on failure, and update the contracts and comments to match. Does this also affect the top-level API in mldsa_native.h?

mkannwischer:
- Resolves #974

### DIFF
--- a/mldsa/src/sign.c
+++ b/mldsa/src/sign.c
@@ -720,8 +720,10 @@ __contract__(
    * Consequently, any value that can be computed from the signature and public
    * key is considered public.
    * w0 and w1 are public as they can be computed from Az - ct = \alpha w1 + w0.
-   * h=c*t0 is public as both c and t0 are public.
-   * For a more detailed discussion, refer to https://eprint.iacr.org/2022/1406.
+   * h=c*t0 is public as both c and t0 are considered public.
+   * While t0 is not part of the public key, it can be reconstructed from
+   * a small number of signatures and need not be regarded as secret
+   * (see @[FIPS204, Section 6.1]).
    */
   MLD_CT_TESTING_DECLASSIFY(w0, sizeof(*w0));
   MLD_CT_TESTING_DECLASSIFY(w1, sizeof(*w1));
--- a/mldsa/src/sign.c
+++ b/mldsa/src/sign.c
@@ -1012,7 +1012,10 @@ int mld_sign(uint8_t *sm, size_t *smlen, const uint8_t *m, size_t mlen,
   }
   ret = mld_sign_signature(sm, smlen, sm + MLDSA_CRYPTO_BYTES, mlen, ctx,
                            ctxlen, sk, context);
-  *smlen += mlen;
+  if (ret == 0)
+  {
+    *smlen += mlen;
+  }
   return ret;
 }
 #endif /* !MLD_CONFIG_NO_RANDOMIZED_API */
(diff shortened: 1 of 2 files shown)


mldsa/src/sign.c
```diff
@@ -964,7 +966,10 @@ int mld_sign(uint8_t *sm, size_t *smlen, const uint8_t *m, size_t mlen,
   }
   ret = mld_sign_signature(sm, smlen, sm + MLDSA_CRYPTO_BYTES, mlen, ctx,
                            ctxlen, sk, context);
-  *smlen += mlen;
+  if (ret == 0)
```
rod-chapman, line 1017: Please update the comment(s) and the `ensures` contract for this function in sign.h
(or is it deliberately left under-specified?)

rod-chapman: Do the comments at the top-level API in mldsa/mldsa_native.h also need to be updated?

mkannwischer: I think the contract and comments exactly what we want to guarantee:
`(return_value == 0 && *smlen == MLDSA_CRYPTO_BYTES + mlen)`
WDYT @hanno-becker?


In the long term I think we should rather consider eliminating `smlen` altogether - see https://github.com/pq-code-package/mldsa-native/issues/789.

rod-chapman: In the failure case(s), should we write *smlen==0 in the ensures clause, or leave it unspecified?

hanno-becker: I'm in favor of harmonizing this with `mld_sign_signature`, and modify the post-condition from 
```c
  ensures((return_value == 0 && *smlen == MLDSA_CRYPTO_BYTES + mlen) ||
          (return_value == MLD_ERR_FAIL
           || return_value == MLD_ERR_OUT_OF_MEMORY
           || return_value == MLD_ERR_RNG_FAIL))
```
to
```c
  ensures((return_value == 0 && *smlen == MLDSA_CRYPTO_BYTES + mlen) ||
          ((return_value == MLD_ERR_FAIL || return_value == MLD_ERR_OUT_OF_MEMORY || return_value == MLD_ERR_RNG_FAIL) && *smlen == 0))
```

hanno-becker (changes requested):
I agree with @rod-chapman that we should note in the CBMC contract for `mld_sign` that `*smlen == 0` if the function fails.

mkannwischer:
> I agree with @rod-chapman that we should note in the CBMC contract for `mld_sign` that `*smlen == 0` if the function fails.

Okay. Adjusted.

hanno-becker (approved):

Issue #765, opened 2025-12-01
RFC: Split `sign.c` into sign/verify/keygen logic


hanno-becker:
Some consumers may only be interested in a subset of sign / verify / key generation. While we can consider adding configuration options for disabling them at compile time, that is only half-satisfactory for those consumers because they would either need to a) take the whole sign.c and accept dead code in it, b) manually simplify sign.c, throwing out the code they don't need.

It would be more convenient if sign/key and verify would be in separate compilation units, so that a consumer merely interested in verification could just drop sign.* from the import.

Happy about thoughts @mkannwischer

mkannwischer:
There are a lot of functions in polyvec.c/poly.c/poly_kl.c/packing.c that are only required for a subset of keypair/sign/verify.
So if you really want to split by compilation units, you will either have to split them as well or turn them into headers and make everything static. The former seems inconvenient for consumers. Do we foresee any larger issues with the latter?
In that case we would only have 3 C files: keypair.c/sign.c/verify.c (+the ct.c) - everything else would be headers.

PR #960, opened 2026-02-10
Add configs to disable unused APIs

### LINKED ISSUE #941: Config flag to only build for signature verification

flynd:
Add configs to disable key generation, signature creation, and/or signature verification to reduce the library size when not needing only one or two of these.
Also adds a config to disable all but the internal APIs, allowing for example to build only crypto_sign_verify_internal() and exclude everything else.

Resolves #941

### DIFF
--- a/dev/aarch64_clean/src/polyz_unpack_table.c
+++ b/dev/aarch64_clean/src/polyz_unpack_table.c
@@ -27,7 +27,8 @@ MLD_ALIGN const uint8_t mld_polyz_unpack_17_indices[] = {
     2,  3,  4,  255, 4,  5,  6,  255, 6,  7,  8,  255, 8,  9,  10, 255,
     11, 12, 13, 255, 13, 14, 15, 255, 15, 16, 17, 255, 17, 18, 19, 255,
 };
-#endif
+#endif /* MLD_CONFIG_MULTILEVEL_WITH_SHARED || MLD_CONFIG_PARAMETER_SET == 44 \
+        */
 
 #if defined(MLD_CONFIG_MULTILEVEL_WITH_SHARED) || \
     (MLD_CONFIG_PARAMETER_SET == 65 || MLD_CONFIG_PARAMETER_SET == 87)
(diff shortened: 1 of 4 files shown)


mkannwischer:
> > Thanks @flynd. We will need a way to test the configuration options in CI. Adding a new example based on the monobuild example seems to be the easiest path.
> 
> I started looking at this, but the standard examples create a signature and then verifies it. When building without signature creation the example code needs to do something else so I'm not sure what would be appropriate here.

We already have https://github.com/pq-code-package/mldsa-native/blob/main/examples/basic/expected_signatures.h. This could be extended also with keys, then you could implement keygen, sign, and verify separately.
Probably we do not want to touch existing examples as it would make them hard to understand, but for a new example, this should be fine.
I can take a look at some point, but this month is rather busy.

flynd:
> This PR adds quite a bit of additional pre-processor conditionals that make maintanance harder, but at the moment I don't see a better way to achieve the same, and I quite like the feature to disable APIs. I support this change, but hope that we can simplify things in follow-up PRs. For example, #765 would help a bit already, I believe. What do you think about that one @flynd?

I agree that splitting sign.c into separate files is probably a good idea, but as you've already pointed out there is a lot of code outside sign.c that is also specific to only some of the main operations so there would still remain a lot of ifdefs.


> What is left is to run full CI (which only works for PRs from local branches) - I have created #1000 for that . We'll also need approval by @hanno-becker.

I was wondering why you were always duplicating my PRs.  I have not worked much with github so thank you for explaining the reason behind this.

> @flynd thanks for your many contributions so far! Would you like me to add you as a contributor to mldsa-native/mlkem-native? Then you could create local branches and we save this extra step above. That would also make merging easier as we do not need to involve both @hanno-becker and me for simple changes.

I don't have anything else large after this one, but it isn't unlikely that I'll find something more minor that I want to contribute so @mkannwischer please add me so it can be done without the extra steps.

What I have locally right now is declaring all the tables as static to hide the linker symbols from other parts of the build.  I'm not sure if it is okay to use MLD_CONFIG_INTERNAL_API_QUALIFIER for things that aren't functions so I don't know if I should try to push that.
The other thing I have i an ugly hack in rej_uniform_asm.S to not use stack memory that I definitely shouldn't share. =)
However I do see some cases of multiple "ldr qXX" that could be shortened by using ldp or ld1 so I will probably do that.

flynd:
> Nice, I have added you in [pq-code-package/tsc#207](https://github.com/pq-code-package/tsc/pull/207) and you should have gotten an e-mail about it. This should also allow you to make changes to #1000. Feel free to do that.

Thank you @mkannwischer.  I was able to push an update for that PR, but the CI gets the same "nix: command not found" as when I pushed via my forked repo so something more seems to be needed.

PR #961, opened 2026-02-11
Armv8.1-M: Add CFI directives for stack unwinding 

mkannwischer:
Extend scripts/cfify with Armv8.1-M architecture support, handling
push/pop, vpush/vpop, sub/add sp, and bx lr.
Enable cfify for armv81m in scripts/autogen.

Also removes pushing/poping r3 and r12 in keccak_f1600_x4_mve_asm - they
do not have to be preserved according ot the AAPCS.

- Ports https://github.com/pq-code-package/mlkem-native/pull/1558
- Depends on #911


mkannwischer:
Superseded by #1048

PR #963, opened 2026-02-11
Autogen: Auto-generate polyz_unpack TBL index tables from autogen

mkannwischer:
Move the hand-written polyz_unpack_{17,19}_indices tables to be auto-generated from a Python generator in scripts/autogen.

### DIFF
--- a/scripts/autogen
+++ b/scripts/autogen
@@ -1069,6 +1069,65 @@ def gen_aarch64_rej_uniform_table():
     update_file("dev/aarch64_clean/src/rej_uniform_table.c", "\n".join(gen()))
 
 
+def gen_aarch64_polyz_unpack_indices(bit_width):
+    """Generate Neon TBL index bytes for polyz_unpack.
+
+    Each loop iteration loads bit_width * 16 / 8 packed bytes into
+    registers v0, v1, v2 and unpacks 16 coefficients. Coefficients
+    are unpacked in 4 groups of 4, one group per TBL/TBL2.
+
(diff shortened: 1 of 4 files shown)


scripts/autogen
```diff
@@ -1069,6 +1069,64 @@ def gen_aarch64_rej_uniform_table():
...
 
+def gen_aarch64_polyz_unpack_indices(bit_width):
+    """Generate Neon TBL index bytes for polyz_unpack.
+
+    Each loop iteration loads bit_width * 16 / 8 packed bytes into
+    registers v0, v1, v2 and unpacks 16 coefficients (4 per TBL/TBL2).
+    Each coefficient is extracted from 3 consecutive bytes.
+
+    Groups 0,1 index from v0 (and v1); groups 2,3 index from v1 (and v2),
+    so their byte offsets are shifted by -16.
```
hanno-becker, line 1081: Nit: I stumbled over the first use of "group" initially
```suggestion
    Each loop iteration loads bit_width * 16 / 8 packed bytes into
    registers v0, v1, v2 and unpacks 16 coefficients. Coefficients
    are unpacked in 4 groups of 4, one group per TBL/TBL2.
    Each coefficient is extracted from 3 consecutive bytes.

    Groups 0,1 index from v0 (and v1); groups 2,3 index from v1 (and v2),
    so their byte offsets are shifted by -16.
```

mkannwischer: Agreed. Thx

Issue #969, opened 2026-02-14
Build failure for test_unit.c on target without native NTT


nibrunieAtSi5:
When trying to compile the version 749490e9 (`main` branch) on a target (RISC-V) with clang version  18.1.3 using:
```
 CC=clang make test
```

I get the following error:
```
  CC      test/build/mldsa44/test/src/test_unit.c.o
test/src/test_unit.c:143:13: error: unused function 'generate_i32_array_zeros' [-Werror,-Wunused-function]
  143 | static void generate_i32_array_zeros(int32_t *data, size_t len)
      |             ^~~~~~~~~~~~~~~~~~~~~~~~
test/src/test_unit.c:148:13: error: unused function 'generate_i32_array_single' [-Werror,-Wunused-function]
  148 | static void generate_i32_array_single(int32_t *data, size_t len, size_t pos,
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~
test/src/test_unit.c:160:13: error: unused function 'generate_i32_array_ranged' [-Werror,-Wunused-function]
  160 | static void generate_i32_array_ranged(int32_t *data, size_t len, int min_incl,
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~
test/src/test_unit.c:173:12: error: unused function 'compare_i32_arrays' [-Werror,-Wunused-function]
  173 | static int compare_i32_arrays(const int32_t *a, const int32_t *b, unsigned len,
      |            ^~~~~~~~~~~~~~~~~~
4 errors generated.
make: *** [test/mk/rules.mk:29: test/build/mldsa44/test/src/test_unit.c.o] Error 1
```

It seems that those functions are only used in some cases (including if the target has native NTT implementation) and the build will fail on other targets with some compiler versions because of the `-Werror,-Wunused-function]`.

PR #970, opened 2026-02-14
Guarding utility functions in test_unit.c with attribute unused.

nibrunieAtSi5:
Fixing #969

Some static inline utility functions from test_unit.c are only used for some target, in particular target for which a native NTT implementation exists. For other targets, the build may fail because the warning of unused functions get promoted to an error.

I am not sure this is the best way to fix this, as another way would be to move the static function definitions inside the guard of the macro 
```
#idef MLD_USE_NATIVE_NTT
```

### DIFF
--- a/test/src/test_unit.c
+++ b/test/src/test_unit.c
@@ -113,6 +113,7 @@ static int compare_u64_arrays(const uint64_t *a, const uint64_t *b,
     defined(MLD_USE_NATIVE_POLYZ_UNPACK_19) ||                      \
     defined(MLD_USE_FIPS202_X1_NATIVE) || defined(MLD_USE_FIPS202_X4_NATIVE)
 
+#ifdef MLD_USE_NATIVE_NTT
 /* Backend unit test helper functions */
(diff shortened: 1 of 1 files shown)


nibrunieAtSi5:
> Thank you for the report and fix @nibrunieAtSi5!
> 
> I think we should guard the functions by suitable `#ifdef MLDSA_USE_NATIVE_XXX`. Can you try that?

Certainly, I will update the PR.

hanno-becker:
It looks like @mkannwischer noticed this as well, and added a fix here: https://github.com/pq-code-package/mldsa-native/pull/911/changes#diff-e276c71b06f7f340d2e6d9bcb3574053ee6d37c9317d10b5f7f33112169afff4

nibrunieAtSi5:
> @nibrunieAtSi5 Can you say more about the context in which you bumped into this? Are you developing a RV backend for Keccak? If so, needless to say we'd be interested in having this upstream if possible 😄

Yeah, I am tickering with a RVV backend but nothing too serious for now, just playing with RISC-V Vector and the Keccak x4. I don't think it is worth upstreaming for now.

PR #973, opened 2026-02-21
x86_64/AArch64: Add AVX2/Neon polyw1_pack to x86_64 native backend

mkannwischer:
Integrate polyw1_pack AVX2 implementations for both GAMMA2 variants into the native backend.

# `polyw1_pack` component benchmarks

## Intel Xeon 8375C (c6i.metal, no Turbo Boost, no SMT)

| | ML-DSA-44 | ML-DSA-65 | ML-DSA-87 |
|---|---|---|---|
| C | 500 | 310 | 311 |
| AVX2 | 216 | 144 | 144 |
| *Speedup* | *2.3x* | *2.2x* | *2.2x* |

## Apple M1

| | ML-DSA-44 | ML-DSA-65 | ML-DSA-87 |
|---|---|---|---|
| C | 203 | 49 | 49 |
| AArch64 | 32 | 22 | 22 |
| *Speedup* | *6.3x* | *2.2x* | *2.2x* |
  
  
  TODO: 
 - [ ] Add CBMC proof
 - [ ] Add Unit tests
 - [X] Try AArch64 implementation
 - [ ] Create HOL-Light + conversion to asm issues


mkannwischer:
It seems worthwhile to still add this.

It currently seems that we should only adopt it for some parameter sets (or maybe change the C code so that the compiler can produce better code by itself. Should we decide to add additional assembly, we'll need HOL-Light proofs before merging. 

@jakemas, do you possibly have time to adopt this PR and evaluate it further? I won't get to it anytime soon. 3% faster signing on Graviton may be of interest to you.

mkannwischer:
> It seems worthwhile to still add this.
> 
> It currently seems that we should only adopt it for some parameter sets (or maybe change the C code so that the compiler can produce better code by itself. Should we decide to add additional assembly, we'll need HOL-Light proofs before merging.
> 
> @jakemas, do you possibly have time to adopt this PR and evaluate it further? I won't get to it anytime soon. 3% faster signing on Graviton may be of interest to you.

actually this conflicts with #1345. We should decide how to proceed with that one first.

PR #985, opened 2026-03-03
CBMC: Correct operator precedence in ensures clause of mld_caddq()

rod-chapman:
A pedantic reviewer points out that the ensures contract on mld_caddq() says:

`ensures(return_value == (a >= 0) ? a : (a + MLDSA_Q))`

but operator == binds more tightly than ternary ? : so this parses as:
`(return_value == (a >= 0)) ? a : (a + MLDSA_Q)`

Parentheses are needed here to ensure correctness and readability.

`ensures(return_value == ((a >= 0) ? a : (a + MLDSA_Q)))`

CBMC says "VERIFICATION SUCCESS" for both cases, but I prefer the latter for readability.

### DIFF
--- a/mldsa/src/reduce.h
+++ b/mldsa/src/reduce.h
@@ -129,7 +129,7 @@ __contract__(
   requires(a < MLDSA_Q)
   ensures(return_value >= 0)
   ensures(return_value < MLDSA_Q)
-  ensures(return_value == (a >= 0) ? a : (a + MLDSA_Q))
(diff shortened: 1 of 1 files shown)


tautschnig:
Just to confirm that CBMC indeed does get operator precedence right (per the C standard):
````c
int main()
{
  int a;
  __CPROVER_assert(a == a ? 1 : 0, "must pass");
}
````
passes with CBMC, both with and without `--no-simplify` (this option disables the internal expression rewrite engine, tested just to make sure we don't get it right in one place, but wrong in some other).

PR #942, opened 2026-02-03
Add HOL Light proof for aarch64 `poly_chknorm`

mkannwischer:
Add a HOL Light functional correctness proof for the aarch64 ML-DSA function poly_chknorm, which checks whether any polynomial coefficient has absolute value >= a given bound.

The correctness proof shows that the assembly computes  `bitval(?i. i < 256 /\ abs(ival(x i)) >= ival bound)`

- Resolves https://github.com/pq-code-package/mldsa-native/issues/506

TODO: 
- [x] Merge 3 missing instructions into s2n-bignum - I opened a PR [here](https://github.com/awslabs/s2n-bignum/pull/344)
- [x] CBMC spec and proof

PR #987, opened 2026-03-03
Correct the contract of mld_poly_chknorm_native()

rod-chapman:
Correct the contract of mld_poly_chknorm_native() in native/api.h to reflect that it can return -1, 0 or +1.
The earlier contract was too strong and excluded the +1 return value.

Update comments accordingly, and proof of calling unit mld_poly_chknorm() to avoid CBMC proof failure on
unsigned overflow when the returned value is +1 and is negated to get -1.

autogen OK
Lint OK
Tests all OK

### DIFF
--- a/mldsa/src/native/api.h
+++ b/mldsa/src/native/api.h
@@ -39,13 +39,13 @@
 
 /* Bound on absolute value of coefficients after NTT.
  *
- * NOTE: This is the same bound as in ntt.h and has to be kept
+ * NOTE: This is the same bound as in poly.h and has to be kept
  * in sync. */
 #define MLD_NTT_BOUND (9 * MLDSA_Q)
 
 /* Absolute exclusive upper bound for the output of the inverse NTT
  *
- * NOTE: This is the same bound as in ntt.h and has to be kept
+ * NOTE: This is the same bound as in poly.h and has to be kept
  * in sync. */
 #define MLD_INTT_BOUND MLDSA_Q
 
@@ -54,7 +54,7 @@
  * NOTE: This is the same bound as in reduce.h and has to be kept
  * in sync. */
 /* check-magic: 6283009 == (MLD_REDUCE32_DOMAIN_MAX - 255 * MLDSA_Q + 1) */
-#define REDUCE32_RANGE_MAX 6283009
+#define MLD_REDUCE32_RANGE_MAX 6283009
 /*
  * This is the C<->native interface allowing for the drop-in of
  * native code for performance critical arithmetic components of ML-DSA.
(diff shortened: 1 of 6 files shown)


mldsa/src/native/api.h
```diff
@@ -429,19 +429,27 @@ __contract__(
...
 static MLD_INLINE int mld_poly_chknorm_native(const int32_t *a, int32_t B)
 __contract__(
   requires(memory_no_alias(a, sizeof(int32_t) * MLDSA_N))
   requires(0 <= B && B <= MLDSA_Q - REDUCE32_RANGE_MAX)
   requires(array_bound(a, 0, MLDSA_N, -REDUCE32_RANGE_MAX, REDUCE32_RANGE_MAX))
-  ensures(return_value == MLD_NATIVE_FUNC_FALLBACK || return_value == MLD_NATIVE_FUNC_SUCCESS)
-  ensures((return_value == 0) == array_abs_bound(a, 0, MLDSA_N, B))
+  ensures(return_value == MLD_NATIVE_FUNC_FALLBACK ||
+          return_value == MLD_NATIVE_FUNC_SUCCESS  ||
+          return_value == 1)
```
hanno-becker: It's a bit confusing to use `MLD_NATIVE_FUNC_SUCCESS` here. I think we should reserve that for functions which only return `MLD_NATIVE_FUNC_FALLBACK` or `MLD_NATIVE_FUNC_SUCCESS`. Here, I prefer @mkannwischer's formulation from https://github.com/pq-code-package/mldsa-native/pull/942/changes/627b7343ebb438d246459fe8b48dc7c84ade3b36:

```c
ensures(return_value == MLD_NATIVE_FUNC_FALLBACK || return_value == 0 || return_value == 1)
```

As for the second post-condition, I think this actually wrong (and also in https://github.com/pq-code-package/mldsa-native/pull/942/changes/627b7343ebb438d246459fe8b48dc7c84ade3b36). We only want to constrain the `return_value` when the function is supported. So, we would rather need

```c
return_value != MLD_NATIVE_FUNC_FALLBACK ==> ((return_value == 0) == array_abs_bound(a, 0, MLDSA_N, B))
```

rod-chapman: Done

mldsa/src/native/api.h
```diff
@@ -429,19 +429,28 @@ __contract__(
  *              Assumes input coefficients were reduced by mld_reduce32().
  *
  * Arguments:   - const int32_t *a: pointer to polynomial
- *              - int32_t B: norm bound
+ *              - int32_t B: norm bound, which must be in the range
+ *                0 .. MLDSA_Q - MLD_REDUCE32_RANGE_MAX inclusive.
```
hanno-becker, line 433: While at it, could you rename this to `MLD_REDUCE32_RANGE_MAX`?

hanno-becker: We seem to have this twice, once with `MLD_` and once without.

hanno-becker: It's also not clear why this isn't flagged -- don't we have a macro checker forcing `MLD_` prefix? Can you check that, too, @rod-chapman ?

rod-chapman: Oh yes... it's defined twice.
In reduce.h, we have 
```
#define MLD_REDUCE32_RANGE_MAX 6283009
```
and in native/api.h, we have
```
#define REDUCE32_RANGE_MAX 6283009
```
Was there an intent here to avoid creating a dependency between the back- and front-ends?

rod-chapman: There are several other constants (e.g. MLD_NTT_BOUND) that are declared twice - once in native/api.h and once in reduce.h. For now I will normalize the naming of REDUCE32_RANGE_MAX to have the MLD_ prefix.

rod-chapman: I have pushed that renaming.

There is some checking of macro names in scripts/autogen (line 1929 .. 1993), but I cannot understand the intent of this code.

mkannwischer (approved):
Thanks @rod-chapman. Probably a good idea to pull this out of https://github.com/pq-code-package/mldsa-native/pull/942 as the s2n-bignum reviews are facing delays.

PR #989, opened 2026-03-04
HOL-Light/x86_64: Replace Keccakx4 intrinsics with AVX2 assembly and prove correct

mkannwischer:
- Ports https://github.com/pq-code-package/mlkem-native/pull/1576


dev/fips202/x86_64/src/keccak_f1600_x4_avx2.S
```diff
@@ -0,0 +1,622 @@
+/*
+ * Copyright (c) The mldsa-native project authors
```
mkannwischer, line 3: this needs the mlkem-native copyright too.

mkannwischer: fixed.

hanno-becker (approved):
I checked that this is a faithful port of https://github.com/pq-code-package/mlkem-native/pull/1576.

Sadly, the performance is getting much worse on some platforms.

PR #996, opened 2026-03-16
CBMC: Add termination proofs for all functions except rejection sampling and poly_challenge

mkannwischer:
- Ports https://github.com/pq-code-package/mlkem-native/pull/1625

Functions currently not covered:
 - ("mldsa/src/poly.c", "mld_poly_uniform"),
 - ("mldsa/src/poly.c", "mld_poly_uniform_4x"),
 - ("mldsa/src/poly_kl.c", "mld_poly_uniform_eta_4x"),
 - ("mldsa/src/poly_kl.c", "mld_poly_uniform_eta"),
 - ("mldsa/src/poly_kl.c", "mld_poly_challenge"),
 - ~("mldsa/src/packing.c", "mld_unpack_hints"),~

For the first 5 functions, we can't prove termination as it's probabilistic. ~For the last, I believe it should be possible to prove termination, but currently the proof does not work.~ I have a proof for `mld_unpack_hints` now by adding an additional invariant.

### DIFF
--- a/scripts/check-contracts
+++ b/scripts/check-contracts
@@ -101,13 +101,14 @@ def check_contracts():
 
 # Loops that only terminate probabilistically (rejection sampling)
 # and therefore cannot have a decreases clause.
-DECREASES_EXCEPTIONS = [
-    ("mldsa/src/poly.c", "mld_poly_uniform"),
-    ("mldsa/src/poly.c", "mld_poly_uniform_4x"),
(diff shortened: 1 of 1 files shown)


hanno-becker (commented):
In principle this looks good, but the performance regression in `polymat_permute_bitrev_to_custom` is severe. We should take a look if some tweaking of flags can prevent this.

Issue #760, opened 2025-12-01
SLOTHY: Apply to `polyz_unpack`


mkannwischer:

mkannwischer:
It's my understanding that @GoulongWang is giving this a try.

PR #999, opened 2026-03-20
AArch64: Fix buffer overread in aarch64 `polyz_unpack_{17,19}`

mkannwischer:
The 3-register ld1 {v0, v1, v2} loads 48 bytes but only 36 (polyz_unpack_17) or 40 (polyz_unpack_19) are needed, causing a 12/8-byte overread past the input buffer on the last iteration. Split into a 2-register ld1 for the first 32 bytes plus a single-element ld1 for the remainder.


dev/aarch64_clean/src/polyz_unpack_17_asm.S
```diff
@@ -65,8 +65,10 @@ MLD_ASM_FN_SYMBOL(polyz_unpack_17_asm)
         mov count, #(64/4)
 
 polyz_unpack_17_loop:
-        ld1	{v0.16b, v1.16b, v2.16b}, [buf]
-        add	buf, buf, #0x24
+        ld1	{v0.16b, v1.16b}, [buf]
```
hanno-becker, line 72: Suggestion: The clean version could do with a comment explaining why it's done this way.

dev/aarch64_clean/src/polyz_unpack_19_asm.S
```diff
@@ -62,8 +62,10 @@ MLD_ASM_FN_SYMBOL(polyz_unpack_19_asm)
         mov count, #(64/4)
 
 polyz_unpack_19_loop:
-        ld1	{v0.16b, v1.16b, v2.16b}, [buf]
-        add	buf, buf, #0x28
+        ld1	{v0.16b, v1.16b}, [buf]
+        add	buf, buf, #0x18
+        ld1	{v2.16b}, [buf], #0x10
```
hanno-becker, line 71: * Is it worth using two `buf` counters so there is no data dependency between the first and second load?
* Can we avoid the `mov` by just tweaking the unpack table? It's already auto-generated, so it should be easy to adjust the indices that target `v2`.

dev/aarch64_clean/src/polyz_unpack_17_asm.S
mkannwischer: Yes. Good point. added.

dev/aarch64_clean/src/polyz_unpack_19_asm.S
mkannwischer: Neat idea to adjust the indices. I have implemented that and it saves a couple of cycles on Graviton 2. 

I have tried using two pointers, but it did not improve performance on Graviton 2 (it's actually a few cycles slower). We may want to revisit that when tackling https://github.com/pq-code-package/mldsa-native/issues/760 as it will give SLOTHY a bit more freedom.

hanno-becker: I'd prefer to switch to two separate pointers here, so there is no interference between #760 and #921. How much slower was it with two pointers? Is it noise or real?

mkannwischer: Old: 189
New w/ 2 pointers: 184
New w/ 1 pointer: 181

hanno-becker: (Reflecting direct conversation) Symbolic execution in HOL Light should not be sensitive anyway to whether we use one or two pointers; so I think @mkannwischer you're right we can leave it for now.

PR #1000, opened 2026-03-21
Add configs to disable unused APIs [full CI]

mkannwischer:
Continuation of #960 by @flynd to run full CI.


dev/fips202/aarch64/auto.h
```diff
@@ -37,6 +37,7 @@
 #include "x1_scalar.h"
 #endif
 
+#if !defined(MLD_CONFIG_SERIAL_FIPS202_ONLY) && !defined(MLD_CONFIG_REDUCE_RAM)
```
hanno-becker: The commit message says that if `MLD_CONFIG_REDUCE_RAM` is set, we don't use batched Keccak. Is that really true? E.g. `mld_poly_uniform_gamma1_4x` is used even if `MLD_CONFIG_REDUCE_RAM` is set.

Generally, I'm not sure guarding the backend metadata is the right approach.

examples/disabled_apis_native/Makefile
```diff
@@ -0,0 +1,118 @@
...
+RNG_SOURCE=$(wildcard test_only_rng/*.c)
+ALL_SOURCE=$(MLD_SOURCE) $(RNG_SOURCE) main.c
+
+BUILD_DIR=build
+
+CFLAGS += -DMLD_CONFIG_NAMESPACE_PREFIX=mldsa
+CFLAGS += -DMLD_CONFIG_USE_NATIVE_BACKEND_ARITH
+CFLAGS += -DMLD_CONFIG_USE_NATIVE_BACKEND_FIPS202
+
+# Disabled API flags
```
hanno-becker: Why don't we use auto-generated configs here?

dev/fips202/aarch64/auto.h
flynd: > The commit message says that if `MLD_CONFIG_REDUCE_RAM` is set, we don't use batched Keccak. Is that really true? E.g. `mld_poly_uniform_gamma1_4x` is used even if `MLD_CONFIG_REDUCE_RAM` is set.
> 
> Generally, I'm not sure guarding the backend metadata is the right approach.

Following the call chains, I find that shake128x4 is only used in mld_poly_uniform_4x(), which is only called from mld_polyvec_matrix_expand() when MLD_CONFIG_REDUCE_RAM is not set.

However shake256x4 is only used in mld_poly_uniform_eta_4x() which is called from mld_sample_s1_s2() which is used in mld_sign_keypair_internal(), and mld_poly_uniform_gamma1_4x() which is called from mld_polyvecl_uniform_gamma1() which is used in mld_attempt_signature_generation().

So @hanno-becker is correct that MLD_CONFIG_REDUCE_RAM isn't the correct condition here.  I had already disabled key generation and signature creation in my build tree when I did this change and it caused me to assume the wrong condition.
When combined with the other commits in this stack, the correct condition should be:
#if (!defined(MLD_CONFIG_NO_KEYPAIR_API) || !defined(MLD_CONFIG_NO_SIGN_API) || !defined(MLD_CONFIG_REDUCE_RAM)) && !defined(MLD_CONFIG_SERIAL_FIPS202_ONLY)

flynd: I have pushed an update to fix the conditions in auto.h.

examples/disabled_apis_native/Makefile
flynd: I'm not sure if this question was directed at me.  The disabled_apis and disabled_apis_native tests were added as a separate commit by @mkannwischer.

flynd: @mkannwischer : This comment is for the tests you created so I don't know what to answer.  Could you please answer this?

mkannwischer: Apologies for the delay. I have added autogeneration for the configs now. I'm not sure if that adds a lot of value though.

examples/multilevel_build_native/mldsa_native/mldsa_native_config.h
```diff
@@ -170,6 +170,16 @@
...
 
+/******************************************************************************
+ * Name:        MLD_CONFIG_INTERNAL_API_ONLY
+ *
+ * Description: Set this to remove all public APIs except
+ *              crypto_sign_keypair_internal, crypto_sign_signature_internal,
+ *              and crypto_sign_verify_internal.
+ *
+ *****************************************************************************/
+/* #define MLD_CONFIG_INTERNAL_API_ONLY */
```
hanno-becker: The name of this option is confusing because it overlaps with the use of "internal API" for internal functions of external linkage.

Suggestion: `MLD_CONFIG_CORE_API_ONLY`

examples/disabled_apis/main.c
```diff
@@ -0,0 +1,91 @@
+/*
```
hanno-becker: Hmm... why is this a custom example rather than a custom config?

Right now, if I take out-of-the-box mldsa-native and just set `MLD_CONFIG_NO_SIGN_API`, a plain `make` fails.

Also, with the current approach we don't exercise a monobuild with reduced config, which would be useful to catch unused functions.

I would prefer custom configurations rather than a custom example here, and adjusting all existing tests and examples to only run those tests which make sense in the chosen configuration. This means `make test` works out of the box for any valid config, plus we extend coverage to monobuild and multilevel build.

This is currently a blocker to me.

hanno-becker (changes requested):
Apologies for the long silence, and thank you very much for all the work on this, @flynd @mkannwischer!

I support the change in principle, but the build + test story needs more work: 
* A plain `make test` should work with any valid config -- right now, it doesn't. The existing tests and examples need adjusting to only exercise the API that is enabled in the config.
* Testing the reduced configurations through examples introduces test gaps; e.g., we don't exercise them in monobuilds.

ISTM that we should test this through config variations rather than new examples; this would force/solve both issues above.

examples/multilevel_build_native/mldsa_native/mldsa_native_config.h
flynd: I have renamed it as suggested.

flynd:
> I support the change in principle, but the build + test story needs more work:
> 
> * A plain `make test` should work with any valid config -- right now, it doesn't. The existing tests and examples need adjusting to only exercise the API that is enabled in the config.
> * Testing the reduced configurations through examples introduces test gaps; e.g., we don't exercise them in monobuilds.
> 
> ISTM that we should test this through config variations rather than new examples; this would force/solve both issues above.

I had a look at the tests and this looks like a big change since the pattern of creating key, signing, and then verifying the signature will no longer work.  As I'm not at all familiar with these tests, I don't know how to contribute such a change.  Do either of you have the time to help out with this?  If not, could you give some clear hints on how the tests should be modified and maybe I can try to figure it out.

I did notice though that there were two commits in the stack that could be merged separately so they are now moved to PRs #1029 and #1032.

mkannwischer:
> > I support the change in principle, but the build + test story needs more work:
> > 
> > * A plain `make test` should work with any valid config -- right now, it doesn't. The existing tests and examples need adjusting to only exercise the API that is enabled in the config.
> > * Testing the reduced configurations through examples introduces test gaps; e.g., we don't exercise them in monobuilds.
> > 
> > ISTM that we should test this through config variations rather than new examples; this would force/solve both issues above.
> 
> I had a look at the tests and this looks like a big change since the pattern of creating key, signing, and then verifying the signature will no longer work. As I'm not at all familiar with these tests, I don't know how to contribute such a change. Do either of you have the time to help out with this? If not, could you give some clear hints on how the tests should be modified and maybe I can try to figure it out.

I agree that this would be a major change. Every test and example would have to be changed to use known keys and signatures (and then compare the computed values against the known ones). ACVP would have to be entirely reworked because only parts of the tests would be executed. The KAT tests would be impossible.

@hanno-becker, would it be a compromise to get the monobuild tests to work with the new flags, but skip the tests that would require major rework for now and keep the examples instead? In a follow up we can address the skipped tests and eventually remove the examples if they are no longer needed. I think that would be more manageable, address the concern that `make test` and the monobuilds don't work with the new flags.

Unfortunately, I won't have time to help out beyond reviews in the next few weeks.

mkannwischer:
> ## CBMC Results (ML-DSA-87)
> ⚠️ **Attention Required**
> 
> Proof	Status	Current	Previous	Change
> `polyvecl_pointwise_acc_montgomery_c`	❌	-	218s	-
> `polymat_permute_bitrev_to_custom`	⚠️	153s	27s	+467%
> Full Results (184 proofs)

This failure is likely unrelated to the changes. It's flaky also in other branches. Going to fix it.

hanno-becker:
I will take a look and get back to you.

hanno-becker:
@flynd @mkannwischer I am having a stab at adjusting the tests. Needs a bit more care for CT but I hope to have it done by the end of the weekend.

hanno-becker:
@flynd @mkannwischer I had a stab at adjusting the tests. Please have a look if you agree with the approach.

flynd:
> @flynd @mkannwischer I had a first stab at adjusting the tests. Please have a look if you agree with the approach. I still need to go over it in detail myself, though, so there may still be some low-hanging bugs in it.

Thank you.
I noticed you had fixed a few of the conditions that I had gotten wrong.  I have pushed an update where my original commits are fixed instead.

scripts/autogen
```diff
@@ -3421,6 +3644,24 @@ def _main():
         default="att",
         help="Assembly syntax for x86_64 disassembly output (att or intel)",
     )
+    parser.add_argument(
+        "--test-vectors",
```
mkannwischer, line 3647: Should this be exercised in CI?

scripts/autogen
```diff
@@ -3382,6 +3382,229 @@ def gen_test_configs():
         gen_test_config(config_spec["path"], config_spec, default_config)
 
 
+def gen_test_vectors(msg, ctx):
+    """Generate expected test vectors using ACVP binaries and notrandombytes.
+
+    For each parameter set (44, 65, 87):
+    - Keygen: reset notrandombytes with seed=0, draw 32 bytes as keypair seed
+    - Signing: reset notrandombytes with seed=1, draw 32 bytes as sign rnd
+    This makes the signature independent of keygen PRNG consumption (e.g. PCT).
```
mkannwischer: I can't follow here what seed is used for. Can you elaborate?

scripts/autogen
```diff
@@ -3382,6 +3382,229 @@ def gen_test_configs():
...
+            raise RuntimeError(f"notrandombytes failed: {result.stderr}")
+        return result.stdout.strip().upper()
+
+    def run_acvp(lvl, *args):
+        acvp_bin = f"./test/build/mldsa{lvl}/bin/acvp_mldsa{lvl}"
+        if not os.path.isfile(acvp_bin):
+            console.print(
+                f"[red]Error:[/] ACVP binary not found: {acvp_bin}\n"
+                f"Build it first: [bold]make acvp OPT=0[/]"
+            )
```
mkannwischer, line 3410: This is the first `autogen` operation that assumes that `make` has already been called to produce the binaries. It would be better if `autogen` would call it internally.

META.sh
```diff
@@ -33,15 +35,30 @@ if (which yq >/dev/null 2>&1); then
...
-    echo "$META $1 $2: FAIL ($VAL != $INPUT)"
-    exit 1
-  else
-    echo "$META $1 $2: OK"
+# Determine the input to compare against
+if [[ $3 != "" ]]; then
+  INPUT=$3
+else
+  STDIN=$(cat)
+  if [[ $STDIN == SKIPPED* ]]; then
```
mkannwischer, line 43: Isn't it more robust to signal this trough a return code rather than matching strings here?

scripts/autogen
```diff
@@ -3382,6 +3382,229 @@ def gen_test_configs():
...
+        yield ""
+        yield from gen_msg_ctx_defines()
+        yield ""
+
+        first = True
+        for lvl in levels:
+            r = results[lvl]
+            directive = "#if" if first else "#elif"
+            yield f"{directive} MLD_CONFIG_PARAMETER_SET == {lvl}"
+            yield f"const uint8_t test_vector_pk[{hex_len(r['pk'])}] = {{"
```
mkannwischer: Should these symbols be `static`?

mkannwischer (commented):
Thanks @hanno-becker for adjusting the tests. Looks great to me! Here are a couple of nits. WDYT?

META.sh
hanno-becker: I did this originally, but it's difficult to not end up with a messy Makefile recipe. I find the current solution cleaner.

scripts/autogen
hanno-becker: I prefer not to have `autogen` silently build the library; I think it's acceptable to get a clean error and ask the user to build.

scripts/autogen
hanno-becker: Oh, this may be a remnant from an earlier version; I'll check.

META.sh
mkannwischer: Okay.

scripts/autogen
hanno-becker: Ok 👍

scripts/autogen
hanno-becker: Yes

hanno-becker (approved):

scripts/autogen
hanno-becker: Fixed

hanno-becker:
@mkannwischer @flynd Ready to go from my perspective.

hanno-becker:
Thank you very much for kicking this off, @flynd !

PR #1002, opened 2026-03-22
Sign: Unpack `s1`, `s2`, and `t0` on the fly in `REDUCE_RAM` mode

mkannwischer:
Introduce `mld_s1vec`, `mld_s2vec`, and `mld_t0vec`, following the same pattern as `mld_polymat` for reduced RAM usage. In normal mode, they store the full NTT'd polyvec. In `REDUCE_RAM` mode, they store a pointer to the packed data in the secret key and unpack + NTT individual polynomials on demand.

This reduces signing memory allocation in `REDUCE_RAM` mode:
- ML-DSA-44: 32,448 -> 20,256 (-37.6%)
- ML-DSA-65: 44,768 -> 27,456 (-38.6%)
- ML-DSA-87: 59,104 -> 35,648 (-39.7%)

TODO:
- [x] CBMC proofs for non-reduced-RAM mode
- [x] Avoid double unpacking of s1/s2/t0 in `pk_from_sk`


mldsa/src/polyvec.h
```diff
@@ -151,6 +154,36 @@ typedef struct
...
+typedef struct
+{
+#if defined(MLD_CONFIG_REDUCE_RAM)
+  const uint8_t *packed;
+#else
+  mld_polyveck vec;
+#endif
+} mld_s2vec;
+
+/* Secret vector t0, see mld_t0vec_init and mld_t0vec_get_poly below */
```
mkannwischer: copy&pasta. this is actually public.

mldsa/src/sign.c
```diff
@@ -1450,20 +1465,35 @@ int mld_sign_pk_from_sk(uint8_t pk[MLDSA_CRYPTO_PUBLICKEYBYTES],
...
+   * TODO: avoid this double unpacking */
+  mld_polyveck_unpack_eta(s2_raw, sk + 2 * MLDSA_SEEDBYTES + MLDSA_TRBYTES +
+                                      MLDSA_L * MLDSA_POLYETA_PACKEDBYTES);
+
+  /* Unpack t0 again in raw form for validation.
+   * TODO: avoid this double unpacking */
+  mld_polyveck_unpack_t0(t0_raw, sk + 2 * MLDSA_SEEDBYTES + MLDSA_TRBYTES +
+                                     MLDSA_L * MLDSA_POLYETA_PACKEDBYTES +
+                                     MLDSA_K * MLDSA_POLYETA_PACKEDBYTES);
+
```
mkannwischer: pk_from_sk gets quite a bit more ugly here, because we need to do bounds check between unpacking and NTTing.
Can't think about a good way right now. Ideas?

mldsa/src/polyvec.h
mkannwischer: Fixed.

mldsa/src/packing.h
```diff
@@ -165,26 +165,26 @@ __contract__(
 MLD_INTERNAL_API
 void mld_unpack_sk(uint8_t rho[MLDSA_SEEDBYTES], uint8_t tr[MLDSA_TRBYTES],
                    uint8_t key[MLDSA_SEEDBYTES], mld_polyveck *t0,
-                   mld_polyvecl *s1, mld_polyveck *s2,
+                   mld_s1vec *s1, mld_polyveck *s2,
```
hanno-becker: It should be documented that `s1` borrows `sk` here in the case of lowram. This is going to be difficult to capture with CBMC...

mldsa/src/polyvec.h
```diff
@@ -576,6 +587,30 @@ __contract__(
     array_bound(p->vec[k1].coeffs, 0, MLDSA_N, MLD_POLYETA_UNPACK_LOWER_BOUND, MLDSA_ETA + 1)))
 );
 
+#define mld_s1vec_init MLD_NAMESPACE_KL(s1vec_init)
+static MLD_INLINE void mld_s1vec_init(
+    mld_s1vec *s1, const uint8_t packed_s1[MLDSA_L * MLDSA_POLYETA_PACKEDBYTES])
```
hanno-becker: As above, it needs to be documented that this borrows `packages_s1`.

mldsa/src/polyvec.h
```diff
@@ -151,6 +152,16 @@ typedef struct
...
+/* Secret vector s1, see mld_s1vec_init and mld_s1vec_get_poly below */
+typedef struct
+{
+#if defined(MLD_CONFIG_REDUCE_RAM)
+  const uint8_t *packed;
+#else
+  mld_polyvecl vec;
+#endif
+} mld_s1vec;
+
```
hanno-becker: As we scale this pattern, I wonder if it is worth separating the two implementations more cleanly.

For example, you could have two entirely separate types `mld_s1vec_lazy` and `mld_s1vec_eager` with `init` and `get_poly` functions on them, and then just `typedef` / `#define` to one or the other depending on whether `REDUCE_RAM` is set. This makes the individual implementations easier to review, and may also allow the addition of unit tests which compile _both_ implementations and check that their behavior is equivalent.

mldsa/src/sign.c
```diff
@@ -648,10 +714,6 @@ __contract__(
 
   mld_H(challenge_bytes, MLDSA_CTILDEBYTES, mu, MLDSA_CRHBYTES, sig,
         MLDSA_K * MLDSA_POLYW1_PACKEDBYTES, NULL, 0);
-  /* Constant time: Leaking challenge_bytes does not reveal any information
-   * about the secret key as H() is modelled as random oracle.
-   * This also applies to challenges for rejected signatures.
-   * See Section 5.5 of @[Round3_Spec]. */
```
hanno-becker, line 647: Why was this removed? Doesn't seem related to the commit.

mkannwischer: That was removed by accident and was reintroduced in a later commit.

mldsa/src/packing.h
mkannwischer: added

mldsa/src/polyvec.h
mkannwischer: added

mldsa/src/sign.c
mkannwischer: moved now.

mldsa/src/polyvec.h
mkannwischer: Good idea. I have introduced a new header called polyvec_lazy.h that implements that.

mldsa/src/sign.c
```diff
@@ -1433,15 +1525,18 @@ int mld_sign_pk_from_sk(uint8_t pk[MLDSA_CRYPTO_PUBLICKEYBYTES],
   MLD_ALLOC(tr, uint8_t, MLDSA_TRBYTES, context);
   MLD_ALLOC(tr_computed, uint8_t, MLDSA_TRBYTES, context);
   MLD_ALLOC(key, uint8_t, MLDSA_SEEDBYTES, context);
-  MLD_ALLOC(s1, mld_polyvecl, 1, context);
-  MLD_ALLOC(s2, mld_polyveck, 1, context);
-  MLD_ALLOC(t0, mld_polyveck, 1, context);
+  MLD_ALLOC(s1, mld_s1vec, 1, context);
+  MLD_ALLOC(s1_raw, mld_polyvecl, 1, context);
```
hanno-becker: Nit on naming: If I hear `_raw` I think of raw bytes. Let's find a better name

mldsa/src/sign.c
```diff
@@ -1450,20 +1545,35 @@ int mld_sign_pk_from_sk(uint8_t pk[MLDSA_CRYPTO_PUBLICKEYBYTES],
   /* Unpack secret key */
   mld_unpack_sk(rho, tr, key, t0, s1, s2, sk);
```
hanno-becker: It is confusing that we compute `s1`, `s2` here but they are never used.

mkannwischer: yes, sorry, `mld_compute_t0_t1_tr_from_sk_components` should take `s1` and `s2`. Will change.

mkannwischer: actually no, we should just not use `unpack_sk` there and inline everything instead.

mldsa/src/sign.c
mkannwischer: fixed

mldsa/src/sign.c
```diff
@@ -450,8 +454,10 @@ __contract__(
  *
  * Arguments:   - uint8_t *sig: output signature
  *              - const mld_poly *cp: challenge polynomial
- *              - const polyvecl *s1: secret vector s1
+ *              - const mld_sk_s1hat *s1: secret vector s1
```
hanno-becker: Nit: 
```suggestion
 *              - const mld_sk_s1hat *s1hat: secret vector s1 in NTT domain
```
etc

mldsa/src/sign.c
```diff
@@ -464,20 +470,22 @@ __contract__(
  **************************************************/
 MLD_MUST_CHECK_RETURN_VALUE
 static int mld_compute_pack_z(uint8_t sig[MLDSA_CRYPTO_BYTES],
-                              const mld_poly *cp, const mld_polyvecl *s1,
-                              const mld_polyvecl *y, mld_poly *z)
+                              const mld_poly *cp, const mld_sk_s1hat *s1,
+                              const mld_polyvecl *y, mld_poly *z, mld_poly *tmp)
```
hanno-becker, line 474: In the eager case, passing `mld_poly` by pointer likely leads to a copy when calling `mld_sk_s1hat_get_poly`? I assume it's done this way to avoid having to allocate in `mld_compute_pack_z`?

mldsa/src/sign.c
```diff
@@ -531,6 +542,90 @@ __contract__(
...
 
+/*************************************************
+ * Name:        mld_polyveck_pointwise_poly_montgomery_s2
+ *
+ * Description: Pointwise multiplication of s2 by a polynomial in NTT domain
+ *              and multiplication of the resulting vector by 2^{-32}.
+ *
+ * Arguments:   - mld_polyveck *h: pointer to output vector
+ *              - const mld_poly *cp: pointer to input polynomial
+ *              - const mld_sk_s2hat *s2: pointer to input s2 vector
```
hanno-becker: Similar comment as above: Let's adjust the name of the parameter

mldsa/src/sign.c
mkannwischer: Yes, saving allocations is the goal here.
We do not see a performance penalty in the eager case, so this appears to be fine.

hanno-becker: Let's leave it for now.

test/src/test_alloc.c
```diff
@@ -71,7 +76,7 @@ typedef struct
 
 #define MLD_MAX_IN_FLIGHT_ALLOCS 100
 #ifndef MLD_BUMP_ALLOC_SIZE
-#define MLD_BUMP_ALLOC_SIZE (128 * 1024) /* 128KB buffer */
+#define MLD_BUMP_ALLOC_SIZE (192 * 1024) /* 192KB buffer */
```
hanno-becker: Why was this change made?

hanno-becker (approved):
Many thanks @mkannwischer 🙏 Please can you do the renaming of parameters still? Then it's good to go.

test/src/test_alloc.c
mkannwischer: An earlier version of this PR had increased the stack usage of `pk_from_sk` to above 128 KiB which is why I had to increase it. However, that has been changed now, so this change should be reverted.
Good catch!

mldsa/src/sign.c

mldsa/src/sign.c

test/src/test_alloc.c
mkannwischer: Fixed.

PR #1005, opened 2026-03-27
PoC: Reduce large struct allocations to <= 13/17/21 KiB for ML-DSA-44/65/87

mkannwischer:
This is a proof of concept how we could get the stack consumption of {keygen,sign,verify} to <= 13/17/21 KiB of allocations via `MLD_ALLOC` in `REDUCE_RAM`-mode. Additionally, as before we need a little bit of stack memory - I have measures 2.5 KiB on my machine, so the overall memory consumption should be comfortably below 16/20/24 KiB.

**Warning** This is a quick-and-dirty PoC and I do not recommend relying on it. The CBMC proofs are work in progress. This PR will definitely not be merged in one piece, but instead in smaller steps.


mkannwischer:
This has been hoisted out in various smaller PRs that has been merged by now.

PR #1007, opened 2026-04-01
Lowram: Share buffers with non-overlapping lifetimes in verify

mkannwischer:
- Hoisted out of #1005 
- Depends on #1066 
- Depends on #1067 
- Depends on #1037 
- Depends on #1070


mldsa/src/sign.c
```diff
@@ -1051,42 +1051,50 @@ int mld_sign_verify_internal(const uint8_t *sig, size_t siglen,
...
 
-  mld_unpack_pk(rho, t1, pk);
+  mld_memcpy(rho, pk, MLDSA_SEEDBYTES);
+
+  mld_memcpy(c, sig, MLDSA_CTILDEBYTES);
+  mld_polyvecl_unpack_z(z, sig + MLDSA_CTILDEBYTES);
 
-  /* mld_unpack_sig and mld_polyvecl_chknorm signal failure through a
-   * single non-zero error code that's not yet aligned with MLD_ERR_XXX.
-   * Map it to MLD_ERR_FAIL explicitly. */
```
hanno-becker, line 1084: This comment may be useful to retain.

mldsa/src/sign.c
```diff
@@ -1111,23 +1119,31 @@ int mld_sign_verify_internal(const uint8_t *sig, size_t siglen,
   }
 
   /* Matrix-vector multiplication; compute Az - c2^dt1 */
+  mld_polyvecl_ntt(z);
+  mld_polyvec_matrix_expand(&reuse->mat, rho);
```
hanno-becker: NB Why don't we directly expand from `pk` here? Is the `memcpy(rho, pk, ..)` necessary?

mkannwischer: No good reason. Let's change it.

mldsa/src/sign.c
```diff
@@ -1111,23 +1119,31 @@ int mld_sign_verify_internal(const uint8_t *sig, size_t siglen,
...
+  mld_polyvec_matrix_pointwise_montgomery(w1, &reuse->mat, z);
+
   mld_poly_challenge(cp, c);
   mld_poly_ntt(cp);
-  mld_polyveck_shiftl(t1);
-  mld_polyveck_ntt(t1);
-  mld_polyveck_pointwise_poly_montgomery(t1, cp);
+  mld_unpack_pk(rho, &reuse->t1, pk);
```
hanno-becker: Mh. Bit unsatisfactory that we call `unpack_pk` here even though we don't need `rho` again.

mkannwischer: We can eliminate unpack_pk and instead unpack t1 directly. Okay?

mldsa/src/sign.c
```diff
@@ -1051,42 +1051,50 @@ int mld_sign_verify_internal(const uint8_t *sig, size_t siglen,
...
   if (buf == NULL || rho == NULL || mu == NULL || c == NULL || c2 == NULL ||
-      cp == NULL || mat == NULL || z == NULL || t1 == NULL || tmp == NULL ||
-      h == NULL)
+      zcp == NULL || w1 == NULL || reuse == NULL)
   {
     ret = MLD_ERR_OUT_OF_MEMORY;
     goto cleanup;
   }
+  z = &zcp->z;
+  cp = &zcp->cp;
```
hanno-becker, line 1091: Nit: We're deviating from the pattern of declaring pointers into the `union` below when we directly use `&reuse->foo`

mldsa/src/sign.c
mkannwischer: Re-added.

mldsa/src/sign.c
mkannwischer: Changed to `unpack_pk_t1`

mldsa/src/sign.c
mkannwischer: Removed.

mldsa/src/sign.c
mkannwischer: Good point. introduced the pointers.

mldsa/src/packing.c
```diff
@@ -173,29 +171,9 @@ void mld_pack_sig_z(uint8_t sig[MLDSA_CRYPTO_BYTES], const mld_poly *zi,
...
- * Description: Unpack raw hint bytes into a polyveck
- *              struct
- *
- * Arguments:   - mld_polyveck *h: pointer to output hint vector h
- *              - const uint8_t packed_hints[MLDSA_POLYVECH_PACKEDBYTES]:
- *                raw hint bytes
- *
- * Returns 1 in case of malformed hints; otherwise 0.
- **************************************************/
-static int mld_unpack_hints(
```
hanno-becker, line 188: Nit: We only call this once, with the signature buffer. Perhaps should we rename this `mld_sig_unpack_hints` and take the signature buffer as the argument, so that the signature offset of the hint is encapsulated in this function and need not be known in the high-level logic?

mldsa/src/sign.c
```diff
@@ -1051,42 +1051,55 @@ int mld_sign_verify_internal(const uint8_t *sig, size_t siglen,
 {
   int ret, cmp;
 
+  typedef union
+  {
+    mld_polyvecl z;
+    mld_poly cp;
+  } zcp_u;
+  mld_polyvecl *z;
+  mld_poly *cp;
```
hanno-becker, line 1060: Not critical, but: Are you aware of any compiler (extension) which allows instrumentation of `union`s so that when you ever try to read from a field not previous written to, you get an error? Where unions are used for memory reduction (as opposed to data conversion -- where the read-X after write-Y pattern is of course the whole point) you should not see this.

mldsa/src/packing.c
mkannwischer: Yes, good idea to hide that away. Changed.

mldsa/src/sign.c
mkannwischer: No, never heard of something like this.

mkannwischer:
I tried digging into why there is a substantial performance regression for x86_64 in this PR, but I cannot quite figure it out. When I run this PR on a c7i.metal-24xl, I get about the same performance as in main - most of the time. Verify on x86 Sapphire Rapids seems to behave bimodally - most runs land in a fast band, ~6-12% land in a slow band that's 10-18% slower. As far as I can tell it's true on main as well as on this branch, at similar rates. It seems to correlate with where stack buffers happen to land mod 4096, which varies per run because of ASLR.

If I align w1 to 4096 bytes, the performance regression goes away in CI, but I don't know how that can help us with fixing it.

I tried disabling ASLR (see #1091). That indeed eliminates the variance for me and it definitely also impacts performance in CI, but it doesn't result in the performance regression in this PR to go away. Also, it does not allow to reproduce the performance regression locally.

My initial hypothesis was that this is x86's 4 KiB store-buffer aliasing: when a load has the same low-12 address bits as a recently-issued store, x86_64 flags a false dependency and delays the load. That model fits the basic shape: stack addresses varies with ASLR while rodata low-12 is fixed within the binary, so you'd expect the alias to fire when the stack happens to land at a low-12 that overlaps a hot rodata access. `perf record -e ld_blocks.address_alias` is consistent with that. But when I tried to confirm by shifting the suspected victims, things didn't fall into place.

My conclusion is that this PR is likely not introducing the regression. Still we should investigate why ASLR has such a massive impact on our performance - it really shouldn't. #1091's benchmark show a speed-up for our native implementations just by disabling ASLR suggesting some of our assembly routines suffer proof performance in the case of unlucky stack placement due to ASLR.

@hanno-becker, can we go ahead and merge this PR and investigate this separately?

hanno-becker:
@mkannwischer Thank you for the analysis. I think it gives sufficient confidence that this is a pre-existing issue, not something introduced in this PR. We should analyze it independently. Can you open an issue?

hanno-becker (approved):
Approving under the assumption that we will investigate the performance degradation on x86 with some urgency.

PR #1009, opened 2026-04-01
HOL-Light: improve tooling and CI

### LINKED ISSUE #1280: Port: HOL-Light: improve tooling and CI
See https://github.com/pq-code-package/mlkem-native/pull/1634

@L-series, do you think you could do that this week? Would be great to have this in before the release.
Otherwise, I can take a look.

L-series:
This PR is a port from: https://github.com/pq-code-package/mlkem-native/pull/1634

Note that this is currently blocked by: https://github.com/pq-code-package/mldsa-native/issues/1001

- Resolves https://github.com/pq-code-package/mldsa-native/issues/1280
- Depends on https://github.com/pq-code-package/mldsa-native/pull/1404

### DIFF
--- a/proofs/hol_light/aarch64/list_thms.sh
+++ b/proofs/hol_light/aarch64/list_thms.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+# Copyright (c) The mldsa-native project authors
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+#
+# This tiny script lists HOL-Light theorem names declared as proofs.
+
+ROOT=$(git rev-parse --show-toplevel)
+cd "$ROOT" || exit
+
+if [[ $# == 0 ]]; then
+  set -- proofs/hol_light/aarch64/proofs/*.ml
+fi
+
+grep -hE "^[[:space:]]*let[[:space:]]+[^[:space:]]+[[:space:]]+=[[:space:]]+(time[[:space:]]+)?prove" "$@" |
+  sed -E "s/^[[:space:]]*let[[:space:]]+([^[:space:]]+)[[:space:]]+=.*/\1/"
--- a/scripts/lint
+++ b/scripts/lint
@@ -399,6 +399,81 @@ gh_group_start "Check HOL-Light imports"
 check-hol-light-imports
 gh_group_end
 
+# Derive the theorem-name prefix from a proof routine's basename:
+#   - strip the platform suffix (_aarch64_asm or _avx2_asm)
+#   - uppercase
+# The routine basename already carries the namespace (`mldsa_` for arithmetic,
+# `keccak_` for the namespace-agnostic keccak proofs), so no fixup is needed.
+theorem-prefix()
+{
+  local routine="$1"
+  local platform_suffix="$2"
+  local base="${routine%${platform_suffix}}"
+  echo "$base" | tr '[:lower:]' '[:upper:]'
+}
+
+check-theorems()
+{
+  local success=true
+  for arch in aarch64 x86_64; do
+    local proofs_dir="$ROOT/proofs/hol_light/$arch/proofs"
+    local platform_suffix="_aarch64_asm"
+    [[ $arch == "x86_64" ]] && platform_suffix="_avx2_asm"
+    local proofs
+    proofs=$("$ROOT/proofs/hol_light/$arch/list_proofs.sh")
+
+    for routine in $proofs; do
+      local file="$proofs_dir/${routine}.ml"
+      local prefix
+      local theorems
+      prefix=$(theorem-prefix "$routine" "$platform_suffix")
+      theorems=$("$ROOT/proofs/hol_light/$arch/list_thms.sh" "$file")
+
+      # rej_uniform uses MEMSAFE in place of SAFE because no constant-time proof exists
+      local safe="SAFE"
+      [[ ${routine} == mldsa_rej_uniform_* ]] && safe="MEMSAFE"
+
+      local expected=(
+        "${prefix}_CORRECT"
+        "${prefix}_SUBROUTINE_CORRECT"
+        "${prefix}_SUBROUTINE_${safe}"
+      )
+      if [[ $arch == "x86_64" ]]; then
+        expected+=(
+          "${prefix}_${safe}"
+          "${prefix}_NOIBT_SUBROUTINE_CORRECT"
+          "${prefix}_NOIBT_SUBROUTINE_${safe}"
+        )
+      fi
+
+      for theorem in "${expected[@]}"; do
+        if ! grep -qx "$theorem" <<<"$theorems"; then
+          gh_error "$file" "" "Missing theorem" "${file}: ${theorem} not found"
+          success=false
+        fi
+      done
+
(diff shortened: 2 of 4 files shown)


L-series:
Hello @mkannwischer @hanno-becker, appologies for the delay with this, I got quite distracted with a math paper i was working on this last month :laughing: . This should be ready for review now, the check theorems script is slightly different here than what it is in mlkem as it also treats the IBT case.

mkannwischer:
> Hello @mkannwischer @hanno-becker, appologies for the delay with this, I got quite distracted with a math paper i was working on this last month 😆 .

No worries. Thanks for porting this!

> This should be ready for review now, the check theorems script is slightly different here than what it is in mlkem as it also treats the IBT case.

Can you elaborate why it needs to be slightly different? IBT isn't something that's specific to mldsa. Should it be ported to mlkem, too?

mkannwischer:
@L-series can you please fix the DCO signoff?

L-series:
> Can you elaborate why it needs to be slightly different? IBT isn't something that's specific to mldsa. Should it be ported to mlkem, too?

@mkannwischer the logic for capturing the `let PROVE = ...` blocks in mlkem is somehow simpler than in mldsa as all the proof declarations exist on a single line and thus the grep can catch them. 

In mldsa, some of the proofs occur with the following slightly more complicated patterns:

```
let MLDSA_NTT_SAFE =
  REWRITE_RULE [...]
  (time prove (...));;
```

and 

```
 let MLDSA_REJ_UNIFORM_ETA2_NOIBT_SUBROUTINE_CORRECT =
   let correct_bound = prove ... in
   let th = prove ... in
   th;;
```

This occurs around a dozen times.

To generalize the logic in the list_thms script, I've made it so that we scan for the theorem producing declaration within the entire top level `let NAME = xyz` block.

There might be some simpler way of doing this however. Please let me know what you think!

scripts/lint
```diff
@@ -399,6 +399,86 @@ gh_group_start "Check HOL-Light imports"
 check-hol-light-imports
 gh_group_end
 
+# Strip the platform suffix, normalize the routine name to its theorem prefix,
+# and prepend MLDSA_ unless the routine is a keccak proof.
+theorem-prefix()
```
mkannwischer, line 407: This function diverges from mlkem-native. Why?

scripts/lint
```diff
@@ -399,6 +399,86 @@ gh_group_start "Check HOL-Light imports"
...
+
+    for routine in $proofs; do
+      local file="$proofs_dir/${routine}.ml"
+      local prefix
+      local theorems
+      prefix=$(theorem-prefix "$routine" "$platform_suffix")
+      theorems=$("$ROOT/proofs/hol_light/$arch/list_thms.sh" "$file")
+
+      local safe="SAFE"
+      [[ $routine == *rej_uniform* ]] && safe="MEMSAFE"
```
mkannwischer: add a comment here like in mlkem-native

proofs/hol_light/aarch64/proofs/keccak_f1600_x4_v8a_scalar_hybrid_aarch64_asm.ml
```diff
@@ -1035,7 +1035,7 @@ let LENGTH_SIMPLIFY_CONV =
 (* Correctness proof                                                         *)
 (* ------------------------------------------------------------------------- *)
 
-let KECCAK_F1600_X4_V8A_SCALAR_CORRECT = prove
+let KECCAK_F1600_X4_V8A_SCALAR_HYBRID_CORRECT = prove
```
mkannwischer, line 1038: Can you do a pass through the files and use the same prefix in all places, please? i.e., KECCAK_F1600_X4_V8A_SCALAR_PREAMBLE_LENGTH -> KECCAK_F1600_X4_V8A_SCALAR_HYBRID_PREAMBLE_LENGTH, etc.

scripts/tests
```diff
@@ -1083,14 +1083,14 @@ class Tests:
...
+            arch = self.args.arch
+        elif platform.machine().lower() in ["arm64", "aarch64"]:
             arch = "aarch64"
-        elif machine in ["x86_64"]:
+        elif platform.machine().lower() in ["x86_64"]:
             arch = "x86_64"
         else:
-            self.fail(f"HOL-Light unsupported architecture: {machine}")
-            self.check_fail()
+            return
```
mkannwischer: Why was this changed?  Please change back.

proofs/hol_light/aarch64/list_thms.sh
mkannwischer, line 1: This file doesn't seem to align to the one in mlkem-native. Was that done intentionally?

mkannwischer (changes requested):
@L-series, sorry for the delay.
A few things should still be changed before we can merge this one.

scripts/tests
L-series: Fixed.

proofs/hol_light/aarch64/proofs/keccak_f1600_x4_v8a_scalar_hybrid_aarch64_asm.ml
L-series: Fixed - sorry about that

scripts/lint
L-series: Fixed

scripts/lint
L-series: Was using a small local llm to port the mlkem patch, I should've inspected closer.. Everything should be in line now!

proofs/hol_light/aarch64/list_thms.sh
L-series: Yes indeed, as I mentioned [in the comment above](https://github.com/pq-code-package/mldsa-native/pull/1009#issuecomment-5171107501), there is a divergence in the format of some of the proofs in mldsa as compared to mlkem. The grep which is used in the mlkem version of list_thms cannot capture these differences, so instead I modified the script to look at entire `let name = ... prove` blocks and search for the correct patterns.

L-series (commented):
The small issues are resolved, however the question of the difference in the list_thms script still persists. Please let me know if you think this approach is correct.

mkannwischer:
> The small issues are resolved, however the question of the difference in the list_thms script still persists. Please let me know if you think this approach is correct.

So sorry for the delay again! I'll prioritize this now. 
I don't think there is a good reason for these proofs have a different shape. I'll try to refactor this in a separate PR, then we can rebase your PR and use the simpler script.

mkannwischer:
> > The small issues are resolved, however the question of the difference in the list_thms script still persists. Please let me know if you think this approach is correct.
> 
> So sorry for the delay again! I'll prioritize this now. I don't think there is a good reason for these proofs have a different shape. I'll try to refactor this in a separate PR, then we can rebase your PR and use the simpler script.

I've hoisted out the proof changes to https://github.com/pq-code-package/mldsa-native/pull/1404 as those need careful review. The remaining 2 commits in this PR are then tooling only and they are fully aligned with mlkem-native.

mkannwischer (approved):
Thanks @L-series for the patience. This now looks good to me.

PR #1011, opened 2026-04-02
Lowram: Share buffers with non-overlapping lifetimes in keygen

mkannwischer:
- Hoisted out from #1005 

Reuse t0 as the accumulator in mld_compute_t0_t1_tr_from_sk_components, and have the caller provide s1 already in NTT form, removing two allocations (s1hat and t) from the helper.

In mld_sign_keypair_internal, share the s1 and t1 buffers via a union since s1hat is consumed before t1 is produced. Pack s1 into the secret key before the in-place NTT so the original coefficients are preserved.

Split mld_pack_sk into mld_pack_sk_s1 and mld_pack_sk_rho_key_tr_s2_t0 to support packing s1 independently before the NTT.


mldsa/src/poly.h
```diff
@@ -213,7 +213,7 @@ void mld_poly_power2round(mld_poly *a1, mld_poly *a0, const mld_poly *a)
 __contract__(
   requires(memory_no_alias(a0, sizeof(mld_poly)))
   requires(memory_no_alias(a1, sizeof(mld_poly)))
-  requires(memory_no_alias(a, sizeof(mld_poly)))
+  requires(a0 == a)
```
hanno-becker, line 227: If we assert `a0 == a`, we should probably change the function signature and document the destructive nature of the function.

mldsa/src/polyvec.h
```diff
@@ -369,7 +369,7 @@ void mld_polyveck_power2round(mld_polyveck *v1, mld_polyveck *v0,
 __contract__(
   requires(memory_no_alias(v1, sizeof(mld_polyveck)))
   requires(memory_no_alias(v0, sizeof(mld_polyveck)))
-  requires(memory_no_alias(v, sizeof(mld_polyveck)))
+  requires(v0 == v)
```
hanno-becker, line 386: Same here

mldsa/src/sign.c
```diff
@@ -215,17 +215,17 @@ MLD_MUST_CHECK_RETURN_VALUE
...
+    const mld_polyvecl *s1hat, const mld_polyveck *s2,
     MLD_CONFIG_CONTEXT_PARAMETER_TYPE context)
 __contract__(
   requires(memory_no_alias(t0, sizeof(mld_polyveck)))
   requires(memory_no_alias(t1, sizeof(mld_polyveck)))
   requires(memory_no_alias(tr, MLDSA_TRBYTES))
   requires(memory_no_alias(pk, MLDSA_CRYPTO_PUBLICKEYBYTES))
   requires(memory_no_alias(rho, MLDSA_SEEDBYTES))
-  requires(memory_no_alias(s1, sizeof(mld_polyvecl)))
+  requires(memory_no_alias(s1hat, sizeof(mld_polyvecl)))
```
hanno-becker: This spec is no longer correct when `s1hat` and `t1` may share their parent object. How would we adjust this?

hanno-becker (changes requested):
Similar to what we did with `poly_add`, I think we should make the now-destructive nature of `power2round` explicit in the signature and documentation.

We'd also need to adjust the CBMC spec for `compute_t0_t1_tr_from_sk_components` to account for `s1hat` and `t1` no longer being disjoint in memory. Can we do this already now (even if the call-site still uses a `struct`)?

mldsa/src/sign.c
mkannwischer: I suggest we worry about it later.
This problem goes away in https://github.com/pq-code-package/mldsa-native/pull/1030 anyway.

mldsa/src/poly.h
mkannwischer: I find an accumulating `poly_add` much more natural than a decomposition function like power2round with only 2 arguments.
Note that the implementation does not, in fact, need the `a0==0` assertion. One could rewrite the contract to allow both options, but I don't see much value in it.

mldsa/src/sign.c
hanno-becker: Ok

mldsa/src/poly.h
hanno-becker: Ok, can you just leave a comment then?

mldsa/src/polyvec.h
hanno-becker: Leave a comment, as above?

mldsa/src/sign.c
```diff
@@ -1524,8 +1519,10 @@ int mld_sign_pk_from_sk(uint8_t pk[MLDSA_CRYPTO_PUBLICKEYBYTES],
   chk2 = mld_polyveck_chknorm(s2, MLDSA_ETA + 1) & 0xFF;
 
   /* Recompute t0, t1, tr, and pk from rho, s1, s2 */
+  *s1hat = *s1;
```
hanno-becker: Sorry for my ignorance... why is this needed rather than an in-place modification of `s1`?

mkannwischer: Good catch. There is no reason. Will fix.

mldsa/src/sign.c
```diff
@@ -325,29 +311,37 @@ int mld_sign_keypair_internal(uint8_t pk[MLDSA_CRYPTO_PUBLICKEYBYTES],
...
+
   /* Sample s1 and s2 */
   mld_sample_s1_s2(s1, s2, rhoprime);
 
-  /* Compute t0, t1, tr, and pk from rho, s1, s2 */
+  /* Pack s1 into sk before NTT */
+  mld_pack_sk_s1(sk, s1);
+
+  /* NTT s1 in-place to use as s1hat */
+  mld_polyvecl_ntt(s1);
```
hanno-becker: Why did we hoist this out of `mld_compute_t0_t1_tr_from_sk_components`? We now need this twice, vs. once within the body of `mld_compute_t0_t1_tr_from_sk_components` as it was before.

hanno-becker (approved):

mldsa/src/sign.c
mkannwischer: No good reason. Must be an artifact of an older version of this PR. Thanks for catching it! Fixed.

Issue #418, opened 2025-08-09
AVX2: Replace intrinsics implementation of `rej_uniform` with assembly


mkannwischer:
For formally verifying functional correctness using HOL-Light, we will require assembly instead of intrinsics.

mkannwischer:
@jakemas, do you have time to look into this? Would be good to complete this before any integrations happen.

PR #1013, opened 2026-04-03
x86_64: Replace rej_uniform intrinsics with assembly

### LINKED ISSUE #926: AVX2: Replace intrinsics implementation of `rej_uniform` with assembly

jakemas:
Resolves #926 and #418 (?)

### Summary

- Replace AVX2 intrinsics implementation of `rej_uniform` with hand-written x86_64 assembly
- Table passed as parameter (consistent with aarch64 approach), avoiding external symbol references for simpasm compatibility
- All constants constructed from immediates (no `.rodata` section), enabling future HOL-Light formal verification
- Register name `#define`s with `#undef` cleanup for SCU builds (following mlkem-native pattern)
- Adds `poly_uniform` to component benchmark


dev/x86_64/src/arith_native_x86_64.h
```diff
@@ -65,7 +65,16 @@ void mld_nttunpack_avx2(int32_t *r);
...
+                              const uint8_t buf[MLD_AVX2_REJ_UNIFORM_BUFLEN],
+                              const uint8_t *table)
+__contract__(
+  requires(memory_no_alias(r, sizeof(int32_t) * MLDSA_N))
+  requires(memory_no_alias(buf, MLD_AVX2_REJ_UNIFORM_BUFLEN))
+  requires(table == (const uint8_t *)mld_rej_uniform_table)
+  assigns(memory_slice(r, sizeof(int32_t) * MLDSA_N))
+  ensures(return_value <= MLDSA_N)
+  ensures(array_bound(r, 0, (unsigned)return_value, 0, MLDSA_Q))
+);
```
mkannwischer: Please don't add contracts for which we do not have HOL-Light proofs yet - that can introduce dangerous gaps.

jakemas: Sorry, this PR is still under construction, so still in progress! I marked as draft.

mkannwischer: I see. No wories then. Let me know when you want me to take a look.

jakemas: Funny enough, I was seeing if I could add the Hol-light proof into this PR also, 3 missing instructions, just modeled them. Just setting up the proof framework! If we're OK with following up with the Hol-Light proof, would you like all contracts out? I know sometimes we also want to avoid CBMC gaps, but I suppose there is already an exception for the asmb files.

jakemas: @mkannwischer Please do check out the assembly and see what you think. I've attempted to make it somewhat similar to ML-KEM's implementation in a bid to help make the Hol-Light proof simpler, but I have chosen to avoid SSE in favor of AVX2 (`vmovmskps`, `vpmovzxbd`, and `vzeroupper`) -- ML-DSA's 23-bit coefficients require 32-bit lanes, which naturally fills a 256-bit YMM register for 8 elements per iteration, whereas ML-KEM's 12-bit coefficients fit in 16-bit SSE lanes.

jakemas: I wanted to see what CI thought of it too, but that Mingw error has been persistent for hours! And I wanna see those benchmarks, local testing gave around 1-2% improvement over intrinsics.

mkannwischer: Sounds like awesome progress- thank you @jakemas. I think the CBMC contracts should go into the same PR as the HOL-Light proofs - also because they have to be reviewed together. If we can do everything in one go that would of course be great, but if you want the intrinsics to be merged first, then that's also fine and the CBMC contracts should be removed. Can you elaborate what CBMC gaps we already have? 

Benchmarking CI should actually work without the Mingw jobs going through. Let me give it another try. If you want to test full CI, you can comment out the MIngw jobs in CI temporarily. If they don't fix it soon, I will disable them in main anyway.

mkannwischer: Ah. Your PR comes from a fork, so we cannot run the benchmarking CI as that needs EC2 auth. Please re-open the PR from a local branch.

jakemas: >Can you elaborate what CBMC gaps we already have?

None -- exactly, I didn't want to bring in any more by introducing this function. We're on the same page, you were just seeing my development live. I was scoping out the proofs as you checked in.

>Ah. Your PR comes from a fork, so we cannot run the benchmarking CI as that needs EC2 auth. Please re-open the PR from a local branch.

That's it! thanks!

PR #1014, opened 2026-04-03
x86_64 + HOL-Light: Replace rej_uniform intrinsics with assembly and HOL-Light CORRECT and MEMSAFE proofs

### LINKED ISSUE #418: AVX2: Replace intrinsics implementation of `rej_uniform` with assembly

### LINKED ISSUE #916: HOL-Light: Prove AVX2 `rej_uniform`

### LINKED ISSUE #926: AVX2: Replace intrinsics implementation of `rej_uniform` with assembly

jakemas:
## Summary

- Resolves #926 
- Resolves #418
- Resolves #916

Hol-light proof needs instructions from https://github.com/awslabs/s2n-bignum/pull/387

- Replace AVX2 intrinsics implementation of `rej_uniform` with hand-written x86_64 assembly
- Table passed as parameter (consistent with aarch64 approach), avoiding external symbol references for simpasm compatibility
- All constants constructed from immediates (no `.rodata` section), enabling future HOL-Light formal verification
- Register name `#define`s with `#undef` cleanup for SCU builds (following mlkem-native pattern)
- Adds `poly_uniform` to component benchmark
- HOL-Light proof infrastructure included (bytecode, table definition, proof skeleton, Makefile)

ML-DSA's 23-bit coefficients require 32-bit lanes, which naturally fills a 256-bit YMM register for 8 elements per iteration.

#### AMD EPYC 3rd gen (c6a) — opt

CYCLE Count
| Benchmark | Before | After | Change |
|---|---|---|---|
| ML-DSA-44 keypair | 68,874 | 66,828 | -3% |
| ML-DSA-44 sign | 187,594 | 184,181 | -2% |
| ML-DSA-44 verify | 68,993 | 65,665 | -5% |
| ML-DSA-65 keypair | 119,089 | 112,640 | -5% |
| ML-DSA-65 sign | 299,488 | 294,836 | -2% |
| ML-DSA-65 verify | 115,385 | 108,494 | -6% |
| ML-DSA-87 keypair | 203,754 | 185,518 | -9% |
| ML-DSA-87 sign | 396,462 | 378,579 | -5% |
| ML-DSA-87 verify | 196,231 | 177,157 | -10% |

### Proof
Includes HOL-Light and CBMC proofs, written by claude opus 4.7.

CORRECT
[HOL-Light / x86_64 HOL Light proof for mldsa_rej_uniform.S (pull_request)](https://github.com/pq-code-package/mldsa-native/actions/runs/25410336039/job/74531257826?pr=1014) Successful in 12m

CORRECT + MEMSAFE
[HOL-Light / x86_64 HOL Light proof for rej_uniform_avx2_asm.S (pull_request)](https://github.com/pq-code-package/mldsa-native/actions/runs/26673091515/job/78620472506?pr=1014)Successful in 21m

It took Claude Opus 4.7 1m around 3weeks to get the MEMSAFE proof. Since it records all token usage per prompt, I got it to go back to the prompt in which I asked it to create the MEMSAFE part of the proof to gather statistics: 

```
 ~10.35 billion tokens
 20,257 turns(i.e distinct API calls)
 ~30 days of development 
 ~100-125 hours of active user engagement (based on user clustering message timestamps)
185 attempted build iterations,
```


mkannwischer:
@jakemas, thanks for getting this into shape! ~~How far are you with the proof of it? Would it be an option to merge it together with the correctness proof?~~ Nevermind, this already has the proof. Sorry.

jakemas:
@mkannwischer ok, ready for review. The instruction PR will need to land in s2n-bignum first, while we wait I'll try the constant time proof -- but I'm really happy to be able to PR the conversion with a hol-light proof. Let me know if anything is missing.

jakemas:
ahh just saw you comment! Yes got the proof in, runs in ~12min!

proofs/hol_light/x86_64/proofs/mldsa_rej_uniform_table.ml
jakemas, line 1: I left this name generic, as it could be shared between arm/x86 proof.

mkannwischer:
> @mkannwischer ok, ready for review. The instruction PR will need to land in s2n-bignum first, while we wait I'll try the constant time proof -- but I'm really happy to be able to PR the conversion with a hol-light proof. Let me know if anything is missing.

Thanks! Yes, I agree that it's great to get the conversion and the proof in at the same time. Would be great if we can do the same for the remaining proofs.
I'll review later today.

Note that a constant-time proof is not needed for this function (all inputs are public), but we do want a memory safety proof like here: https://github.com/pq-code-package/mlkem-native/blob/2bf8e59f4330697b3924c572924136c96eb96960/proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml#L1562

dev/x86_64/src/arith_native_x86_64.h
```diff
@@ -73,10 +73,21 @@ __contract__(
...
 MLD_MUST_CHECK_RETURN_VALUE
-unsigned mld_rej_uniform_avx2(int32_t *r,
-                              const uint8_t buf[MLD_AVX2_REJ_UNIFORM_BUFLEN]);
+unsigned mld_rej_uniform_avx2_asm(
+    int32_t *r, const uint8_t buf[MLD_AVX2_REJ_UNIFORM_BUFLEN],
+    const uint8_t *table)
+__contract__(
+  requires(memory_no_alias(r, sizeof(int32_t) * MLDSA_N))
+  requires(memory_no_alias(buf, MLD_AVX2_REJ_UNIFORM_BUFLEN))
+  requires(memory_no_alias(table, 256 * sizeof(uint64_t)))
```
mkannwischer: the table actually needs to be == mld_rej_uniform_table

dev/x86_64/src/rej_uniform_avx2_asm.S
```diff
@@ -0,0 +1,173 @@
...
+
+        cmpl    $8380417, good                 // MLDSA_Q
+        jae     rej_uniform_avx2_asm_scalar    // reject if >= Q
+
+        movl    good, (out, %rax, 4)           // store valid coefficient
+        addl    $1, ctr                        // ctr++
+        jmp     rej_uniform_avx2_asm_scalar
+
+rej_uniform_avx2_asm_done:
+        vzeroupper
```
mkannwischer: We don't have `vzeroupper` in any other routine. I don't know enough about x86_64 to know how important it is, but we should either have it everywhere or nowhere.

nix/s2n_bignum/default.nix
```diff
@@ -4,12 +4,15 @@
...
+  # Pinned to https://github.com/awslabs/s2n-bignum/pull/387 head,
+  # which adds VMOVMSKPS, VPMOVZXBD, and VZEROUPPER instruction models
+  # required by the x86_64 rej_uniform proof.
+  version = "4c4fe1dfc8b79720013517a7b4dec9014c85fcf2";
   src = fetchFromGitHub {
     owner = "awslabs";
     repo = "s2n-bignum";
     rev = "${version}";
-    hash = "sha256-kfc8X2e+voefttshSUdifDc3Qn+dx0Gq5ENNLhWIdw0=";
+    hash = "sha256-64MJOqoDunpn6fx1j9P4+fDoRNZ8GRTB/d4C2JWvxFA=";
```
mkannwischer: Leaving a comment here to remind us that we still have to change this.

proofs/hol_light/x86_64/mldsa/rej_uniform_avx2_asm.S
mkannwischer, line 1: This file should be autogenerated via `autogen`

proofs/hol_light/x86_64/proofs/mldsa_rej_uniform_table.ml
```diff
@@ -0,0 +1,268 @@
+(*
+ * Copyright (c) The mldsa-native project authors
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT-0
+ *)
+
+(* Lookup table for ML-DSA rejection uniform sampling. *)
+(* Each entry is 8 bytes: permutation indices for VPERMD. *)
+
+let mldsa_rej_uniform_table = (REWRITE_RULE[MAP] o define)
```
mkannwischer, line 17: This file should be autogenerated via autogen

proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml
```diff
@@ -0,0 +1,4038 @@
+(*
+ * Copyright (c) The mldsa-native project authors
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT-0
+ *)
+
+(* ========================================================================= *)
```
mkannwischer, line 7: please include this proof in the README.

proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml
```diff
@@ -0,0 +1,4038 @@
...
+              read RSP s = word_add stackpointer (word 8) /\
+              (let outlist = SUB_LIST(0,256) (REJ_SAMPLE inlist) in
+               let outlen = LENGTH outlist in
+               C_RETURN s = word outlen /\
+               read(memory :> bytes(res,4 * outlen)) s =
+                 num_of_wordlist outlist))
+         (MAYCHANGE [RSP] ,, MAYCHANGE_REGS_AND_FLAGS_PERMITTED_BY_ABI ,,
+          MAYCHANGE [memory :> bytes(res,1024)])`,
+  MATCH_ACCEPT_TAC(ADD_IBT_RULE MLDSA_REJ_UNIFORM_NOIBT_SUBROUTINE_CORRECT));;
+
```
mkannwischer, line 4169: Leaving a comment here to remind us that the `MEMSAFE` proof still needs to be added here.

proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml
```diff
@@ -0,0 +1,4038 @@
...
+let DIMINDEX_24 = DIMINDEX_CONV `dimindex(:24)`;;
+
+let VAL_MOD_23_EQ_AND = prove
+ (`!w:24 word. (word(val w MOD 2 EXP 23):int32) =
+               word_and (word_zx w:int32) (word 8388607)`,
+  GEN_TAC THEN CONV_TAC WORD_BLAST);;
+
+let REJ_SAMPLE = define
+ `REJ_SAMPLE l = FILTER (\x:int32. val x < 8380417)
+    (MAP (\x:24 word. word(val x MOD 2 EXP 23):int32) l)`;;
```
mkannwischer: The spec should be moved to `mldsa_specs.ml` so we can re-use it for the aarch64 proof.

proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml
```diff
@@ -0,0 +1,4038 @@
...
+              C_ARGUMENTS [res; buf; table] s /\
+              read(memory :> bytes(buf,840)) s = num_of_wordlist inlist /\
+              read(memory :> bytes(table,2048)) s =
+                num_of_wordlist(mldsa_rej_uniform_table:byte list))
+         (\s. read RIP s = returnaddress /\
+              read RSP s = word_add stackpointer (word 8) /\
+              (let outlist = SUB_LIST(0,256) (REJ_SAMPLE inlist) in
+               let outlen = LENGTH outlist in
+               C_RETURN s = word outlen /\
+               read(memory :> bytes(res,4 * outlen)) s =
```
mkannwischer, line 4161: Please add an explicit bound post-condition that all coefficients < q here to match the CBMC spec.

mldsa/src/native/x86_64/src/arith_native_x86_64.h
```diff
@@ -73,10 +73,21 @@ __contract__(
...
+ * in proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml */
 MLD_MUST_CHECK_RETURN_VALUE
-unsigned mld_rej_uniform_avx2(int32_t *r,
-                              const uint8_t buf[MLD_AVX2_REJ_UNIFORM_BUFLEN]);
+unsigned mld_rej_uniform_avx2_asm(
+    int32_t *r, const uint8_t buf[MLD_AVX2_REJ_UNIFORM_BUFLEN],
+    const uint8_t *table)
+__contract__(
+  requires(memory_no_alias(r, sizeof(int32_t) * MLDSA_N))
+  requires(memory_no_alias(buf, MLD_AVX2_REJ_UNIFORM_BUFLEN))
```
mkannwischer: Let's change this to say 840 so it matches the HOL-light spec exactly.

mldsa/src/native/x86_64/src/arith_native_x86_64.h
```diff
@@ -73,10 +73,21 @@ __contract__(
...
+#define mld_rej_uniform_avx2_asm MLD_NAMESPACE(rej_uniform_avx2_asm)
+/* This must be kept in sync with the HOL-Light specification
+ * in proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml */
 MLD_MUST_CHECK_RETURN_VALUE
-unsigned mld_rej_uniform_avx2(int32_t *r,
-                              const uint8_t buf[MLD_AVX2_REJ_UNIFORM_BUFLEN]);
+unsigned mld_rej_uniform_avx2_asm(
+    int32_t *r, const uint8_t buf[MLD_AVX2_REJ_UNIFORM_BUFLEN],
+    const uint8_t *table)
+__contract__(
```
mkannwischer, line 87: Add the comment here that this needs to be kept in sync with the HOL-light spec.

proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml
```diff
@@ -0,0 +1,4038 @@
...
+                 num_of_wordlist outlist))
+         (MAYCHANGE [RSP] ,, MAYCHANGE_REGS_AND_FLAGS_PERMITTED_BY_ABI ,,
+          MAYCHANGE [memory :> bytes(res,1024)])`,
+  X86_PROMOTE_RETURN_NOSTACK_TAC mldsa_rej_uniform_tmc
+    MLDSA_REJ_UNIFORM_CORRECT);;
+
```
mkannwischer, line 4134: Add a comment here that this needs to be kept in sync wityh the CBMC spec.

mkannwischer (changes requested):
Thanks @jakemas. Here is a first set of comments.

dev/x86_64/src/arith_native_x86_64.h
jakemas: Fixed — `requires(table == (const uint8_t *)mld_rej_uniform_table)` in both dev/ and mldsa/src/native/ copies.

dev/x86_64/src/rej_uniform_avx2_asm.S
jakemas: Fixed — dropped `vzeroupper` from both dev/ and mldsa/src/native/ copies of the .S for consistency with the other ML-DSA routines. Proof adjusted accordingly.

nix/s2n_bignum/default.nix
jakemas: Still pinned to s2n-bignum PR #387 / #401 (the branch that carries VMOVMSKPS, VPMOVZXBD, VZEROUPPER + the mldsa_rej_uniform proof). Will update the pin once those are merged into s2n-bignum main.

proofs/hol_light/x86_64/mldsa/rej_uniform_avx2_asm.S
jakemas: Fixed — added `rej_uniform_avx2_asm.S` to the x86_64 joblist in `scripts/autogen` (`joblist_x86_64`), so `proofs/hol_light/x86_64/mldsa/rej_uniform_avx2_asm.S` is now regenerated by `scripts/autogen`.

proofs/hol_light/x86_64/proofs/mldsa_rej_uniform_table.ml
jakemas: Fixed — added `gen_avx2_hol_light_rej_uniform_table` to `scripts/autogen`, invoked from `gen_zeta_tables`. `proofs/hol_light/x86_64/proofs/mldsa_rej_uniform_table.ml` is now regenerated alongside the C/aarch64 lookup tables (mirrors the mlkem-native pattern).

proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml
jakemas: Fixed — added an entry under "x86_64 / ML-DSA Arithmetic" pointing at `x86_64/mldsa/rej_uniform_avx2_asm.S`.

proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml
jakemas: Fixed — `REJ_SAMPLE`, `REJ_SAMPLE_EMPTY`, `REJ_SAMPLE_APPEND` now live in `proofs/hol_light/common/mldsa_specs.ml` (shared between arches, matching the shape used by s2n-bignum #378 for aarch64). The x86-only derived lemmas (`REJ_SAMPLE_SPLIT`, `REJ_SAMPLE_PREFIX_256`, `REJ_SAMPLE_STEP_LE`) stay in `rej_uniform_avx2_asm.ml` since they're only used by the AVX2 scalar-tail analysis.

mldsa/src/native/x86_64/src/arith_native_x86_64.h
jakemas: Fixed — `requires(memory_no_alias(buf, 840))` (literal 840 matches the HOL-Light spec exactly), in both dev/ and mldsa/src/native/ copies.

mldsa/src/native/x86_64/src/arith_native_x86_64.h
jakemas: Fixed — added `/* This contract must be kept in sync with the HOL-Light specification in proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml */` above the `__contract__` block in both dev/ and mldsa/src/native/ copies.

proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml
jakemas: Fixed — added a comment block above `SUBROUTINE_CORRECT variants` noting these specifications must be kept in sync with the CBMC contract in `dev/x86_64/src/arith_native_x86_64.h` / `mldsa/src/native/x86_64/src/arith_native_x86_64.h`.

proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml
jakemas: think i got this, testing now added:

```
(!i. i < outlen
    ==> val(read(memory :> bytes32
                  (word_add res (word(4 * i)))) s) < 8380417)))
```

proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml
jakemas: ooft, its not an easy one

jakemas: ok got it! Testing in CI now! woo!

jakemas:
@mkannwischer MEMSAFE is done now, all review addressed

jakemas:
Updated to follow format introduced in https://github.com/pq-code-package/mldsa-native/commit/d417b6bb60527a378ed9b8170c0e2d8d71cf572e

scripts/autogen
```diff
@@ -1328,6 +1328,41 @@ def gen_avx2_rej_uniform_table():
...
+    def get_set_bits_idxs(i):
+        bits = list(map(int, format(i, "08b")))
+        bits.reverse()
+        return [bit_idx for bit_idx in range(8) if bits[bit_idx] == 1]
+
+    def gen_rows():
+        for i in range(256):
+            idxs = get_set_bits_idxs(i)
+            idxs = idxs + [0] * (8 - len(idxs))
+            yield idxs
```
mkannwischer: This function should just call the one used to generate the actual constants (gen_avx2_rej_uniform_table_rows) so we can be sure they are the same.

mldsa/src/native/x86_64/meta.h
```diff
@@ -80,7 +80,8 @@ static MLD_INLINE int mld_rej_uniform_native(int32_t *r, unsigned len,
   }
 
   /* Safety: outlen is at most MLDSA_N and, hence, this cast is safe. */
-  return (int)mld_rej_uniform_avx2(r, buf);
+  return (int)mld_rej_uniform_avx2_asm(r, buf,
+                                       (const uint8_t *)mld_rej_uniform_table);
```
mkannwischer: can we declare mld_rej_uniform_table as a flat array so this cast here is not needed?

mldsa/src/native/x86_64/src/arith_native_x86_64.h
```diff
@@ -77,10 +77,21 @@ __contract__(
...
+/* This contract must be kept in sync with the HOL-Light specification
+ * in proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml */
+MLD_MUST_CHECK_RETURN_VALUE MLD_SYSV_ABI
+unsigned mld_rej_uniform_avx2_asm(
+    int32_t *r, const uint8_t buf[MLD_AVX2_REJ_UNIFORM_BUFLEN],
+    const uint8_t *table)
+__contract__(
+  requires(memory_no_alias(r, sizeof(int32_t) * MLDSA_N))
+  requires(memory_no_alias(buf, 840))
+  requires(table == (const uint8_t *)mld_rej_uniform_table)
```
mkannwischer: can we declare mld_rej_uniform_table as a flat array so this cast here is not needed?

proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml
```diff
@@ -0,0 +1,7200 @@
...
+                        DIMINDEX_32; DIMINDEX_64] THEN
+            CONV_TAC NUM_REDUCE_CONV THEN
+            IMP_REWRITE_TAC[MOD_LT] THEN
+            UNDISCH_TAC `24 * N + 3 * i <= 837` THEN ARITH_TAC;
+            (* Memory *)
+            SUBGOAL_THEN `curlen + SUC 0 = curlen + 1` SUBST1_TAC THENL
+             [ARITH_TAC; ALL_TAC] THEN
+            SUBGOAL_THEN `4 * (curlen + 1) = 4 * curlen + 4` SUBST1_TAC THENL
+             [ARITH_TAC; ALL_TAC] THEN
+            FIRST_ASSUM(fun th ->
```
mkannwischer: Can we remove the debug `Printf.printf ` statements or do they still serve any purpose?

mkannwischer: I'm confused. Why are there separate `_MEMSAFE` and `_SAFE` proofs? Shouldn't this proof only have the `_MEMSAFE` theorems?

mkannwischer (changes requested):
Thanks @jakemas and sorry for the long silence. 
The core of the proof looks good to me - I have a few more comments/questions below.

dev/x86_64/src/rej_uniform_avx2_asm.S
```diff
@@ -0,0 +1,172 @@
...
+        .balign 4
+        .global MLD_ASM_NAMESPACE(rej_uniform_avx2_asm)
+MLD_ASM_FN_SYMBOL(rej_uniform_avx2_asm)
+
+/*
+ * Construct the shuffle mask for extracting 8 x 23-bit values from 24 bytes.
+ *
+ * After vpermq with 0x94, the 32 loaded bytes are rearranged as:
+ *   Low  128 bits: bytes [0..15]  (original 64-bit lanes 0, 1)
+ *   High 128 bits: bytes [8..23]  (original 64-bit lanes 1, 2)
```
hanno-becker: Nit: The comment is slightly confusing by itself, because the semantic granularity of the input is in bytes, not 64-bit. 64-bit granularity comes from `vpermq`, but this isn't mentioned here.

Suggestion:
```
* vpermq with 0x94(=0b 10 01 01 00) permutes 64-bit lanes via (0,1,2,3) -> (0,1,1,2).
* The loaded 32 bytes are thus rearranged as:
*   Low  128 bits: bytes [0..15] of original 32-byte
*   High 128 bits: bytes [8..23] of original 32-byte
```

dev/x86_64/src/rej_uniform_avx2_asm.S
```diff
@@ -0,0 +1,172 @@
...
+        vmovd   good, %xmm1
+        vpbroadcastd %xmm1, mask              // mask: 23-bit extraction
+
+        movl    $8380417, good                 // MLDSA_Q
+        vmovd   good, %xmm2
+        vpbroadcastd %xmm2, bound             // bound: rejection threshold
+
+// Initialize counters
+        xorl    ctr, ctr                       // ctr = 0
+        xorl    pos, pos                       // pos = 0
```
hanno-becker: Nit nit: indentation of comments. Also, can we use `/* .. */` style? I know this is wrong in many other ASM files, but when we touch/add code we may as well adjust it.

dev/x86_64/src/rej_uniform_avx2_asm.S
```diff
@@ -0,0 +1,172 @@
...
+
+/*
+ * Main SIMD loop: process 24 input bytes into up to 8 coefficients
+ * per iteration. Loops while ctr <= MLDSA_N - 8 and pos <= BUFLEN - 32.
```
hanno-becker, line 136: It's noteworthy that this approach differs from the mlkem-native rejection sampling, which buffers into a temporary slightly oversized stack buffer, to avoid the scalar tail loop.

I don't know which is better, nor is it necessary that we align this now, but just noting.

My gut feeling is that it's probably better the way it's done here.

dev/x86_64/src/rej_uniform_avx2_asm.S
```diff
@@ -0,0 +1,172 @@
...
+ */
+rej_uniform_avx2_asm_loop:
+        cmpl    $248, ctr                      // MLDSA_N - 8
+        ja      rej_uniform_avx2_asm_scalar
+        cmpl    $808, pos                      // MLD_AVX2_REJ_UNIFORM_BUFLEN - 32
```
hanno-becker: Nit: Can you use the macro constant instead of a comment here?

dev/x86_64/src/rej_uniform_avx2_asm.S
```diff
@@ -0,0 +1,172 @@
...
+        vpermq  $0x94, data, data              // rearrange 64-bit lanes: [2,1,1,0]
+        vpshufb idx8, data, data               // extract 8 x 3-byte groups
+        vpand   mask, data, data               // mask to 23 bits
+
+        vpsubd  bound, data, cmp_result        // d - Q: negative if d < Q (valid)
+        vmovmskps cmp_result, good             // extract sign bits as 8-bit mask
+
+        popcntl good, cnt                      // count valid coefficients
+
+        vmovq   (tab, %r8, 8), %xmm4          // load permutation from table[good]
```
hanno-becker: Nit: mix of symbolic `good` vs raw `r8`. Can we get the notation consistent (ideally symbolics only)? (If you need variants of the same register, you can do something like `name_x,name_y,..`

dev/x86_64/src/rej_uniform_avx2_asm.S
```diff
@@ -0,0 +1,172 @@
...
+        vpmovzxbd %xmm4, cmp_result            // zero-extend to 8 dword indices
+        vpermd  data, cmp_result, data          // compact valid coefficients to front
+
+        vmovdqu data, (out, %rax, 4)           // store at r[ctr]
```
hanno-becker: Same comment as above: Let's stick to symbolic register names.

dev/x86_64/src/rej_uniform_avx2_asm.S
```diff
@@ -0,0 +1,172 @@
...
+        jmp     rej_uniform_avx2_asm_loop
+
+/*
+ * Scalar fallback loop: process remaining bytes one coefficient at a time.
+ * Loops while ctr < MLDSA_N and pos <= BUFLEN - 3.
+ */
+rej_uniform_avx2_asm_scalar:
+        cmpl    $256, ctr                      // MLDSA_N
+        jae     rej_uniform_avx2_asm_done
+        cmpl    $837, pos                      // MLD_AVX2_REJ_UNIFORM_BUFLEN - 3
```
hanno-becker: This can never fire, can it? If so, can we remove it?

hanno-becker (changes requested):
Thanks @jakemas, and sorry this is again taking so long.

In addition to @mkannwischer's important comment that the constant-tables should come from the same source (that's our only argument for why they are consistent across proof and code -- we have no other check!), there's a dead branch in the rejection sampling I believe that can be removed.

dev/x86_64/src/rej_uniform_avx2_asm.S
jakemas: Here's the reasoning:

The scalar loop bumps `pos += 3` on **every** sample, but `ctr` only increments on an *accepted* one (the `pos` advance happens before the `>= Q` reject check). So in a rejection-heavy tail, `pos` keeps climbing while `ctr` stalls.

Concretely: the buffer is 840 bytes = at most 280 three-byte samples, and we need 256 acceptances — so we can only absorb ~24 rejections before running dry with `ctr` still `< 256`. We can also enter the scalar loop with `pos` as high as 832 (the main loop exits after `pos += 24` with `pos` having been `<= 808`). From there, if the tail keeps rejecting, `pos` walks 832 → 835 → 838 and at 838 the next 3-byte read would touch `buf[838..840]`, one past the end of the 840-byte buffer. The `cmpl $SCALAR_POS_BOUND, pos; ja done` is what stops that.

It's also load-bearing in the proof: the loop exit is disjunctive (hit 256 coefficients *or* exhaust the buffer), and the memory-safety argument for the 3-byte load depends on `pos <= 837`, which this branch establishes. Removing it reintroduces the over-read on adversarial input and breaks the MEMSAFE proof.

jakemas: Worth noting the removed intrinsics had the identical guard as its scalar loop condition:

```c
while (ctr < MLDSA_N && pos <= MLD_AVX2_REJ_UNIFORM_BUFLEN - 3)
```

which is exactly the `pos <= 837` that `cmpl $SCALAR_POS_BOUND, pos; ja done` implements. The asm didn't add the check, it preserves what was already there.

The header comment on that file also spells out why it has to be there:

> The pqcrystals implementation assumes a buffer that is 8 bytes larger as the first loop overreads by 8 bytes that are then discarded. We instead do not pad the buffer and do not overread.

So the bound is load-bearing by design: it's what lets us run on the unpadded 840 byte buffer without over-reading.

dev/x86_64/src/rej_uniform_avx2_asm.S
jakemas: done

dev/x86_64/src/rej_uniform_avx2_asm.S
jakemas: done

dev/x86_64/src/rej_uniform_avx2_asm.S
jakemas: done

dev/x86_64/src/rej_uniform_avx2_asm.S
jakemas: done

dev/x86_64/src/rej_uniform_avx2_asm.S
jakemas: done

proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml
jakemas: changed

jakemas: yes, done

mldsa/src/native/x86_64/src/arith_native_x86_64.h
jakemas: done

mldsa/src/native/x86_64/meta.h
jakemas: done

scripts/autogen
jakemas: done

.mcp.json
mkannwischer, line 1: Remove this file

.claude/settings.json
mkannwischer, line 1: Remove this file

proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml
```diff
@@ -0,0 +1,7182 @@
...
+   - pivot_lemma.ml          — PIVOT_VAL_EQ
+   - memory_conjunct_lemma.ml — MEMORY_CONJUNCT_CLOSURE
+   - case_b_close.ml         — VAL_RCX_ADD3, VAL_RCX_ADD3_ZX
+
+   3 remaining CHEAT_TACs (all in the ACCEPT i+1=K offset-exit arm):
+   - count-exit branch: curlen+1=256 case (trivial closure, similar to Case A)
+   - Case A: offset-exit with curlen+1=256
+   - Case B: offset-exit with curlen+1<256 (the interesting case —
+     interactively validated via case_b_script.ml with 0 CHEATs,
+     but file-load has subtle seqapply interaction — see reject_progress.md)
```
mkannwischer: What is this about?

proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml
```diff
@@ -0,0 +1,7182 @@
...
+   Main proof uses MATCH_MP_TAC SCALAR_BODY_LEMMA; the wiring is verified
+   working at mldsa_rej_uniform.ml:1939.
+
+   Status: structural proof loads in ~15s (down from 1hr) after extracting
+   PIVOT_VAL_EQ.
+
+   Dependencies (must be loaded BEFORE this file):
+   - pivot_lemma.ml          — PIVOT_VAL_EQ
+   - memory_conjunct_lemma.ml — MEMORY_CONJUNCT_CLOSURE
+   - case_b_close.ml         — VAL_RCX_ADD3, VAL_RCX_ADD3_ZX
```
mkannwischer: These files do not exist here. Maybe outdaed and shoul be removed?

mkannwischer (changes requested):
Thanks @jakemas. I think there are some artifacts that need to be cleaned up.

.mcp.json
jakemas: Thank you!

.claude/settings.json
jakemas: Thank you!

proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml
jakemas: Left over notes, removed!

proofs/hol_light/x86_64/proofs/rej_uniform_avx2_asm.ml
jakemas: removed, thanks!

mkannwischer (approved):
Thanks @jakemas! This now looks good to me.

@hanno-becker, could you also take another look, please?

jakemas:
After rebasing CI was being flaky on CBMC proof, Claude's note:

**CBMC `rej_uniform_native_x86_64` timeout — root cause & fix**

The `rej_uniform_native_x86_64` CBMC proof was intermittently failing in CI (flipping between ML-DSA-65 and ML-DSA-87 across runs). It is **not** a property violation — it was a **per-proof timeout** (`Timeout | TIMEOUT` against the 1800s CI limit).

**Diagnosis (local):** the proof is correct but was solving in ~**1613s** with the bare `--smt2` (default Z3) backend — ~90% of the 1800s cap, so it tipped over depending on runner load/platform. This matches the note in `proofs/cbmc/lib/z3_smt_only` about Z3's initial *simplify* tactic causing solve-time divergence across platforms.

**Fix:** switch this proof to the existing `z3_smt_only` solver wrapper (already used by the other heavy array proofs, e.g. `polyvecl_pointwise_acc_montgomery`, `poly_invntt_tomont_c`), which skips Z3's initial simplify tactic:

```make
CBMCFLAGS=--external-smt2-solver $(PROOF_ROOT)/lib/z3_smt_only --z3
```

**Result (local, ML-DSA-65):** `VERIFICATION SUCCESSFUL`, **249 properties, 0 failures**, solve time **~10s** (down from ~1613s). The full contract is still checked (postcondition `array_bound`, preconditions on `mld_rej_uniform_avx2_asm`, pointer/overflow safety, assigns/frees). No CI-wide timeout change needed.

test/bench/bench_components_mldsa.c
```diff
@@ -97,6 +97,15 @@ static int bench(void)
   /* polyz_unpack */
   BENCH("polyz_unpack", mld_polyz_unpack(&poly_out, (const uint8_t *)data0))
 
+  /* rej_uniform (via poly_uniform) */
```
hanno-becker, line 100: Nit: You can just use `data0` here as in the other tests?

hanno-becker (approved):
LGTM, thanks a lot @jakemas. What a proof.

PR #1015, opened 2026-04-03
lowram: Compute h incrementally in signing

mkannwischer:
Instead of allocating a full polyveck for h in attempt_signature_generation,
compute cs2, ct0, and hints one polynomial at a time using scratch polys.

This eliminates the polyveck h from the yh_u union, replacing
mld_pack_sig_c_h with incremental packing via mld_pack_sig_c and
mld_pack_sig_h_poly.
This is a prerequisite for also eliminating y in a follow up.

- Depends on #1002 
- Hoisted out from #1005


mldsa/src/packing.c
```diff
@@ -100,75 +100,54 @@ void mld_unpack_sk(uint8_t rho[MLDSA_SEEDBYTES], uint8_t tr[MLDSA_TRBYTES],
...
-      if (h->vec[i].coeffs[j] != 0 && k < number_of_hints)
-      {
-        /* The enclosing if condition AND the loop invariant infer  */
-        /* that k < MLDSA_OMEGA, so writing to sig[k] is safe and k */
-        /* can be incremented.                                      */
-        sig[k++] = (uint8_t)j;
-      }
+      /* The enclosing if condition AND the loop invariant infer   */
+      /* that n < MLDSA_OMEGA, so writing to sig_h[n] is safe and */
+      /* n can be incremented.                                     */
```
hanno-becker: I think this can be removed while at it. I don't think it adds much value, and it is out of line with the verbosity of comments elsewhere.

mldsa/src/sign.c
```diff
@@ -795,15 +718,34 @@ __contract__(
...
   {
-    ret = MLD_ERR_FAIL; /* reject */
-    goto cleanup;
+    unsigned int hints = mld_poly_make_hint(z, &w0->vec[k], &w1->vec[k]);
+    if (n + hints > MLDSA_OMEGA)
+    {
+      ret = MLD_ERR_FAIL; /* reject */
+      goto cleanup;
+    }
+    mld_pack_sig_h_poly(sig, z, k, n);
```
hanno-becker, line 745: Not a request for rework at this point, but it seems that `z` is ephemeral and one could generate the hints directly into the signature?

mkannwischer: Yes, but that buffer is needed in the previous loop anyway, so it won't save memory to eliminate here.

mldsa/src/sign.c
```diff
@@ -795,15 +718,34 @@ __contract__(
...
+            memory_slice(z, sizeof(mld_poly)),
+            memory_slice(sig, MLDSA_CRYPTO_BYTES))
+    invariant(k <= MLDSA_K)
+    invariant(n <= MLDSA_OMEGA)
+    decreases(MLDSA_K - k)
+  )
   {
-    ret = MLD_ERR_FAIL; /* reject */
-    goto cleanup;
+    unsigned int hints = mld_poly_make_hint(z, &w0->vec[k], &w1->vec[k]);
```
hanno-becker, line 739: Note that we track the number of hints twice: `mld_poly_make_hint` returns the number of hints, and `sig_h_poly` also computes it, but doesn't validate again that the two line up. This is not a problem per-se, but I was confused reviewing `mld_pack_sig_h_poly` reading that it receives the tally of hints but doesn't return the updated one.

That would naturally go away if one consolidated `make_hint` and `pack_sig_h_poly`.

mkannwischer: Agreed - this also confused me.
I will give it a try to consolidate it.

mldsa/src/sign.c
```diff
@@ -753,37 +651,62 @@ __contract__(
...
+    mld_poly_invntt_tomont(z);
 
-  /* Compute hints for w1 */
-  mld_polyveck_pointwise_poly_montgomery_t0(h, cp, t0hat, z);
-  mld_polyveck_invntt_tomont(h);
-  mld_polyveck_reduce(h);
+    /* TODO: Remove this workaround for CBMC performance issues */
+    *t = w0->vec[k];
+    mld_poly_sub(t, z);
+    mld_poly_reduce(t);
```
hanno-becker, line 676: Nit: While the workaround is needed, one may as well assign `w0->vec[k]` only once instead of twice.

hanno-becker (approved):
Thank you @mkannwischer, this looks sound and pretty clean. I left some smaller notes but leave it to your discretion whether to act on them at this point or not.

mldsa/src/packing.c
mkannwischer: Removed.

mldsa/src/sign.c
mkannwischer: I don't see how you can do that without requiring another scratch polynomial (note that t is used for something else in between the two assignments).
I will open an issue to track eliminating the workaround: https://github.com/pq-code-package/mldsa-native/issues/1026

mldsa/src/sign.c
mkannwischer: Tracking issue: https://github.com/pq-code-package/mldsa-native/issues/1027

PR #1019, opened 2026-04-05
lowram: Stream matrix A element-by-element to reduce memory

mkannwischer:
Replace the row-level matrix buffer (mld_polyvecl) with a single-poly
buffer in REDUCE_RAM mode. In the lazy path, matrix elements A[k][l]
are sampled on demand one at a time, and the matrix-vector product
accumulates element-by-element instead of row-by-row.

Restructure polymat into eager/lazy variants following the same pattern
as s1hat/s2hat/t0hat:
- mld_polymat_eager: stores full K x L matrix
- mld_polymat_lazy: stores rho + single poly_buffer + tmp
- mld_polyvec_matrix_expand_eager/_lazy: separate implementations
- mld_polyvec_matrix_pointwise_montgomery_eager/_lazy: separate
  implementations with CBMC contracts only on the eager variants

Move all polymat-related code from polyvec.h/polyvec.c into
polyvec_lazy.h/polyvec_lazy.c.

- Hoisted out from #1005 
- Depends on #1015


mldsa/src/polyvec_lazy.h
```diff
@@ -183,27 +207,158 @@ static MLD_INLINE void mld_sk_t0hat_get_poly_lazy(mld_poly *buf,
 }
 #endif /* MLD_CONFIG_REDUCE_RAM || MLD_UNIT_TEST */
 
+/* polymat */
+
+/* Eager: precompute and store full matrix */
+typedef struct
+{
+  mld_polyvecl vec[MLDSA_K];
+} mld_polymat_eager;
```
mkannwischer, line 232: This needs to be guarded, otherwise this will fail on AVR.

hanno-becker: Why?

mkannwischer: The compiler failed for me saying that you can't declare a struct larger than 32 KiB even if the struct is not used.

mldsa/src/polyvec_lazy.c
```diff
@@ -0,0 +1,240 @@
...
+void mld_polyvec_matrix_pointwise_montgomery_lazy(mld_polyveck *t,
+                                                  mld_polymat_lazy *mat,
+                                                  const mld_polyvecl *v)
+{
+  unsigned int i, l;
+
+  for (i = 0; i < MLDSA_K; ++i)
+  {
+    const mld_poly *a_kl = mld_polymat_get_poly_lazy(mat, i, 0);
+    mld_poly_pointwise_montgomery(&t->vec[i], a_kl, &v->vec[0]);
```
hanno-becker: The CBMC spec+proof for `mld_polymat_get_poly_lazy` may be tricky because we return part of `mat`. It may be easier to remove the return value and have `mld_polymat_get_poly_lazy` write the desired element to `mat->tmp` (perhaps renamed to `cur` or `out`)?

mldsa/src/polyvec.h
hanno-becker, line 100: This also means that the corresponding backend functionality can be dropped in the reduced-RAM build?

mldsa/src/polyvec_lazy.c
```diff
@@ -0,0 +1,240 @@
...
+MLD_INTERNAL_API
+void mld_polyvec_matrix_expand_eager(mld_polymat_eager *mat,
+                                     const uint8_t rho[MLDSA_SEEDBYTES])
+{
+  unsigned int i, j;
+  /*
+   * We generate four separate seed arrays rather than a single one to work
+   * around limitations in CBMC function contracts dealing with disjoint slices
+   * of the same parent object.
+   */
```
hanno-becker: This comment appears outdated and can be removed on this occasion?

mldsa/src/polyvec_lazy.c
```diff
@@ -0,0 +1,240 @@
...
+  __loop__(
+    assigns(j, object_whole(seed_ext))
+    invariant(j <= 4)
+    decreases(4 - j)
+  )
+  {
+    mld_memcpy(seed_ext[j], rho, MLDSA_SEEDBYTES);
+  }
+
+#if !defined(MLD_CONFIG_SERIAL_FIPS202_ONLY) && !defined(MLD_CONFIG_REDUCE_RAM)
```
hanno-becker: Now we only use this for `!MLD_CONFIG_REDUCE_RAM`, so it can be simplified to `#if !defined(MLD_CONFIG_SERIAL_FIPS202_ONLY)`?

mldsa/src/polyvec_lazy.c
```diff
@@ -0,0 +1,240 @@
...
+  for (i = 0; i < MLDSA_K; ++i)
+  __loop__(
+    assigns(i, memory_slice(t, sizeof(mld_polyveck)))
+    invariant(i <= MLDSA_K)
+    invariant(forall(k0, 0, i,
+                     array_abs_bound(t->vec[k0].coeffs, 0, MLDSA_N, MLDSA_Q)))
+    decreases(MLDSA_K - i)
+  )
+  {
+    const mld_polyvecl *row = mld_polymat_get_row_eager(mat, i);
```
hanno-becker: There's little/no value in keeping this as a separate function now that we distinguish between full sampling and element-by-element

mldsa/src/polyvec_lazy.c
hanno-becker: You already have a separate pointer for the output poly, so my suggestion just reduces to using this directly rather than returning it.

mldsa/src/polyvec_lazy.c
```diff
@@ -0,0 +1,240 @@
...
+
+  for (i = 0; i < MLDSA_K; ++i)
+  {
+    const mld_poly *a_kl = mld_polymat_get_poly_lazy(mat, i, 0);
+    mld_poly_pointwise_montgomery(&t->vec[i], a_kl, &v->vec[0]);
+    for (l = 1; l < MLDSA_L; ++l)
+    {
+      a_kl = mld_polymat_get_poly_lazy(mat, i, l);
+      mld_poly_pointwise_montgomery(&mat->tmp, a_kl, &v->vec[l]);
+      mld_poly_add(&t->vec[i], &mat->tmp);
```
hanno-becker: Maybe leave a TODO note that we could get rid of `mat->tmp` here by multiplying in-place, assuming we strengthen the corresponding CBMC and HOL Light specs.

mldsa/src/polyvec_lazy.h
```diff
@@ -183,27 +207,158 @@ static MLD_INLINE void mld_sk_t0hat_get_poly_lazy(mld_poly *buf,
...
+  MLD_ALIGN uint8_t seed_ext[MLD_ALIGN_UP(MLDSA_SEEDBYTES + 2)];
+  mld_memcpy(seed_ext, mat->rho, MLDSA_SEEDBYTES);
+  seed_ext[MLDSA_SEEDBYTES + 0] = (uint8_t)l;
+  seed_ext[MLDSA_SEEDBYTES + 1] = (uint8_t)k;
+  mld_poly_uniform(&mat->poly_buffer, seed_ext);
+  mld_poly_permute_bitrev_to_custom_optional(&mat->poly_buffer);
+  /* @[FIPS204, Section 3.6.3] Destruction of intermediate values. */
+  mld_zeroize(seed_ext, sizeof(seed_ext));
+  return &mat->poly_buffer;
+}
```
hanno-becker: Do we need this function? Can we inline it into its call-site and share its logic with that of eager matrix expansion?

hanno-becker (changes requested):
I'm worried about the complexity we're building up here. There are (too) many small functions which make the code very difficult to oversee and which, I believe, aren't all necessary -- see comments. Let's see if we can clean this up a bit more before merging.

In principle, I'm OK with the optimization, though I don't think it's necessary to meet a 32K RAM target -- assuming all the other optimizations get merged, it seems like the row-by-row expansion is already enough? The latter is less intrusive and more performant since it allows you to still use the faster vector-vector scalar product.

mkannwischer:
> In principle, I'm OK with the optimization, though I don't think it's necessary to meet a 32K RAM target -- assuming all the other optimizations get merged, it seems like the row-by-row expansion is already enough? The latter is less intrusive and more performant since it allows you to still use the faster vector-vector scalar product.

If speed is a goal, the first optimization in `REDUCE_RAM` mode to drop is the recomputation of `y` (https://github.com/pq-code-package/mldsa-native/pull/1031). That costs an L-polyvec and saves a lot of Keccak invocations inside of the main signing loop. I'm fairly sure this outweighs the gains made by vector-vector polymul on most platforms.

mldsa/src/polyvec_lazy.c
mkannwischer: okay, done.

mldsa/src/polyvec.h
mkannwischer: Yes, done.

mldsa/src/polyvec_lazy.c
mkannwischer: removed.

mldsa/src/polyvec_lazy.c
mkannwischer: Yes, done.

mldsa/src/polyvec_lazy.c
mkannwischer: yes, agreed. Removed.

mldsa/src/polyvec_lazy.c
mkannwischer: added.

mldsa/src/polyvec_lazy.h
mkannwischer: We can inline this function yes. I do not understand what you mean by share its logic with that of the eager matrix expansion. I have inlined the function for now.

hanno-becker: Well the core expansion logic is the same

hanno-becker: In more detail, I was wondering if it makes sense to hoist out the logic of
```c
seed_ext[MLDSA_SEEDBYTES + 0] = (uint8_t)l;
seed_ext[MLDSA_SEEDBYTES + 1] = (uint8_t)k;
mld_poly_uniform(&mat->poly_buffer, seed_ext);
mld_poly_permute_bitrev_to_custom_optional(&mat->poly_buffer);
```
into a (`static inline`) helper function that can be called from both the eager expansion (for the residual entries) and from the lazy expansion. This would require removing `mld_polymat_permute_bitrev_to_custom` and instead call `mld_poly_permute_bitrev_to_custom_optional` explicitly in the 4x-batch expansion. But thay may actually be simpler than what we have at the moment, seeing that we need a vec- and a mat-version of the conditional permute.

It may not be worth it to specify this function in CBMC -- TBD -- but I think it would help code clarity.

WDYT?

hanno-becker: So, e.g.
```c
  for (i = 0; i < MLDSA_K; ++i)
  {
    seed_ext[MLDSA_SEEDBYTES + 0] = 0;
    seed_ext[MLDSA_SEEDBYTES + 1] = (uint8_t)i;
    mld_poly_uniform(&mat->cur, seed_ext);
    mld_poly_permute_bitrev_to_custom_optional(&mat->cur);
    mld_poly_pointwise_montgomery(&t->vec[i], &mat->cur, &v->vec[0]);

    for (l = 1; l < MLDSA_L; ++l)
    {
      seed_ext[MLDSA_SEEDBYTES + 0] = (uint8_t)l;
      seed_ext[MLDSA_SEEDBYTES + 1] = (uint8_t)i;
      mld_poly_uniform(&mat->cur, seed_ext);
      mld_poly_permute_bitrev_to_custom_optional(&mat->cur);
      /* TODO: if mld_poly_pointwise_montgomery's CBMC and HOL Light specs
       * are strengthened to permit aliasing, the product can be written
       * in place into mat->cur and the separate mat->tmp field dropped. */
      mld_poly_pointwise_montgomery(&mat->tmp, &mat->cur, &v->vec[l]);
      mld_poly_add(&t->vec[i], &mat->tmp);
    }
    mld_poly_reduce(&t->vec[i]);
  }
```
would turn into
```c
  for (i = 0; i < MLDSA_K; ++i)
  {
    mld_polymat_expand(&mat->cur, seed_ext, 0, i);
    mld_poly_pointwise_montgomery(&t->vec[i], &mat->cur, &v->vec[0]);

    for (l = 1; l < MLDSA_L; ++l)
    {
      mld_polymat_expand(&mat->cur, seed_ext, l, i);
      /* TODO: if mld_poly_pointwise_montgomery's CBMC and HOL Light specs
       * are strengthened to permit aliasing, the product can be written
       * in place into mat->cur and the separate mat->tmp field dropped. */
      mld_poly_pointwise_montgomery(&mat->tmp, &mat->cur, &v->vec[l]);
      mld_poly_add(&t->vec[i], &mat->tmp);
    }
    mld_poly_reduce(&t->vec[i]);
  }
```
and in the eager expansion, we'd replace
```c
  /* Entries omitted by the batch-sampling are sampled individually. */
  while (i < MLDSA_K * MLDSA_L)
  __loop__(
    assigns(i, object_whole(seed_ext), memory_slice(mat, sizeof(mld_polymat_eager)))
    invariant(i <= MLDSA_K * MLDSA_L)
    /* vectors 0 .. i / MLDSA_L are completely sampled */
    invariant(forall(k1, 0, i / MLDSA_L, forall(l1, 0, MLDSA_L,
      array_bound(mat->vec[k1].vec[l1].coeffs, 0, MLDSA_N, 0, MLDSA_Q))))
    /* last vector is sampled up to i % MLDSA_L */
    invariant(forall(k2, i / MLDSA_L, i / MLDSA_L + 1, forall(l2, 0, i % MLDSA_L,
      array_bound(mat->vec[k2].vec[l2].coeffs, 0, MLDSA_N, 0, MLDSA_Q))))
    decreases(MLDSA_K * MLDSA_L - i)
  )
  {
    uint8_t x = (uint8_t)(i / MLDSA_L);
    uint8_t y = (uint8_t)(i % MLDSA_L);
    mld_poly *this_poly = &mat->vec[i / MLDSA_L].vec[i % MLDSA_L];

    seed_ext[0][MLDSA_SEEDBYTES + 0] = y;
    seed_ext[0][MLDSA_SEEDBYTES + 1] = x;

    mld_poly_uniform(this_poly, seed_ext[0]);
    i++;
  }
```
by
```c
  /* Entries omitted by the batch-sampling are sampled individually. */
  while (i < MLDSA_K * MLDSA_L)
  __loop__(
    assigns(i, object_whole(seed_ext), memory_slice(mat, sizeof(mld_polymat_eager)))
    invariant(i <= MLDSA_K * MLDSA_L)
    /* vectors 0 .. i / MLDSA_L are completely sampled */
    invariant(forall(k1, 0, i / MLDSA_L, forall(l1, 0, MLDSA_L,
      array_bound(mat->vec[k1].vec[l1].coeffs, 0, MLDSA_N, 0, MLDSA_Q))))
    /* last vector is sampled up to i % MLDSA_L */
    invariant(forall(k2, i / MLDSA_L, i / MLDSA_L + 1, forall(l2, 0, i % MLDSA_L,
      array_bound(mat->vec[k2].vec[l2].coeffs, 0, MLDSA_N, 0, MLDSA_Q))))
    decreases(MLDSA_K * MLDSA_L - i)
  )
  {
    uint8_t x = (uint8_t)(i / MLDSA_L);
    uint8_t y = (uint8_t)(i % MLDSA_L);
    mld_polymat_expand(&mat->vec[x].vec[y],  seed_ext[0], x, y);
    i++;
  }
```

mkannwischer: Yes, that seems to work well and makes the code a little more readable.
I just pushed another commit (the fixup commits should all be squashed into the first commit prior to merge).

Could you take another look, please?

hanno-becker (approved):
Thank you very much @mkannwischer for the initiative and for the various cleanups. It looks good to me now 🙏

PR #1025, opened 2026-04-07
lowmem: Unpack z lazily in verification

mkannwischer:
Introduce mld_zvec following the lazy polyvec pattern (eager / lazy
variants with #define dispatch on MLD_CONFIG_REDUCE_RAM):

  - mld_zvec_init: in eager mode unpacks the full polyvecl, performs the
    polyvecl-wide infinity-norm bound check, and NTTs in place. In lazy
    mode it just stores a pointer to the packed signature bytes.
  - mld_zvec_get_poly: in eager mode copies a single polynomial from
    the precomputed vector. In lazy mode unpacks one polynomial,
    performs the per-poly infinity-norm bound check, and NTTs into
    the caller-provided buffer.

The norm check thus moves out of mld_sign_verify_internal into the
zvec init / get_poly accessors, so the verify body no longer has to
sequence chknorm explicitly.

Add a fused matrix-vector helper
mld_polyvec_matrix_pointwise_montgomery_zvec used by verify:

  - The eager variant is a thin wrapper around the existing
    mld_polyvec_matrix_pointwise_montgomery_eager (z is already NTT'd
    by mld_zvec_init).
  - The lazy variant streams z via mld_zvec_get_poly_lazy and generates
    the matrix on-the-fly column-by-column,
    accumulating A[*,l] * z[l] into w.

In REDUCE_RAM mode this avoids holding the full unpacked polyvecl z
in memory at once, reducing verify allocation by 2-5 KiB per parameter
set.

- Depends on #1019 
- Depends on #1007
- Depends on #1065
- Hoisted out from #1005


mkannwischer:
We merged #1095 instead which is much cleaner.

PR #1028, opened 2026-04-08
sign: Consolidate make_hint and pack_sig_h_poly 

### LINKED ISSUE #1027: Signing: Combine `mld_poly_make_hint` and `mld_pack_sig_h_poly`
We currently need to keep track of the number of hints in both `make_hints` and `pack_sig_h` which complicates things unnecessarily and is somewhat confusing. Combining the two seems more natural.

See https://github.com/pq-code-package/mldsa-native/pull/1015#discussion_r3049108306

mkannwischer:
- Resolves #1027
- Depends on #1015


mldsa/src/packing.c
```diff
@@ -132,19 +140,21 @@ void mld_pack_sig_h_poly(uint8_t sig[MLDSA_CRYPTO_BYTES], const mld_poly *h,
...
-    /* number of hints being less than OMEGA, assuming h is valid. */
-    /* In mldsa-native, we check this explicitly to ease proof of  */
-    /* type safety.                                                */
-    if (h->coeffs[j] != 0 && n < MLDSA_OMEGA)
+    const unsigned int hint_bit = mld_make_hint(a0->coeffs[j], a1->coeffs[j]);
+    if (hint_bit)
     {
+      if (n == MLDSA_OMEGA)
+      {
+        return MLDSA_OMEGA + 1;
```
hanno-becker: I'd prefer a different sentinel -- why not stick with a negative value to signal an error?

hanno-becker: Yet an alternative approach would be to keep counting the hints and return their exact number, but only write them so long as `n < MLDSA_OMEGA`.

mkannwischer: I think just returning an error is a good idea. 
The number of hints is not needed in case it reaches MLDSA_OMEGA anyway.

hanno-becker (changes requested):
This overall looks good, but I'd prefer a clean error signal when there are too many hints, see comment.

mldsa/src/packing.c
mkannwischer: Fixed.

mldsa/src/sign.c
```diff
@@ -728,22 +728,18 @@ __contract__(
   n = 0;
   for (k = 0; k < MLDSA_K; k++)
   __loop__(
-    assigns(k, n,
-            memory_slice(z, sizeof(mld_poly)),
-            memory_slice(sig, MLDSA_CRYPTO_BYTES))
+    assigns(k, n, ret, memory_slice(sig, MLDSA_CRYPTO_BYTES))
```
jakemas: Nice side effect: `z` is no longer reused as scratch in the hint loop, so its lifetime can end before hint packing begins. Perhaps relevant for the low-memory implementation since that's one fewer `mld_poly` (~1 KiB ?) on the stack during this phase.

jakemas (approved):

mkannwischer:
Rebased on top of main. Please take another look @hanno-becker.

mldsa/src/packing.h
```diff
@@ -108,29 +108,41 @@ __contract__(
...
  *              The hint section of sig must be zeroed before the first call.
  *
  * Arguments:   - uint8_t sig[]: byte array containing signature
- *              - const mld_poly *h: pointer to hint polynomial (0/1 coeffs)
+ *              - const mld_poly *a0: pointer to low part of input polynomial
+ *              - const mld_poly *a1: pointer to high part of input polynomial
  *              - unsigned int k: index of polynomial in vector (0..K-1)
  *              - unsigned int n: total number of hints written so far
+ *
+ * Returns:     - the updated running tally of hints (non-negative, at most
```
hanno-becker: Nit: Unify the language, `total number of hints` vs. `running tally of hints`

mldsa/src/packing.h
```diff
@@ -108,29 +108,41 @@ __contract__(
...
 
-#define mld_pack_sig_h_poly MLD_NAMESPACE_KL(pack_sig_h_poly)
+#define mld_make_pack_sig_h_poly MLD_NAMESPACE_KL(make_pack_sig_h_poly)
 /*************************************************
- * Name:        mld_pack_sig_h_poly
+ * Name:        mld_make_pack_sig_h_poly
  *
- * Description: Pack hints for one polynomial into the hint section of sig.
+ * Description: Compute hints for one polynomial of (w0, w1) and pack them
+ *              into the hint section of sig.
```
hanno-becker: Nit: Provide a reference or more context, otherwise it's not clear what `(w0,w1)` means here, and one wonders if it's a typo with the `(a0,a1)` arguments.

mldsa/src/packing.h
```diff
@@ -108,29 +108,41 @@ __contract__(
...
  *              Must be called once per polynomial in order k = 0, ..., K-1.
```
hanno-becker: And the total number of hints need to be passed through?

mldsa/src/packing.h
```diff
@@ -108,29 +108,41 @@ __contract__(
...
  *              The hint section of sig must be zeroed before the first call.
```
hanno-becker: Does it make sense to make this part of the call for `k=0`?

mldsa/src/sign.c
```diff
@@ -744,22 +744,18 @@ __contract__(
...
-    if (n + hints > MLDSA_OMEGA)
+    ret = mld_make_pack_sig_h_poly(sig, &w0->vec[k], &w1->vec[k], k, n);
+    if (ret == MLD_ERR_FAIL)
     {
-      ret = MLD_ERR_FAIL; /* reject */
       goto cleanup;
     }
-    mld_pack_sig_h_poly(sig, z, k, n);
-    n += hints;
+    n = (unsigned int)ret;
```
hanno-becker: Add a safety comment here?

mldsa/src/packing.h
```diff
@@ -108,29 +108,41 @@ __contract__(
...
+ *                MLDSA_OMEGA) on success;
+ *              - MLD_ERR_FAIL if writing all hints for this polynomial
+ *                would exceed MLDSA_OMEGA. In this case the caller must
+ *                reject the signature.
  **************************************************/
 MLD_INTERNAL_API
-void mld_pack_sig_h_poly(uint8_t sig[MLDSA_CRYPTO_BYTES], const mld_poly *h,
-                         unsigned int k, unsigned int n)
+MLD_MUST_CHECK_RETURN_VALUE
+int mld_make_pack_sig_h_poly(uint8_t sig[MLDSA_CRYPTO_BYTES],
```
hanno-becker: Nit: I'd prefer to stick with uniform `mld_pack_sig_foo`

mldsa/src/packing.c
```diff
@@ -147,19 +154,21 @@ void mld_pack_sig_h_poly(uint8_t sig[MLDSA_CRYPTO_BYTES], const mld_poly *h,
...
-    /* type safety.                                                */
-    if (h->coeffs[j] != 0 && n < MLDSA_OMEGA)
+    const unsigned int hint_bit = mld_make_hint(a0->coeffs[j], a1->coeffs[j]);
+    if (hint_bit)
     {
+      if (n == MLDSA_OMEGA)
+      {
+        return MLD_ERR_FAIL;
+      }
       sig_h[n] = (uint8_t)j;
```
hanno-becker: Add safety comment while at it?

mldsa/src/packing.c
```diff
@@ -147,19 +154,21 @@ void mld_pack_sig_h_poly(uint8_t sig[MLDSA_CRYPTO_BYTES], const mld_poly *h,
...
       n++;
     }
   }
   /* Record the running tally into the correct slot for this     */
   /* polynomial in the final K bytes.                            */
   sig_h[MLDSA_OMEGA + k] = (uint8_t)n;
```
hanno-becker: Add safety comment while at it?

mldsa/src/packing.h
```diff
@@ -108,29 +108,41 @@ __contract__(
...
   requires(memory_no_alias(sig, MLDSA_CRYPTO_BYTES))
-  requires(memory_no_alias(h, sizeof(mld_poly)))
+  requires(memory_no_alias(a0, sizeof(mld_poly)))
+  requires(memory_no_alias(a1, sizeof(mld_poly)))
   requires(k < MLDSA_K)
   requires(n <= MLDSA_OMEGA)
-  requires(array_bound(h->coeffs, 0, MLDSA_N, 0, 2))
   assigns(memory_slice(sig, MLDSA_CRYPTO_BYTES))
+  ensures(return_value == MLD_ERR_FAIL ||
+          (return_value >= 0 && return_value <= MLDSA_OMEGA))
```
hanno-becker: This weak version of the spec works. But, as for selected CT helpers, we could consider getting a bit more out of CBMC here and strengthening the spec. Handling `return_value >= n` should be trivial. But we could also consider expressing that the slice of bytes written is ascending. This isn't needed by the caller, but still an important correctness property.

hanno-becker (changes requested):
Overall, this looks good and like a good improvement!

I left a few smaller nits on documentation and naming.

mldsa/src/packing.h
mkannwischer: I went for the trivial option now. I briefly tried the latter, but couldn't get it through CBMC quickly. We can do that in a follow up, but I think it would be better to put this effort into unpack_hint where correctness is much harder to test using testvectors.
There a correctness spec would be much more valuable in my opinion.

mldsa/src/packing.c
mkannwischer: done.

mldsa/src/packing.c
mkannwischer: done.

mldsa/src/packing.h
mkannwischer: done.

mldsa/src/packing.h
mkannwischer: done.

mldsa/src/packing.h
mkannwischer: I removed that. It can just be read from the signature.

mldsa/src/packing.h
mkannwischer: Renamed.

mldsa/src/packing.h
mkannwischer: Done.

mldsa/src/packing.h
hanno-becker: https://github.com/pq-code-package/mldsa-native/issues/1099

mldsa/src/packing.h
```diff
@@ -112,25 +112,49 @@ __contract__(
...
-  requires(memory_no_alias(h, sizeof(mld_poly)))
+  requires(memory_no_alias(w0, sizeof(mld_poly)))
+  requires(memory_no_alias(w1, sizeof(mld_poly)))
   requires(k < MLDSA_K)
-  requires(n <= MLDSA_OMEGA)
-  requires(array_bound(h->coeffs, 0, MLDSA_N, 0, 2))
-  assigns(memory_slice(sig, MLDSA_CRYPTO_BYTES))
+  requires(k == 0 ||
+           sig[MLDSA_CTILDEBYTES + MLDSA_L * MLDSA_POLYZ_PACKEDBYTES +
+               MLDSA_OMEGA + k - 1] <= MLDSA_OMEGA)
```
hanno-becker: These compound offset expressions are increasingly hard to read; we should introduce offset abbreviations in a single place where the layout of the signature is also briefly documented, and then use them throughout.

Here, e.g. `sig[MLD_SIG_OFFSET_HINT_TALLY(k)]`. Which is still a handful, but still shorter, and arguably more digestible, than `MLDSA_CTILDEBYTES + MLDSA_L * MLDSA_POLYZ_PACKEDBYTES + MLDSA_OMEGA + k`.

hanno-becker: This affects other functions as well, though, so we may want to do it in a separate PR.

hanno-becker: https://github.com/pq-code-package/mldsa-native/issues/1103

mldsa/src/sign.c
```diff
@@ -742,31 +742,26 @@ __contract__(
...
+    invariant(k == 0 ||
+              sig[MLDSA_CTILDEBYTES + MLDSA_L * MLDSA_POLYZ_PACKEDBYTES +
+                  MLDSA_OMEGA + k - 1] <= MLDSA_OMEGA)
     decreases(MLDSA_K - k)
   )
   {
-    unsigned int hints = mld_poly_make_hint(z, &w0->vec[k], &w1->vec[k]);
-    if (n + hints > MLDSA_OMEGA)
+    ret = mld_pack_sig_h_poly(sig, &w0->vec[k], &w1->vec[k], k);
+    if (ret != 0)
```
hanno-becker: Seeing what remains of the loop, it's unclear why the entire hint generation is not hoisted into a helper.

hanno-becker (commented):
Thank you @mkannwischer, it's getting neater with every change. 

Taking a step back, now hint-generation is a plain `k=0,..,MLDSA_K-1` loop over the hint-pack-k function. It begs the question why we don't move that loop into the hint packing function and have a single function doing all the hint packing. WDYT?

mldsa/src/packing.h
```diff
@@ -81,29 +81,34 @@ __contract__(
...
- *              - const mld_poly *h: pointer to hint polynomial (0/1 coeffs)
- *              - unsigned int k: index of polynomial in vector (0..K-1)
- *              - unsigned int n: total number of hints written so far
+ *              - const mld_polyveck *w0: pointer to low part of input vector
+ *              - const mld_polyveck *w1: pointer to high part of input vector
+ *
+ * Returns:     - 0 on success;
+ *              - MLD_ERR_FAIL if the total number of hints exceeds
+ *                MLDSA_OMEGA. In this case the caller must reject the
+ *                signature.
```
hanno-becker: This currently says nothing about the state of the hint buffer upon failure -- which is partially written. We should consider a) clearing upon failure, or b) explicitly stating that the hint buffer is left in incomplete state upon failure and should be cleared.

A broader question is whether the temporary writes to the signature buffer pose a risk in case the signature buffer is attacker owned (e.g., the trusted signing process signs directly into a buffer shared with the malicious attacker process, who continually polls what is being written there). In that case, our current remedy of clearing the signature buffer at the end of the entire routine, would be insufficient. Instead, one would need a copy. It feels like this should rather be the caller's decision, though, but we may still want to clarify this in the top-level API.

mkannwischer: The rejection condition (|hints| > OMEGA) is not needed for security - it's merely a way to decrease the signature size by a couple of bytes. The hints are public regardless, so we are not leaking anything secret. Also note that the signature buffer has been partially written even before this function is called. 
I have adjusted the documentation of pack_sig_h to state that the buffer will be partially written.

If you think that overwriting the signature buffer is a problem, we should discuss that separately. Can you open an issue? [Here](https://github.com/pq-code-package/mldsa-native/blob/2c5f4ad299e5dd4dae8d1ff6e50df1873a6f0a2d/mldsa/src/sign.c#L722) we are abusing the signature buffer to hold w1 which won't even be part of the final signature. It is, however, public.

hanno-becker: I opened https://github.com/pq-code-package/mldsa-native/issues/1108 to discuss this further.

hanno-becker (approved):
I'm very happy with how this turned out. Many thanks @mkannwischer for your patience and willingness to iterate. LGTM

PR #1029, opened 2026-04-08
aarch64/src/polyz_unpack_table.c: Disable unused tables

flynd:
Only one table is used for each parameter set, so add conditions to remove the unused table from non-shared builds.

This change was part of  PR #1000 but it is really a separate change so I've opened this PR for it instead.


mkannwischer (approved):
Thanks @flynd! LGTM.

PR #1030, opened 2026-04-08
lowram: Per-row t0/t1 computation in keygen 

mkannwischer:
Replace mld_compute_t0_t1_tr_from_sk_components with the per-row
mld_compute_t0k_t1k. Both keygen and pk_from_sk now process one
row at a time, packing t1[k] into pk and t0[k] into sk immediately,
which eliminates the full polyveck allocations for t0 and t1.

To express this cleanly, refactor mld_polyvecl_pointwise_acc_montgomery
to take (mat, k, v) instead of (u, v) and move it into polyvec_lazy
with eager and lazy variants. Add per-row pack helpers
mld_pack_sk_t0 / mld_pack_pk_t1, and split mld_pack_sk_rho_key_tr_s2_t0
to drop the t0 packing.

Drop now-dead mld_polyveck_add, mld_polyveck_pack_t0,
mld_polyveck_power2round and mld_pack_pk along with their CBMC proofs.

Update the affected CBMC proofs.

- Depends on #1019 
- Depends on #1011
- Depends on #1095
- Hoisted out from #1005


mkannwischer:
CBMC proof performance will still need investigation, but I'll do that after the other PRs have been merged.

mldsa/src/sign.c
```diff
@@ -197,50 +195,60 @@ __contract__(
...
+ *              - const mld_polyvecl *s1hat: s1 in NTT domain
+ *              - const mld_poly *s2k: s2 polynomial for row k
  **************************************************/
-MLD_MUST_CHECK_RETURN_VALUE
-static int mld_compute_t0_t1_tr_from_sk_components(
-    mld_polyveck *t0, mld_polyveck *t1, uint8_t tr[MLDSA_TRBYTES],
-    uint8_t pk[MLDSA_CRYPTO_PUBLICKEYBYTES], const uint8_t rho[MLDSA_SEEDBYTES],
-    mld_polyvecl *s1, const mld_polyveck *s2,
-    MLD_CONFIG_CONTEXT_PARAMETER_TYPE context)
+static void mld_compute_t0k_t1k(mld_poly *t0k, mld_poly *t1k, mld_poly *tk,
```
mkannwischer: One could try to in eliminate this function again. I wrote this code a long time ago and back then CBMC couldn't handle it, but it may be worth a try.

mldsa/src/packing.c
```diff
@@ -17,22 +17,10 @@
 
 #if !defined(MLD_CONFIG_NO_KEYPAIR_API)
 MLD_INTERNAL_API
-void mld_pack_pk(uint8_t pk[MLDSA_CRYPTO_PUBLICKEYBYTES],
-                 const uint8_t rho[MLDSA_SEEDBYTES], const mld_polyveck *t1)
+void mld_pack_pk_t1(uint8_t pk[MLDSA_CRYPTO_PUBLICKEYBYTES], unsigned int k,
```
hanno-becker: Nit: This is so trivial that is feels like it should be a `static inline` function in `packing.h`, or a static function in `sign.c`.

mldsa/src/sign.c
```diff
@@ -197,50 +195,60 @@ __contract__(
...
- *              - const mld_polyveck *s2: input s2
+ *              The accumulator scratch tk is reused as t0k by the
+ *              caller (the contract of mld_poly_power2round requires
+ *              t0k == tk).
+ *
+ * Arguments:   - mld_poly *t0k: output t0 polynomial (must alias tk)
+ *              - mld_poly *t1k: output t1 polynomial
+ *              - mld_poly *tk: scratch / accumulator for the matrix row
+ *              - mld_polymat *mat: expanded matrix (sampled lazily in
+ *                                    REDUCE_RAM mode)
```
hanno-becker: This is a bit confusing. If `t0k == tk`, why do we have `tk` at all? The documentation should focus on the things that matter to the caller.

mldsa/src/sign.c
hanno-becker: Yes

mldsa/src/sign.c
```diff
@@ -314,34 +306,56 @@ int mld_sign_keypair_internal(uint8_t pk[MLDSA_CRYPTO_PUBLICKEYBYTES],
...
+  /* NTT s1 in place to use as s1hat */
+  mld_polyvecl_ntt(s1);
+
+  /* Expand matrix and pack rho into pk */
+  mld_polyvec_matrix_expand(mat, rho);
+  mld_memcpy(pk, rho, MLDSA_SEEDBYTES);
+
+  /* Compute t row by row, decompose into t1[k]/t0[k], and pack
+   * immediately into pk and sk. The t0k buffer aliases the accumulator
+   * scratch, as required by mld_poly_power2round. */
```
hanno-becker: > The t0k buffer aliases the accumulator scratch, as required by mld_poly_power2round.

This comment is unnecessary here and more confusing than helpful. Remove it?

mldsa/src/packing.c
```diff
@@ -55,12 +43,22 @@ void mld_pack_sk_s1(uint8_t sk[MLDSA_CRYPTO_SECRETKEYBYTES],
 }
 
 MLD_INTERNAL_API
-void mld_pack_sk_rho_key_tr_s2_t0(uint8_t sk[MLDSA_CRYPTO_SECRETKEYBYTES],
-                                  const uint8_t rho[MLDSA_SEEDBYTES],
-                                  const uint8_t tr[MLDSA_TRBYTES],
-                                  const uint8_t key[MLDSA_SEEDBYTES],
-                                  const mld_polyveck *t0,
-                                  const mld_polyveck *s2)
+void mld_pack_sk_t0(uint8_t sk[MLDSA_CRYPTO_SECRETKEYBYTES], unsigned int k,
```
hanno-becker: If we remove this helper and instead pass the target buffers for packed `t0` / `t1` to `mld_compute_t0k_t1k`, then we could add the packing to `mld_compute_t0k_t1k` and share a bit more code.

In that case, it might make sense though to introduce macros for the different offsets in the signing buffer, so you don't have to replicate `sk + 2 * MLDSA_SEEDBYTES + MLDSA_TRBYTES + MLDSA_L * MLDSA_POLYETA_PACKEDBYTES + MLDSA_K * MLDSA_POLYETA_PACKEDBYTES + k * MLDSA_POLYT0_PACKEDBYTES` in the high-level logic. Having such abbreviations in a single place might be useful anyway.

mldsa/src/sign.c
```diff
@@ -1535,45 +1550,68 @@ int mld_sign_pk_from_sk(uint8_t pk[MLDSA_CRYPTO_PUBLICKEYBYTES],
...
-    goto cleanup;
+    mld_compute_t0k_t1k(t0k, t1k, t0k, mat, k, s1, &s2->vec[k]);
+    mld_pack_pk_t1(pk, k, t1k);
+    mld_polyt0_pack(t0k_packed, t0k);
+    cmp0 |= mld_ct_memcmp(t0k_packed,
+                          sk + 2 * MLDSA_SEEDBYTES + MLDSA_TRBYTES +
+                              MLDSA_L * MLDSA_POLYETA_PACKEDBYTES +
+                              MLDSA_K * MLDSA_POLYETA_PACKEDBYTES +
+                              k * MLDSA_POLYT0_PACKEDBYTES,
+                          MLDSA_POLYT0_PACKEDBYTES);
```
hanno-becker: Is it important that we to this check incrementally here? Otherwise, we could hoist the entire loop into `mld_compute_t0k_t1k` (which should then be renamed to `mld_compute_pack_t0_t1` or something) and have a single CT-memcmp on the resulting slice of SK.

hanno-becker (changes requested):
Thank you @mkannwischer. I think there may be more potential for consolidation/simplification of code; see the comments.

mldsa/src/sign.c
```diff
@@ -197,76 +195,96 @@ __contract__(
...
+ *              - mld_poly *t0k: scratch polynomial
+ *              - mld_poly *t1k: scratch polynomial
  **************************************************/
-MLD_MUST_CHECK_RETURN_VALUE
-static int mld_compute_t0_t1_tr_from_sk_components(
-    mld_polyveck *t0, mld_polyveck *t1, uint8_t tr[MLDSA_TRBYTES],
-    uint8_t pk[MLDSA_CRYPTO_PUBLICKEYBYTES], const uint8_t rho[MLDSA_SEEDBYTES],
-    mld_polyvecl *s1, const mld_polyveck *s2,
-    MLD_CONFIG_CONTEXT_PARAMETER_TYPE context)
+static void mld_compute_pack_t0_t1(
```
hanno-becker: You can move the matrix expansion in here, too, or not?

mkannwischer: Yes, can do and will do.

mkannwischer: Done.

mldsa/src/sign.c
```diff
@@ -200,33 +197,55 @@ __contract__(
 }
 
 /*************************************************
- * Name:        mld_compute_t0_t1_tr_from_sk_components
+ * Name:        mld_compute_pack_t0_t1
+ *
+ * Description: Compute t = A*s1hat + s2 row by row, decompose each row
+ *              into t0[k] and t1[k] via power2round, and bit-pack t1[k]
+ *              into pk and t0[k] into the t0_packed buffer. Used by
```
hanno-becker: Nit: `s/pk/pk_t1`

mldsa/src/sign.c
```diff
@@ -317,34 +340,45 @@ int mld_sign_keypair_internal(uint8_t pk[MLDSA_CRYPTO_PUBLICKEYBYTES],
...
+  /* Compute tr = H(pk) */
+  mld_shake256(tr, MLDSA_TRBYTES, pk, MLDSA_CRYPTO_PUBLICKEYBYTES);
+
+  /* Pack remaining secret key components (s1 and t0 already packed) */
+  mld_pack_sk_rho_key_tr_s2(sk, rho, tr, key, s2);
 
   /* Constant time: pk is the public key, inherently public data */
   MLD_CT_TESTING_DECLASSIFY(pk, MLDSA_CRYPTO_PUBLICKEYBYTES);
 
+  ret = 0;
```
hanno-becker: This is dead code? I'm surprised the compilers don't warn about it.

hanno-becker (approved):
Two nits, otherwise LGTM

mldsa/src/sign.c
mkannwischer: Indeed. removed.

mldsa/src/sign.c

PR #1031, opened 2026-04-08
lowram: Eliminate y vector in REDUCE_RAM mode in sign

mkannwischer:
Add mld_yvec following the lazy polyvec pattern: in REDUCE_RAM mode it
stores only the seed and nonce and regenerates y[i] on demand. The
fused matrix-vector helper mld_polyvec_matrix_pointwise_montgomery_yvec
samples y column-by-column in REDUCE_RAM mode.

- Depends on #1019 
- Depends on #1060
- Hoisted out from #1005


mldsa/src/polyvec_lazy.c
```diff
@@ -204,6 +218,50 @@ void mld_polyvec_matrix_pointwise_montgomery_lazy(mld_polyveck *t,
   mld_zeroize(seed_ext, sizeof(seed_ext));
 }
 
+#if !defined(MLD_CONFIG_NO_SIGN_API)
+MLD_INTERNAL_API
+void mld_polyvec_matrix_pointwise_montgomery_yvec_lazy(mld_polyveck *w,
```
hanno-becker, line 247: I am wary of the repetition here, and of the dependency between the laziness of `yvec` vs. `mat`. It would be better if we could arrange things so that the lazy/eager implementations are entirely independent. Have you thought about this?

Are there more lazy wrappers to come which would introduce further duplication?

hanno-becker (commented):
I am wary of the code duplication introduced here and furthered with the introduction of the lazy zvec, but I agree with you @mkannwischer that this is better discussed once everything is in place.

I'm happy to proceed with this PR once #1060 is merged and this one is rebased atop (as I understand it the plan?)

mkannwischer:
> I am wary of the code duplication introduced here and furthered with the introduction of the lazy zvec, but I agree with you @mkannwischer that this is better discussed once everything is in place.
> 
> I'm happy to proceed with this PR once #1060 is merged and this one is rebased atop (as I understand it the plan?)

Rebased now.

hanno-becker (approved):

PR #1032, opened 2026-04-08
FIPS202/x86_64: Don't use native x4 backend on x86_64 when not needed

flynd:
Match the AArch64 behavior and skip the native Keccak-f1600x4 backend when MLD_CONFIG_SERIAL_FIPS202_ONLY or MLD_CONFIG_REDUCE_RAM is set.

This change was part of PR #1000 but doesn't need to be part of that stack.

### DIFF
--- a/dev/fips202/aarch64/auto.h
+++ b/dev/fips202/aarch64/auto.h
@@ -37,6 +37,7 @@
 #include "x1_scalar.h"
 #endif
 
+#if !defined(MLD_CONFIG_SERIAL_FIPS202_ONLY)
 /*
  * Keccak-f1600x2/x4
  *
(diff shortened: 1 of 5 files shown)


mkannwischer:
Thanks @flynd. Can you elaborate how this matches the AArch64 behavior? I don't see a matching guard in the AArch64 backend.

flynd:
> Thanks @flynd. Can you elaborate how this matches the AArch64 behavior? I don't see a matching guard in the AArch64 backend.

This was a commit you made (added to my stack in PR #1000) that I noticed didn't have any dependencies to that stack.  I see now that you're referring to one of my commits so the commit message is now referring to something that doesn't exist on master yet.

I will grab the relevant parts of my commit and add to this PR to make it consistent.

mkannwischer (approved):
Thanks @flynd! LGTM.

PR #1038, opened 2026-04-14
Fix context-length validation in ACVP test binary

hanno-becker:
Three places in acvp_mldsa.c validated the context argument length using `mlen > MAX_MSG_LENGTH` instead of `ctxlen > MAX_CTX_LENGTH`, meaning an oversized context would not be caught before being written into a MAX_CTX_LENGTH-sized stack buffer.


mkannwischer (approved):
LGTM. Thanks!

PR #1042, opened 2026-04-15
poly_ntt_c add --slice-formula to stabilize proof

rod-chapman:
Add --slice-formula in proof of poly_ntt_c() to stabilize proof performance on both macOS and Linux platforms.


rod-chapman:
SMT files generated on macOS and Linux are identical, but Z3 is not stable on this problem. Proof on Linux takes <1 second, but >2 hours on macOS.

Addition of `--slice-formula` fixes the problem locally. A separate defect report on Z3 will be filed.

mkannwischer (approved):

PR #1043, opened 2026-04-16
native: Allow tables to be declared static

flynd:
When building single-CU builds and MLD_CONFIG_INTERNAL_API_QUALIFIER is set to make internal API static, also make the data tables static. This makes the data symbols local instead of global, hiding them from the scope of the rest of the application.


hanno-becker:
Thank you @flynd, I have no issue with this change in principle. Could you think about how we could test this? Perhaps we can run an extra check on the monobuild example confirming that only the public API appears in the symbol table?

flynd:
This was minor thing I thought would be nice, but it seems this wasn't as simple as I thought.

If I understand correctly the failed CI test, I cannot use forward declarations for static data without getting an error, at least with some compiler flags.
I tried to just remove the forward declaration and reorder the files in mldsa_native.c to include the C files with tables first, but the tables are used in meta.h which is pulled in by common.h, which must be included first to get all the defines, so without moving around code between files, this is a circular dependency.

I don't know if it is worth trying to solve this.  @hanno-becker, do you have a good idea on how to resolve this that doesn't involve a lot of work?  If not I'll just abandon this PR as it was just a nice to have, not something important.

mldsa/src/common.h
```diff
@@ -26,8 +26,12 @@
  * this can be overwritten by the user, e.g. for single-CU builds. */
 #if !defined(MLD_CONFIG_INTERNAL_API_QUALIFIER)
 #define MLD_INTERNAL_API
+#define MLD_INTERNAL_DATA_DECLARATION extern
+#define MLD_INTERNAL_DATA_DEFINITION
 #else
 #define MLD_INTERNAL_API MLD_CONFIG_INTERNAL_API_QUALIFIER
+#define MLD_INTERNAL_DATA_DECLARATION MLD_CONFIG_INTERNAL_API_QUALIFIER
+#define MLD_INTERNAL_DATA_DEFINITION MLD_CONFIG_INTERNAL_API_QUALIFIER
```
hanno-becker, line 34: I am slightly worried that somebody might want to use an internal API qualifier for functions that makes no sense for data declarations. E.g. see https://github.com/CHERIoT-Platform/cheriot-pqc/blob/main/include/mldsa_native_config.h#L20 using `__cheriot_libcall` as the _external_ API qualifier.

On the other hand, if such a customer comes along, we can still split things up in two separate config options. So I am leaning towards going with the proposed approach.

@mkannwischer WDYT?

flynd:
> > I don't know if it is worth trying to solve this. @hanno-becker, do you have a good idea on how to resolve this that doesn't involve a lot of work? If not I'll just abandon this PR as it was just a nice to have, not something important.
> 
> This is not a blocker for this PR, but a nice-to-have. We can proceed with the review and potentially merge regardless.

To be clear, I wasn't primarily referring to being able to test this feature but to the fact that the CI test fails and I don't know how to fix it.

Making the tables static can be implemented two ways:  With forward declarations or without.
The PR currently has forward declarations, but at least with some compiler or compiler flags, the compilation fails when forward declaring static data so I don't think forward declarations can be used.
I have tried to rework it to remove the forward declarations, however that requires the tables to appear before they are used, but due to how the headers are included it would require a bigger move or splitting of code which I didn't feel comfortable doing.
Possibly it could be resolved by adding pointers to the tables that can be forward declared and then the compiler optimizes away the pointers so they never appear in the built binary.

So currently I have no version of this change that can pass the CI tests and I'm not sure this is important enough to try and solve.

hanno-becker:
@flynd I think adding array sizes solves the issue.

flynd:
> @flynd I think adding array sizes solves the issue.

That's great.  I have updated my original commit to add the missing signed-off footer so hopefully CI can pass now.

hanno-becker (approved):

PR #1046, opened 2026-04-18
aarch64: Use more pairwise load/store

flynd:
Replace pairs of ldr/str instructions with ldp/stp when possible to reduce the size of the code.

I only replaced the trivial cases here.  It should also be possible to also shorten ntt.S and intt.S.


flynd:
I thought this would be a simple change to make, but some of the proofs fail and I have no idea why changing the load/store instructions would result in problems with PC (the error given is "`Failure("ARM_CONV: can't find 'read PC .. = ..' from ths")`").

hanno-becker (changes requested):
Performance-wise, I believe it is discouraged to update the address register with post-increment repeatedly, as this creates unnecessary data dependencies between the load instructions. It also prevents SLOTHY superoptimization (for the same reason).

@flynd If code-size is so critical in your application, perhaps this may be a change worth taking locally?

flynd:
> Performance-wise, I believe it is discouraged to update the address register with post-increment repeatedly, as this creates unnecessary data dependencies between the load instructions. It also prevents SLOTHY superoptimization (for the same reason).

I was not aware of this.  I simply preferred post-incrementing as I think it makes the code easier to read.  I can change the commit to avoid post-incrementing.

I notice that in some places the single post-increment is in the first load with the following loads using negative offsets, while in other places the single post-increment is in the last load with the other loads using positive offsets.  Is there a preference for which of these styles should be used?

> @flynd If code-size is so critical in your application, perhaps this may be a change worth taking locally?

Important but not critical enough that this is strictly necessary.  I only thought it was an obvious optimization to do when I saw the repeated ldr/str sequences.  If using ldp/stp has performance costs compared to using twice the number of ldr/str instructions or if there is another reason to prefer not using ldp/stp then we can abandon this change.

I considered using ld1/st1 to load/store 4 register per instructions, but I couldn't be sure doing 64bit data accesses instead of 128bit wouldn't affect performance enough to be noticeable so I thought it safer to just use ldp/stp.

hanno-becker:
> I was not aware of this. I simply preferred post-incrementing as I think it makes the code easier to read. I can change the commit to avoid post-incrementing.

For completeness, I was referring to the sections "4.3 Optimizing memory routines" in Arm's Software Optimization Guides (e.g. https://developer.arm.com/documentation/110659/7-0/) which call out non-writeback memory operations as more performant. Admittedly, when looking at the preamble of a complex function, it may not make any meaningful difference.

If we can avoid repeated post-increment, though, the change may still be a win and compatible with SLOTHY.

hanno-becker:
I remembered that another reason why we didn't use pairwise load/stores is that it removes re-scheduling flexibility in SLOTHY: While not as problematic as the data dependencies from repeated post-increments, forcing the load/store of two values happen together limits the amount of reordering/re-allocation that SLOTHY can do. I don't know how significant this is in our examples, though.

flynd:
> I remembered that another reason why we didn't use pairwise load/stores is that it removes re-scheduling flexibility in SLOTHY: While not as problematic as the data dependencies from repeated post-increments, forcing the load/store of two values happen together limits the amount of reordering/re-allocation that SLOTHY can do. I don't know how significant this is in our examples, though.

I see.  I only changed load/stores that are made in bulk in the generated files.  The places where load/stores are interleaved with processing (as in intt.S) I haven't modified.

The current version of the change preserves the use of offset with post-increment or post-increment with negative offset as the previous code had, i.e. I'm not doing any more post-increments than before.

This isn't an important change to me.  I just thought it was an obvious optimization that I could contribute.  If you think it is a bad idea, it is okay to reject it.  Otherwise I need help with making the proofs pass as I don't understand the error messages and HOL-Light seems complicated.

Also, the lint failure is for a broken link in the README which I didn't change.

flynd:
So the benchmark say this is worse.  This surprises me as I would expect using ldp instead of ldr+ldr would at worst be the same and at best be slightly better due to needing to fetch fewer instructions.  I guess there is some deeper magic in the CPU that optimizes single load/store instructions better.

I noticed that many places has post-increment with the first load/store, making the later load/stores depend on the increment of the first one.  I'd expect it to be better if always doing the post-increment with the last load/store instruction instead.  If that matters it could be a relevant change regardless if using single or pair accesses.

flynd:
I pushed an update that always puts the post-increment last.  I assumed the benchmarks would be rerun but I'm not seeing any new result, so how do I trigger them to run again?

flynd:
I thought this was a simple improvement.  It wasn't.  I don't think it's worth spending any more time on this, including resolving the merge conflict with the whitespace change that just happened on main.

PR #1049, opened 2026-04-20
Remove --slice-formula from all Makefiles and set in Makefile.common

rod-chapman:
Remove --slice-formula from all Makefiles and set in Makefile.common

Also, use SMT-only Z3 tactic for proof of poly_invntt_tomont_c() which is faster with --slice-formula

### DIFF
--- a/proofs/cbmc/Makefile.common
+++ b/proofs/cbmc/Makefile.common
@@ -543,9 +543,27 @@ SPACE :=$() $()
 COMMA :=,
 
 ################################################################
-# Set C compiler defines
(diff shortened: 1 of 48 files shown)


rod-chapman:
Proof times on Grv4/Ubuntu, CPU user time in minutes:seconds for param sets 44,65,87:
```
main branch: 32:38, 36:54, 39:25
this branch: 32:06, 36:24, 38:56
```

mkannwischer (approved):
Thanks @rod-chapman. LGTM. Performance improved slightly.

PR #1050, opened 2026-04-21
CBMC: Run proofs with MLD_CONFIG_REDUCE_RAM in CI

mkannwischer:
Extend the CBMC test harness and CI to exercise proofs under MLD_CONFIG_REDUCE_RAM in addition to the default configuration.

- scripts/tests: new --reduce-ram flag on the cbmc subcommand that exports MLD_CONFIG_REDUCE_RAM=1 to run-cbmc-proofs.py.
- proofs/cbmc/Makefile.common: propagate MLD_CONFIG_REDUCE_RAM from the environment into -DMLD_CONFIG_REDUCE_RAM.
- .github/workflows/cbmc.yml: collapse the per-parameter-set jobs into a single matrix over parameter_set x reduce_ram, yielding six runs.
- .github/workflows/ci_ec2_reusable.yml, .github/actions/cbmc/action.yml: thread reduce_ram through to the action; use dev/cbmc-reduce-ram as a separate baseline directory so runtime-regression data does not collide with the default-config baseline.
- .github/actions/cbmc/report.py: new --variant option that suffixes the PR comment tag/title so default and reduce_ram variants don't overwrite each other's comments.
- proofs/cbmc/reduce_ram_skip.txt: skiplist of proofs that currently to not work - either because the taret function does not exist in REDUCE_RAM builds or because the proof needs to be adapted.

### DIFF
--- a/.github/actions/cbmc/action.yml
+++ b/.github/actions/cbmc/action.yml
@@ -21,6 +21,9 @@ inputs:
   mldsa_parameter_set:
     description: "Security level for ML-DSA (2,3,5)"
     default: "2"
+  reduce_ram:
+    description: "Run CBMC proofs with MLD_CONFIG_REDUCE_RAM enabled"
+    default: "false"
   gh_token:
     description: Github access token to use
     required: true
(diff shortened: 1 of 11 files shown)


hanno-becker (changes requested):
Thank you very much @mkannwischer, I think this is a good approach!

One small thing: Can we add the skipped proofs to the CBMC CI summary? They could be special table entries and perhaps warned about at the top, like performance regressions.

mkannwischer:
> One small thing: Can we add the skipped proofs to the CBMC CI summary? They could be special table entries and perhaps warned about at the top, like performance regressions.

I like the idea and just pushed a commit adding it.

hanno-becker (approved):
LGTM, thank you @mkannwischer !

PR #955, opened 2026-02-06
HOL-Light: Add x86 AVX2 nttunpack proof

jakemas:
- Resolves https://github.com/pq-code-package/mldsa-native/issues/913

PR #1051, opened 2026-04-22
check-contracts: Detect contracts separated from prototype by a comment

mkannwischer:
The previous regex skipped contracts with a block comment between the prototype's `)` and `__contract__`, silently missing six AVX2 contracts (ntt_avx2, invntt_avx2, pointwise_avx2, pointwise_acc_l{4,5,7}_avx2). Strip comments before matching, and add `_avx2` to the exception list alongside `_native` and `_asm`.

Discovered in #955.

### DIFF
--- a/scripts/check-contracts
+++ b/scripts/check-contracts
@@ -33,6 +33,16 @@ def gen_proofs():
     return proofs
 
 
+def strip_comments(content):
+    """Replace block and line comments with whitespace, preserving line numbers."""
+
+    def repl(m):
+        return re.sub(r"[^\n]", " ", m.group(0))
+
+    # Block comments and line comments.
+    return re.sub(r"/\*[\s\S]*?\*/|//[^\n]*", repl, content)
+
+
 def gen_contracts():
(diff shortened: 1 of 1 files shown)


scripts/check-contracts
```diff
@@ -33,14 +33,26 @@ def gen_proofs():
...
 
 
+def strip_comments(content):
+    """Replace block and line comments with whitespace, preserving line numbers."""
+
+    def repl(m):
+        return re.sub(r"[^\n]", " ", m.group(0))
+
+    # Block comments and line comments.
+    return re.sub(r"/\*[\s\S]*?\*/|//[^\n]*", repl, content)
```
hanno-becker, line 43: Does this work for multiline block comments?

mkannwischer: yes. that's what we have.

hanno-becker (approved):
Thanks for spotting this @mkannwischer !

jakemas:
Thank you!

PR #1052, opened 2026-04-22
CodeQL: Add mldsa/ declaration visibility check

hanno-becker:
CodeQL is GitHub's semantic code analysis engine: it builds a typed AST database from a build trace, then answers queries written in QL (a Datalog-flavoured language) against it.

This commit adds a custom query that flags any externally-linked declaration under mldsa/ lacking the appropriate visibility annotation (MLD_INTERNAL_API, MLD_EXTERNAL_API, MLD_API_QUALIFIER for functions; MLD_INTERNAL_DATA_DECLARATION / MLD_INTERNAL_DATA_DEFINITION for variables), catching new cross-TU symbols that miss the macro required for single-CU builds. Assembly symbols (*_asm) and empty-CU placeholders (*empty_cu_*) are excluded.

A scripts/ql driver wraps the CLI: it manages the CodeQL database (auto-rebuilding when sources change), dispatches queries, and runs query-specific post-processors that add colored output and pass/fail logic. The mldsa-declarations post-processor is registered for the visibility check; other queries get plain table output.

A codeql.yml workflow invoked from all.yml runs the check in CI.

The new query flagged two missing annotations:
mld_polyeta_unpack in poly_kl.c and mld_prepare_domain_separation_prefix in sign.c lacked MLD_INTERNAL_API / MLD_EXTERNAL_API, respectively. Both are fixed in this commit, too.

Licensing: CodeQL uses a GitHub-specific license and is marked 'unfree' in nixpkgs. It is free for open-source analysis and GitHub code scanning, covering our use case. The nix flake does not globally allowlist the package; developers opt in via
    NIXPKGS_ALLOW_UNFREE=1 nix develop --impure .#codeql
to avoid silent acceptance of the non-OSS license.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,25 @@
+# Copyright (c) The mldsa-native project authors
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+
+name: CodeQL
+permissions:
+  contents: read
+on:
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+  mldsa-declarations:
+    name: mldsa-declarations
+    runs-on: ubuntu-latest
+    env:
+      NIXPKGS_ALLOW_UNFREE: 1
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: ./.github/actions/setup-shell
+        with:
+          nix-shell: codeql
+          impure: 'true'
+          gh_token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            scripts/ql run mldsa-declarations
--- a/flake.nix
+++ b/flake.nix
@@ -118,6 +118,9 @@
           devShells.cbmc = util.mkShell {
             packages = builtins.attrValues { inherit (config.packages) cbmc toolchains_native; } ++ [ pkgs.gh ];
           };
+          devShells.codeql = util.mkShell {
+            packages = [ pkgs.codeql ];
+          };
           devShells.slothy = util.mkShell {
             packages = builtins.attrValues { inherit (config.packages) slothy linters toolchains_native; };
           };
--- a/scripts/ql
+++ b/scripts/ql
@@ -1,0 +1,640 @@
+#!/usr/bin/env python3
+# Copyright (c) The mldsa-native project authors
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+
+#
+# Convenience driver for running the custom CodeQL queries in test/codeql/
+# against the mldsa-native tree.
+#
+# Usage:
+#   scripts/ql list                 List available queries
+#   scripts/ql run <name> [flags]   Run a query, building the DB if needed
+#   scripts/ql build-db [flags]     (Re)build the CodeQL database
+#
+# Database management:
+#   The CodeQL database is cached at .codeql-db/ and reused across runs.
+#   Before reuse, the driver runs `CC=cc make -n` to detect whether any
+#   source has changed since the last build; if so, the database is
+#   automatically rebuilt.  Pass --rebuild to force a rebuild regardless.
+#   A `make clean` is always run before creating a new database so that
+#   CodeQL's build tracer sees every compilation.
+#
+# Query dispatch:
+#   Queries with @kind problem / path-problem are run via
+#   `codeql database analyze` (SARIF or CSV alerts).  All other queries
+#   (@kind table or no @kind) are run via `codeql query run`, which
+#   produces a binary result file (.bqrs) that is then decoded to the
+#   requested output format (CSV, JSON, or text).
+#
+# Per-query post-processing:
+#   Tabular queries can register a post-processor function in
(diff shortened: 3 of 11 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,25 @@
...
+  mldsa-declarations:
+    name: mldsa-declarations
+    runs-on: ubuntu-latest
+    env:
+      NIXPKGS_ALLOW_UNFREE: 1
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: ./.github/actions/setup-shell
+        with:
+          nix-shell: codeql
```
mkannwischer, line 21: what's the benefit of installing codeql via nix in CI? Wouldn't `github/codeql-action/init` be the easier and less intrusive path?

mkannwischer: Or do we expect this to be very version sensitive? That's the only reason I can think of.

scripts/ql
mkannwischer, line 1: I'm not sure I see the value of this 640-line Python script.
What's the benefit over just changing the rules to actually throw errors and the workflow to 
```
  - uses: github/codeql-action/init@x                                                                                                                                                                     
    with:                                                                                                            
      languages: cpp
      queries: ./test/codeql/mldsa-declarations.ql
```

One could still keep the `ql` script for local testing, but that sounds more like 10 line bash script rather than a 640 line Python script.

.github/workflows/codeql.yml
hanno-becker: I felt it's in keeping with our existing setup, and yes, more robust towards version changes. But I am OK either way.

flake.nix
```diff
@@ -118,6 +118,9 @@
           devShells.cbmc = util.mkShell {
             packages = builtins.attrValues { inherit (config.packages) cbmc toolchains_native; } ++ [ pkgs.gh ];
           };
+          devShells.codeql = util.mkShell {
```
mkannwischer, line 121: The following diff would allow us to install it unconditionally: 
```
diff --git a/flake.nix b/flake.nix
index 3231ee183..83c7721d1 100644
--- a/flake.nix
+++ b/flake.nix
@@ -50,6 +50,8 @@
         {
           _module.args.pkgs = import inputs.nixpkgs {
             inherit system;
+            config.allowUnfreePredicate = pkg:
+              builtins.elem (inputs.nixpkgs.lib.getName pkg) [ "codeql" ];
             overlays = [
               (_:_: {
                 clang_22 = pkgs-unstable.clang_22;
```

With an appropriate comment explaining the license, I believe that may be preferable?

test/codeql/qlpack.yml
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) The mldsa-native project authors
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+name: mldsa-native/queries
+version: 0.0.1
+library: false
+dependencies:
+  # Pinned to the version shipped with the CodeQL CLI currently packaged in
+  # nixpkgs (2.23.3). Bump together with the CLI.
```
mkannwischer, line 8: What happens if we forget to bump this? I would prefer leaving out this comment as this is guaranteed to be outdated soon.

flake.nix
mkannwischer: If we expect the developer to run this, this should be included in the default shell.

mkannwischer (commented):
Thanks @hanno-becker. Thanks for looking into this and finding the two places where we were missing the markers.

I have some questions. If we plan to scale this, I think it makes sense to try to simplify this as much as possible.

flake.nix
hanno-becker: I would like the non-free license to be explicit, so the lack of convenience in the current PR is by design.

scripts/ql
hanno-becker: Can you elaborate? What are you suggesting?

mkannwischer: I'll give it a try.

test/codeql/qlpack.yml
hanno-becker: Ok!

flake.nix
mkannwischer: If we consider this non-free, then I prefer to not use it at all.
There seem to be free alternatives. Let me try that.

hanno-becker: > If we consider this non-free, then I prefer to not use it at all.

No need to be undifferentiated here. The license terms cover our use case. But I don't want to expose a nix shell which hides from the user the fact that not all use cases are covered. For that reason I also don't want it in the default shell, which I want to continue to work as-is.

scripts/ql
hanno-becker: Thanks! If we can directly trigger an error from the .ql that may be easier indeed.

hanno-becker:
Shelving for now, @mkannwischer is not convinced of the approach and will explore alternatives first. We can always come back to this.

mkannwischer:
Reopening so we don't forget this exists.

mkannwischer:
@hanno-becker, could you rebase this please?

PR #977, opened 2026-02-27
HOL-Light: Add AArch64 poly_decompose_{32,88} correctness proofs

mkannwischer:
- Based on https://github.com/pq-code-package/mldsa-native/pull/971
- Resolves https://github.com/pq-code-package/mldsa-native/issues/922

PR #1058, opened 2026-04-24
native: Unify asm backend symbol naming

mkannwischer:
- Port of https://github.com/pq-code-package/mlkem-native/pull/1663
- Depends on #955
- Depends on #977


hanno-becker (changes requested):
We should also cover the new AVX2 Keccak ASM.

hanno-becker (approved):

PR #1059, opened 2026-04-24
sign: Add MLD_CONFIG_MAX_SIGNING_ATTEMPTS and dedicated error code

### LINKED ISSUE #1055: RFC: Consider configurable bound for the number of signing attempts
**Issue:** In timing-sensitive environments, the theoretically unbounded nature of the ML-DSA signing routine can pose problems.

**Proposal/Task:** Add a configuration option `MLD_CONFIG_MAX_SIGNING_ATTEMPTS` bounding the number of signing attempts. A new error code would likely be needed that can be returned when the maximum number of attempts is exhausted.

hanno-becker:
* Resolves https://github.com/pq-code-package/mldsa-native/issues/1055

Introduce a new configuration option, MLD_CONFIG_MAX_SIGNING_ATTEMPTS, that bounds the number of rejection-sampling iterations performed by the ML-DSA signing routine (FIPS 204, Algorithm 7). This is useful in timing-sensitive environments that require a deterministic worst-case bound on signing time.

The default is the maximum value before non-overflow -- practically infinite. A compile-time check rejects values below the FIPS 204 Appendix C minimum of 814 (at which exhaustion has probability < 2^-256).

When the bound is reached, signing now returns a dedicated error code MLD_ERR_SIGN_ATTEMPTS_EXHAUSTED instead of the generic MLD_ERR_FAIL. CBMC specifications and documentation for all signing entry points are adjusted accordingly. The PCT path (mld_check_pct) and keypair contracts are also updated because the new code can propagate through MLD_CONFIG_KEYGEN_PCT.

- [x] Add test for new configuration option
- [x] Consider a test-only configuration with a lower-than-FIPS204-allowed number of signature attempts, so tests can exercise the error code `MLD_ERR_SIGN_ATTEMPTS_EXHAUSTED`.


test/configs/configs.yml
```diff
@@ -165,6 +165,14 @@ configs:
           }
           #endif
 
+  - path: test/configs/low_signing_bound_config.h
+    description: "Test configuration with signing bound set to 1 to exercise MLD_ERR_SIGN_ATTEMPTS_EXHAUSTED"
+    defines:
+      MLD_CONFIG_MAX_SIGNING_ATTEMPTS:
+        content: |
+          #define MLD_ALLOW_NONCOMPLIANT_SIGNING_BOUND
```
mkannwischer, line 173: Why is this set twice? Once in the config and once in the action itself?

test/src/test_mldsa.c
```diff
@@ -21,6 +22,14 @@
...
 #define crypto_sign_pk_from_sk MLD_API_NAMESPACE(pk_from_sk)
 
+#if defined(MLD_ALLOW_NONCOMPLIANT_SIGNING_BOUND)
+#define MLD_SIGN_EXHAUSTION_TOLERATED 1
+static unsigned mld_sign_attempts = 0;
+static unsigned mld_sign_exhaustion_count = 0;
+#else
+#define MLD_SIGN_EXHAUSTION_TOLERATED 0
+#endif
+
```
mkannwischer: I'm not a big fan of complicating our test with a configuration that is not recommended.
For what you are currently testing in CI, it would be sufficent to output an error string in case an error occurs. That would be indepentent of this feature.
Could we switch to that?

mldsa/src/sign.h
```diff
@@ -311,7 +322,7 @@ __contract__(
   assigns(memory_slice(sig, MLDSA_CRYPTO_BYTES))
   assigns(object_whole(siglen))
   ensures((return_value == 0 && *siglen == MLDSA_CRYPTO_BYTES) ||
-          ((return_value == MLD_ERR_FAIL || return_value == MLD_ERR_OUT_OF_MEMORY || return_value == MLD_ERR_RNG_FAIL) && *siglen == 0))
+          ((return_value == MLD_ERR_FAIL || return_value == MLD_ERR_OUT_OF_MEMORY || return_value == MLD_ERR_RNG_FAIL || return_value == MLD_ERR_SIGN_ATTEMPTS_EXHAUSTED) && *siglen == 0))
```
mkannwischer, line 325: This becoming harder to read. Any ideas how to make it more readable?

mldsa/src/sign.c
```diff
@@ -555,6 +556,38 @@ __contract__(
...
+
+#if MLD_CONFIG_MAX_SIGNING_ATTEMPTS < 1
+#error Bad configuration: MLD_CONFIG_MAX_SIGNING_ATTEMPTS must be >= 1
+#endif
+
+#if MLD_CONFIG_MAX_SIGNING_ATTEMPTS > MLD_NONCE_UB
+#error Bad configuration: MLD_CONFIG_MAX_SIGNING_ATTEMPTS exceeds MLD_NONCE_UB (the 16-bit nonce type-safety upper bound)
+#endif
+
+static MLD_INLINE uint16_t mld_get_max_signing_attempts(void)
```
mkannwischer, line 580: add MLD_MUST_CHECK_RETURN_VALUE

mldsa/src/sign.c
```diff
@@ -555,6 +556,38 @@ __contract__(
...
+    MLD_CONFIG_MAX_SIGNING_ATTEMPTS < 814
+#error Bad configuration: MLD_CONFIG_MAX_SIGNING_ATTEMPTS must be >= 814 for FIPS 204 compliance @[FIPS204, Appendix C]
+#endif
+
+#if MLD_CONFIG_MAX_SIGNING_ATTEMPTS < 1
+#error Bad configuration: MLD_CONFIG_MAX_SIGNING_ATTEMPTS must be >= 1
+#endif
+
+#if MLD_CONFIG_MAX_SIGNING_ATTEMPTS > MLD_NONCE_UB
+#error Bad configuration: MLD_CONFIG_MAX_SIGNING_ATTEMPTS exceeds MLD_NONCE_UB (the 16-bit nonce type-safety upper bound)
```
mkannwischer: I don't think `MLD_NONCE_UB (the 16-bit nonce type-safety upper bound)` can be understood by anyone who's not fairly familar with the code base. Please rephrase.

mldsa/src/sign.c
```diff
@@ -555,6 +556,38 @@ __contract__(
   return 0;
 }
 
+/* User-facing bound on signing attempts. See the doc block for
+ * MLD_CONFIG_MAX_SIGNING_ATTEMPTS in mldsa_native_config.h. Default is
+ * MLD_NONCE_UB, which practically infinite. */
```
mkannwischer: there is a word missing in `which practically infinite.`
maybe we can also be more precise:
`The default value MLD_NONCE_UB is exceeded with probability less than 2^-256, i.e., partically never.`

mkannwischer (changes requested):
Thanks @hanno-becker - looks mostly good.
I'm not very happy that it complicates the test for no good reason (the CI tests don't actually check the new values being captured). I'd prefer if we can keep the test simple and independent of this configuration option.

mldsa/mldsa_native.h
```diff
@@ -226,6 +231,9 @@ extern "C"
  *     - MLD_ERR_OUT_OF_MEMORY: If MLD_CONFIG_CUSTOM_ALLOC_FREE is
  *         used and an allocation via MLD_CUSTOM_ALLOC returned NULL.
  *     - MLD_ERR_RNG_FAIL: Random number generation failed.
+ *     - MLD_ERR_SIGN_ATTEMPTS_EXHAUSTED: The PCT's signing step exhausted
+ *         MLD_CONFIG_MAX_SIGNING_ATTEMPTS iterations. Only possible when
+ *         MLD_CONFIG_KEYGEN_PCT is enabled.
```
mkannwischer, line 236: I think it is confusing that keygen can return this error. Should we just map it to `MLD_ERR_FAIL`?

hanno-becker: Hm no, I think the way it is is the cleanest solution -- why would we obfuscate the source of the error?

mldsa/src/sign.c
hanno-becker: Good idea

mldsa/src/sign.c
hanno-becker: Good point

test/src/test_mldsa.c
hanno-becker: I think there's some value to seeing those rejection rates and confirming that they align with what they should be. But if you insist I can remove it.

mldsa/src/sign.h
hanno-becker: One could consider adding a helper macros like `any_error(return_value)`; but let's discuss that separately.

test/src/test_mldsa.c
hanno-becker: Removed

mldsa/src/sign.h
hanno-becker: https://github.com/pq-code-package/mldsa-native/issues/1096

mldsa/src/sign.c
hanno-becker: Done

mldsa/src/sign.c
hanno-becker: Done

mldsa/src/sign.c
hanno-becker: Done

mldsa/mldsa_native.h
hanno-becker: Kept

test/configs/configs.yml
hanno-becker: We set the numeric `MLD_CONFIG_MAX_SIGNING_ATTEMPTS` option, but to allow it being lower than the FIPS204 threshold, we also need to set `MLD_ALLOW_NONCOMPLIANT_SIGNING_BOUND`. Since that's not a documented option (because we don't want to encourage using it), it's snuck in here.

mkannwischer: My question was why do you need to set `MLD_ALLOW_NONCOMPLIANT_SIGNING_BOUND` twice.
It's set once here and once in the CI action.

mldsa/mldsa_native.h
mkannwischer: Okay.

test/configs/configs.yml
hanno-becker: Sorry, I misunderstood. I removed it from the action.

mkannwischer (approved):
Thanks @hanno-becker LGTM now. 

As discussed offline, we should test this feature more thoroughly in a separate test to be added as a follow up. Could you open an issue for that please?

hanno-becker:
@mkannwischer https://github.com/pq-code-package/mldsa-native/issues/1097

PR #1060, opened 2026-04-25
CBMC: Cover `MLD_CONFIG_REDUCE_RAM` mode 

mkannwischer:


scripts/check-contracts
```diff
@@ -99,6 +99,12 @@ def check_contracts():
         if funcname in proofs:
             continue
 
+        if any(
+            funcname.endswith(suffix) and funcname[: -len(suffix)] in proofs
+            for suffix in ("_eager", "_lazy")
+        ):
+            continue
+
```
hanno-becker: Can we do better than just ignore those? E.g. can we drop the suffix to derive the directory name, and then check that the Makefile has a corresponding `CHECK_FUNCTION_CONTRACT` line in it? Still an approximation, but better than before. Checking for `CHECK_FUNCTION_CONTRACTS` would be a general improvement to `check-contracts` not specific to the lazy/eager functions.

mkannwischer: Yes, that would be better. Lets do that in a follow-up?

mldsa/src/packing.h
```diff
@@ -228,6 +228,13 @@ __contract__(
     ensures(forall(k2, 0, MLDSA_K,
       array_abs_bound(s2->vec.vec[k2].coeffs, 0, MLDSA_N, MLD_NTT_BOUND)))
   )
+  MLD_IF_REDUCE_RAM(
+    ensures(s1->packed == old(sk) + 2 * MLDSA_SEEDBYTES + MLDSA_TRBYTES)
+    ensures(s2->packed == old(sk) + 2 * MLDSA_SEEDBYTES + MLDSA_TRBYTES +
+                          MLDSA_L * MLDSA_POLYETA_PACKEDBYTES)
+    ensures(t0->packed == old(sk) + 2 * MLDSA_SEEDBYTES + MLDSA_TRBYTES +
+                          (MLDSA_L + MLDSA_K) * MLDSA_POLYETA_PACKEDBYTES)
```
hanno-becker, line 236: It's a bit odd to be talking about `old(sk)` here seeing this is a parameter passed by value -- the caller has no notion of or interest in whether `sk` is internally modified.

Anyway, not much one can do about it except working with a copy internally, which would clean up the contract.

proofs/cbmc/compute_pack_z/Makefile
```diff
@@ -20,13 +20,17 @@ PROOF_SOURCES += $(PROOFDIR)/$(HARNESS_FILE).c
...
-USE_FUNCTION_CONTRACTS+=mld_poly_reduce 
-USE_FUNCTION_CONTRACTS+=mld_poly_chknorm 
+USE_FUNCTION_CONTRACTS=mld_poly_pointwise_montgomery
+USE_FUNCTION_CONTRACTS+=mld_poly_invntt_tomont
+USE_FUNCTION_CONTRACTS+=mld_poly_add
+USE_FUNCTION_CONTRACTS+=mld_poly_reduce
+USE_FUNCTION_CONTRACTS+=mld_poly_chknorm
 USE_FUNCTION_CONTRACTS+=mld_pack_sig_z
 USE_FUNCTION_CONTRACTS+=mld_zeroize
+ifdef MLD_CONFIG_REDUCE_RAM
```
hanno-becker, line 30: Why are we not calling `mld_sk_s1hat_get_poly` by contract here?

proofs/cbmc/attempt_signature_generation/Makefile
```diff
@@ -22,7 +22,13 @@ PROJECT_SOURCES += $(SRCDIR)/mldsa/src/sign.c
 CHECK_FUNCTION_CONTRACTS=mld_attempt_signature_generation
 USE_FUNCTION_CONTRACTS=mld_polyvecl_uniform_gamma1
 USE_FUNCTION_CONTRACTS+=mld_polyvecl_ntt
+ifdef MLD_CONFIG_REDUCE_RAM
+USE_FUNCTION_CONTRACTS+=mld_polyvec_matrix_pointwise_montgomery_lazy
+USE_FUNCTION_CONTRACTS+=mld_polyeta_unpack
+USE_FUNCTION_CONTRACTS+=mld_polyt0_unpack
```
hanno-becker: Similar here, why aren't we calling all lazy/eager functions by contract?

proofs/cbmc/compute_pack_z/Makefile
mkannwischer: No particular reason. It wasn't needed to get the proof through.
If there are performance issues later on, we can still call them by contracts.

proofs/cbmc/attempt_signature_generation/Makefile
mkannwischer: Why should we?

hanno-becker (changes requested):
Thank you @mkannwischer, overall this looks great!

The main question I have is whether we can arrange things so that every function which uses generic API only (like `attempt_signature_generation`) always calls `foo_{lazy/eager}` by contract, effectively making the `CHECK_FUNCTION_CONTRACTS` list independent of the lazy/eager distinction. I think this would be easier to maintain.

I understand that the invariants and specs of the lazy/eager structures and functions differ and inevitably leak into the specs/invariants of the higher-level functions, but I don't see yet why inlining would be necessary.

proofs/cbmc/attempt_signature_generation/Makefile
hanno-becker: Because we should try to abstract away the lazy/eager distinction as much as we can.

mkannwischer:
> Thank you @mkannwischer, overall this looks great!
> 
> The main question I have is whether we can arrange things so that every function which uses generic API only (like `attempt_signature_generation`) always calls `foo_{lazy/eager}` by contract, effectively making the `CHECK_FUNCTION_CONTRACTS` list independent of the lazy/eager distinction. I think this would be easier to maintain.

It wouldn't be independent, you would need one contract each  for lazy/eager and the contracts wouldn't be the same. 

> 
> I understand that the invariants and specs of the lazy/eager structures and functions differ and inevitably leak into the specs/invariants of the higher-level functions, but I don't see yet why inlining would be necessary.

Inlining is not necessary, but it's less code and it seems to work fine. 

I would prefer if we can defer that until we actually need to change it. Or alternatively, if inlining is bad in general, we need to rework various proofs in mlkem-native/mldsa-native as well.

proofs/cbmc/attempt_signature_generation/Makefile
mkannwischer: I don't think that works. You would still need to different contracts and you need to check against both.

proofs/cbmc/compute_pack_z/Makefile
hanno-becker: I was more thinking about maintenance. It's easier to maintain higher level proofs if they don't reflect implementation details of the lower-level functions they call -- same reason why we ordinarily call functions by contract instead of inlining them.

hanno-becker:
Please can you (or Claude) have a stab at removing the inlining? It is inconsistent with our existing practice elsewhere in the CBMC proofs. I understand that _some_ details inevitably leak -- namely the contracts of the lazy/eager functions, which differ -- but we don't need to leak the implementation details like which functions are called. It is difficult enough to oversee the split between default vs. reduced-RAM mode. Any step to make it simpler we should take.

scripts/check-contracts
hanno-becker: Okay. Can you open an issue?

proofs/cbmc/attempt_signature_generation/Makefile
hanno-becker: Yes, I understand that and say as as much below. But you don't need to leak which functions are being called and the contract of the lazy/eager function provides the template for what exactly you need to hoist into the higher-level function.

proofs/cbmc/attempt_signature_generation/Makefile
```diff
@@ -22,7 +22,15 @@ PROJECT_SOURCES += $(SRCDIR)/mldsa/src/sign.c
...
 USE_FUNCTION_CONTRACTS+=mld_polyvecl_ntt
+ifdef MLD_CONFIG_REDUCE_RAM
+USE_FUNCTION_CONTRACTS+=mld_polyvec_matrix_pointwise_montgomery_lazy
+USE_FUNCTION_CONTRACTS+=mld_sk_s2hat_get_poly_lazy
+USE_FUNCTION_CONTRACTS+=mld_sk_t0hat_get_poly_lazy
+else
 USE_FUNCTION_CONTRACTS+=mld_polyvec_matrix_pointwise_montgomery_eager
+USE_FUNCTION_CONTRACTS+=mld_sk_s2hat_get_poly_eager
+USE_FUNCTION_CONTRACTS+=mld_sk_t0hat_get_poly_eager
+endif
```
hanno-becker, line 33: Another minor improvement we could make here -- which would be in line with the naming of the CBMC proof directories for the lazy/eager functions with dropped suffix -- would be to unconditionally use 
```
USE_FUNCTION_CONTRACTS+=mld_polyvec_matrix_pointwise_montgomery
USE_FUNCTION_CONTRACTS+=mld_sk_s2hat_get_poly
USE_FUNCTION_CONTRACTS+=mld_sk_t0hat_get_poly
```
here and add a stub to Makefile.common which adds the right suffix based on whether eager/lazy is used.

hanno-becker (approved):
LGTM, thanks a lot @mkannwischer. One more suggestion for an improvement to the Makefiles, but it's not blocking.

proofs/cbmc/attempt_signature_generation/Makefile
mkannwischer: I think this also serves as documentation which contract is used in which configuration.
Keeping this verbose could help when debugging proofs in the future, don't you think?

hanno-becker: I addressed this in #1065 instead

PR #1063, opened 2026-04-27
Test: Add Wycheproof tests for ML-DSA

mkannwischer:
Ports the Wycheproof test driver from mlkem-native (pq-code-package/mlkem-native#1588). InvalidPrivateKey vectors are reported as `skipped` since mldsa-native does not validate secret-key coefficient bounds.

### DIFF
--- a/test/wycheproof/wycheproof_client.py
+++ b/test/wycheproof/wycheproof_client.py
@@ -1,0 +1,350 @@
+#!/usr/bin/env python3
+# Copyright (c) The mldsa-native project authors
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+
+# Wycheproof test client for ML-DSA
+# See https://github.com/C2SP/wycheproof
+# Invokes `wycheproof_mldsa{lvl}` under the hood.
+
+import argparse
+import os
+import json
+import sys
+import subprocess
+import urllib.request
+from enum import Enum, auto
+from pathlib import Path
+
+exec_prefix = os.environ.get("EXEC_WRAPPER", "")
+exec_prefix = exec_prefix.split(" ") if exec_prefix != "" else []
+
+# Pinned to a specific commit (2026-04-21).
+WYCHEPROOF_COMMIT = "744697a257d41941320b4ea10987b85b25f8c8bc"
+WYCHEPROOF_BASE_URL = f"https://raw.githubusercontent.com/C2SP/wycheproof/{WYCHEPROOF_COMMIT}/testvectors_v1"
+
+WYCHEPROOF_FILES = [
+    "mldsa_44_sign_seed_test.json",
+    "mldsa_44_sign_noseed_test.json",
+    "mldsa_44_verify_test.json",
+    "mldsa_65_sign_seed_test.json",
+    "mldsa_65_sign_noseed_test.json",
+    "mldsa_65_verify_test.json",
+    "mldsa_87_sign_seed_test.json",
+    "mldsa_87_sign_noseed_test.json",
+    "mldsa_87_verify_test.json",
+]
+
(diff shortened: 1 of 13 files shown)


test/wycheproof/wycheproof_client.py
```diff
@@ -0,0 +1,350 @@
...
+class TestResult(Enum):
+    OK = auto()
+    SKIPPED = auto()
+
+
+def check_sign_result(tc, out):
+    if tc["result"] == "invalid":
+        # InvalidPrivateKey: implementation does not validate sk
+        if "InvalidPrivateKey" in tc.get("flags", []):
+            return TestResult.SKIPPED
```
hanno-becker, line 130: Could we add a mode to the C client for `pk_from_sk` and invoke that here instead?

hanno-becker (changes requested):
Looks good overall. Since we download untrusted data from the web, we need to ensure that `wycheproof_client.py` and `wycheproof_mldsa.c` can safely deal with potentially malicious data -- which seems to be the case. The same holds for ACVP.

One smaller request for improvement: Where Wycheproof expected "Invalid SK" failure, can we invoke `pk_from_sk` and check that this would fail?

mkannwischer:
> 
> One smaller request for improvement: Where Wycheproof expected "Invalid SK" failure, can we invoke `pk_from_sk` and check that this would fail?

I would find it a little confusing to call `pk_from_sk` (which is essentially keygen), in a signature test case.
What I would find more acceptable is to make the secret key validation API part of the public API, and call that one instead prior to calling signing. But I don't think this is very important as secret key validation is not described in FIPS204 and I have not yet seen an implementation that implements it as a part of signing.

hanno-becker:
> I would find it a little confusing to call pk_from_sk (which is essentially keygen), in a signature test case.
What I would find more acceptable is to make the secret key validation API part of the public API, and call that one instead prior to calling signing. But I don't think this is very important as secret key validation is not described in FIPS204 and I have not yet seen an implementation that implements it as a part of signing.

But with the current approach, we just ignore the invalid-SK tests entirely -- even though we do have code which means to catch them. Calling `pk_from_sk` fills that test gap, and with appropriate comment it should not be confusing either; we can still include a note about potentially introducing a key-checking API in the future.

hanno-becker:
Reflecting private discussion: It makes sense to use the Wycheproof vectors to exercise our invalid-SK checking in `pk_from_sk`, but it should not be done under the umbrella of the signing test, because that can wrongly leave the impression that SK validation is part of signing. Instead, we may want to filter the invalid-SK tests from the JSON and run them as a separate class of tests. @mkannwischer Are you going to adjust the PR accordingly?

mkannwischer:
> Reflecting private discussion: It makes sense to use the Wycheproof vectors to exercise our invalid-SK checking in `pk_from_sk`, but it should not be done under the umbrella of the signing test, because that can wrongly leave the impression that SK validation is part of signing. Instead, we may want to filter the invalid-SK tests from the JSON and run them as a separate class of tests. @mkannwischer Are you going to adjust the PR accordingly?

Yes, I can adapt this accordingly, but I may not get to it before the weekend. Feel free to pick it up in the meantime if you like.

hanno-becker:
We have too many PRs; let's not make the perfect the enemy of the good. I created https://github.com/pq-code-package/mldsa-native/issues/1078 to reflect our discussion, and the addition of further SK validation tests derived from Wycheproofs can be a separate PR.

test/wycheproof/wycheproof_client.py
hanno-becker: Deferred; https://github.com/pq-code-package/mldsa-native/issues/1078

hanno-becker (approved):

PR #1065, opened 2026-04-28
CBMC: Auto-derive lazy/eager suffix in proof Makefiles

hanno-becker:
Add a LAZY_EAGER_FUNCTIONS list to Makefile.common that automatically appends _lazy or _eager to USE_FUNCTION_CONTRACTS and CHECK_FUNCTION_CONTRACTS entries based on MLD_CONFIG_REDUCE_RAM.

This allows proof Makefiles to use base function names (e.g., mld_polyvec_matrix_pointwise_montgomery) instead of repeating ifdef/else/endif blocks for each lazy/eager variant.


mkannwischer (approved):
Thanks @hanno-becker - this LGTM.

Is it worthwhile to have the CBMC comments in the PRs print if the eager or the lazy proof was checked?

PR #1066, opened 2026-04-28
CBMC: Add contracts and proofs for yvec init/get_poly

mkannwischer:
I forgot to change those to be called by contract in https://github.com/pq-code-package/mldsa-native/pull/1031.


hanno-becker (dismissed):

mldsa/src/polyvec_lazy.c
```diff
@@ -318,6 +307,28 @@ void mld_polyvec_matrix_pointwise_montgomery_yvec_lazy(mld_polyveck *w,
 
 #endif /* MLD_CONFIG_REDUCE_RAM || MLD_UNIT_TEST */
 
+MLD_INTERNAL_API
+void mld_polyvec_matrix_pointwise_montgomery(mld_polyveck *t, mld_polymat *mat,
```
hanno-becker, line 314: So far we don't have any generic functionality in `polyvec_lazy.c`, so we should move this probably. Can be done in one of the follow-up PRs.

mldsa/src/polyvec_lazy.h
```diff
@@ -525,20 +523,36 @@ __contract__(
   assigns(memory_slice(mat, sizeof(mld_polymat_lazy)))
 );
 
+/*************************************************
+ * Name:        mld_polyvec_matrix_pointwise_montgomery_row_lazy
+ *
+ * Description: Compute one row of the matrix-vector product in Montgomery
+ *              form (NTT domain), with the result reduced to abs_bound Q:
```
hanno-becker: Nit: "reduced to absolute bound MLDSA_Q". Can be done in a follow-up.

test/src/test_unit.c
```diff
@@ -860,19 +860,21 @@ static int test_polyvec_lazy_eager(void)
...
-    CHECK(memcmp(&tk_eager, &tk_lazy, sizeof(mld_polyveck)) == 0);
+    for (i = 0; i < MLDSA_K; i++)
+    {
+      mld_polyvec_matrix_pointwise_montgomery_row_eager(&poly_eager, &mat_eager,
+                                                        &v, i);
+      mld_polyvec_matrix_pointwise_montgomery_row_lazy(&poly_lazy, &mat_lazy,
+                                                       &v, i);
+      /* Compare mod q */
+      mld_poly_caddq(&poly_eager);
+      mld_poly_caddq(&poly_lazy);
```
hanno-becker, line 879: Why is this necessary? Aren't we expecting strict functional equivalence?

mkannwischer: No, because one can be asm?

hanno-becker: Right

mldsa/src/polyvec_lazy.c
mkannwischer: Moved to polyvec.c

mkannwischer: Actually, no. This is bad. We can't move the protyotype to polyvec.h because that would lead to a circular dependency polyvec_lazy.h -> polyvec.h -> polyvec_lazy.h. All solutions I can think of are hacky. Moving back to polyvec_lazy.c

mldsa/src/polyvec_lazy.h
mkannwischer: Rephrased the comment to be more aligned with the rest of the repo.

hanno-becker: Thanks!

hanno-becker (approved):
LGTM

PR #1067, opened 2026-04-28
Make poly_pointwise_montgomery destructive

mkannwischer:
Rework mld_poly_pointwise_montgomery from (c, a, b) to in-place (a, b) where a := a * b * R^{-1} mod q. The change propagates through the C wrapper, the native dispatcher, the Arm aarch64 and x86_64 AVX2 assembly, all call sites, the test harness, and the allocation constants.

This allows eliminating one temporary polynomial in the the REDUCE_RAM definition in polymat, and it will also enable further memory optimizations in follow-up PRs.

CBMC and HOL Light proofs are not yet updated.

- Depends on #1066


hanno-becker (approved):
In `sign.c` it seems that perhaps further memory optimizations can be unlocked -- I didn't check in detail -- but this may be what you're alluding to in the PR description?

Otherwise, this looks clean to me. Sweet that the proofs just continue working.

mkannwischer:
> In `sign.c` it seems that perhaps further memory optimizations can be unlocked -- I didn't check in detail -- but this may be what you're alluding to in the PR description?
> 
> Otherwise, this looks clean to me. Sweet that the proofs just continue working.

Yes, one of the optimizations will be done for verification in https://github.com/pq-code-package/mldsa-native/pull/1007.

PR #1037, opened 2026-04-13
HOL-Light: Add HOL Light poly_use_hint proofs for AArch64

jakemas:
Resolves:
- https://github.com/pq-code-package/mldsa-native/issues/486

### Summary

Proof times are decent `~15min for _32` and `~22min for _88`.

- Port ML-DSA poly_use_hint HOL Light proofs from s2n-bignum for AArch64
  - `mldsa_poly_use_hint_32`: GAMMA2=(Q-1)/32, for l=5,7 (s2n-bignum PR: https://github.com/awslabs/s2n-bignum/pull/372)
  - `mldsa_poly_use_hint_88`: GAMMA2=(Q-1)/88, for l=4 (s2n-bignum PR: https://github.com/awslabs/s2n-bignum/pull/375)
- Includes correctness proofs, subroutine forms, and constant-time/memory safety proofs for both
- Adds CBMC contracts for the assembly functions

PR #1070, opened 2026-04-29
Make poly_use_hint destructive

mkannwischer:
Rework mld_poly_use_hint, its native dispatchers, the underlying
aarch64/x86_64 implementations, and the polyveck wrapper from
(b, a, h) to in-place (b, a) where b holds the hint polynomial on
entry and the corrected high bits on exit — i.e. the first output
and second input now share a buffer.

This will enable reducing memory consumption in #1007. 

- Depends on #1066 
- Depends on #1067
- Depends on #1037


hanno-becker (approved):
LGTM, thank you @mkannwischer

PR #1076, opened 2026-04-30
x86_64: Fix wrong qbound 87*q -> 31*q in use_hint_32

### LINKED ISSUE #1074: Potentially wrong value for q_bound in poly_use_hint_32_avx2.c
In dev/x86_64/src/poly_use_hint_32_avx2.c, the constant initialization
```
const __m256i q_bound = _mm256_set1_epi32(87 * ((MLDSA_Q - 1) / 32));
```
does not look correct. The `87` multiplier should be `31`

This mirrors the `87 * ((MLDSA_Q - 1) / 88)` pattern from `poly_use_hint_88_avx2.c` — both the mismatched multiplier 87 and the divisor 32
for this variant. For the 32-variant, the "wrap-around" threshold should be `31 * ((MLDSA_Q - 1) / 32)` (i.e., Q - γ2), matching
`poly_decompose_32_avx2.c`. As written, q_bound = 87 * 261888 = 22784256, which is strictly greater than Q = 8380417, so the
`_mm256_cmpgt_epi32(f, q_bound)` comparison is identically false — the wrap-around correction branch is dead code.
  
Why it doesn't break output: for f > 31·γ2, the raw f1 lands at 16 instead of being corrected to 0, but the final `f1 =
_mm256_and_si256(f1, mask=15)` masks 16 → 0. And the (f0 > 0) ? h : -h selector still agrees with what the corrected f0 would choose,
because the uncorrected and corrected f0 differ by exactly 1 (the spec's a0 -= 1), and in the wrap range f0 is already negative, so both
versions fall on the f0 <= 0 (non-positive) side. I verified this empirically by exhaustively simulating mld_poly_use_hint_32_avx2 vs
mld_poly_use_hint_c for (a, h) ∈ [0, Q) × {0, 1} — zero mismatches.
  
So the code is functionally correct today, but the q_bound is misleading and the safety margin is thin. If the mask width or the f0
heck ever changes, the bug becomes a real defect. The fix is trivial: `q_bound = 31 * ((MLDSA_Q - 1) / 32)`.

mkannwischer:
use_hint_32 in the AVX2 backend contains a wrong constant that was copied from use_hint_88.
This disabled the check if w1=16 after the decompose step. Luckily, this cancels out later on in the use_hint for non-trivial reasons. We have also exhaustively tested that this code produces correct results for all inputs.

We will be able to optimize this away in a follow-up commit. For now, this commit just corrects the constant to avoid confusion.

- Fixes https://github.com/pq-code-package/mldsa-native/issues/1074

### DIFF
--- a/dev/x86_64/src/poly_use_hint_32_avx2.c
+++ b/dev/x86_64/src/poly_use_hint_32_avx2.c
@@ -39,7 +39,7 @@ void mld_poly_use_hint_32_avx2(int32_t *b, const int32_t *a,
 {
   unsigned int i;
   __m256i f, f0, f1, h, t;
-  const __m256i q_bound = _mm256_set1_epi32(87 * ((MLDSA_Q - 1) / 32));
+  const __m256i q_bound = _mm256_set1_epi32(31 * ((MLDSA_Q - 1) / 32));
   /* check-magic: 1025 == floor(2**22 / 4092) */
   const __m256i v = _mm256_set1_epi32(1025);
   const __m256i alpha = _mm256_set1_epi32(2 * ((MLDSA_Q - 1) / 32));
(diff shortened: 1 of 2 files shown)


hanno-becker (approved):

jakemas:
Very interesting!

PR #1077, opened 2026-04-30
sign: Use MLD_ALLOC for all secret buffers in sign.c

mkannwischer:
As a follow-up to 64067ca5f252367d518f58b1c1380bd095e35562, vonvert the stack-allocated rnd/seed/pre buffers in the public sign API
wrappers to MLD_ALLOC so that consumers using a custom allocator
(MLD_CONFIG_CUSTOM_ALLOC_FREE) place all sensitive material through the
same allocation path.

Fixed-size, non-secret stack buffers (hpk, ph, message[1] in PCT) are
left on the stack so that all verify routines perform the same set of
MLD_ALLOC allocations (and likewise for sign), which keeps a single
MLD_TOTAL_ALLOC_*_VERIFY / *_SIGN constant accurate without one variant
inflating the bound for the others.


hanno-becker (changes requested):
I'm not sure this is the right move -- the cost of an alloc/free is negligible, less important IMO than uniformity for the allocation mechanism. For the security-critical RNG bytes in particular, one could imagine a use case where the consumer has registered a dedicated allocator from a "special" -- in some sense -- to hold the security sensitive material.

Can we rather go the other way, and strive to use MLD_ALLOC/MLD_FREE exclusively?

mkannwischer:
> I'm not sure this is the right move -- the cost of an alloc/free is negligible, less important IMO than uniformity for the allocation mechanism. For the security-critical RNG bytes in particular, one could imagine a use case where the consumer has registered a dedicated allocator from a "special" -- in some sense -- to hold the security sensitive material.
> 
> Can we rather go the other way, and strive to use MLD_ALLOC/MLD_FREE exclusively?

I don't have a strong opinion which way - I just wanted it to be consistent and the majority of tiny buffers were on the stack.
Happy to change. ~I think the same applies to mlkem-native.~ No changes needed in mlkem-native.

mldsa/src/sign.c
```diff
@@ -1173,9 +1187,15 @@ int mld_sign_verify(const uint8_t *sig, size_t siglen, const uint8_t *m,
                     const uint8_t pk[MLDSA_CRYPTO_PUBLICKEYBYTES],
                     MLD_CONFIG_CONTEXT_PARAMETER_TYPE context)
 {
-  MLD_ALIGN uint8_t pre[MLD_DOMAIN_SEPARATION_MAX_BYTES];
   size_t pre_len;
   int ret;
+  MLD_ALLOC(pre, uint8_t, MLD_DOMAIN_SEPARATION_MAX_BYTES, context);
+
```
hanno-becker: Let's follow the PR description "Fixed-size, non-secret stack buffers (hpk, ph, message[1] in PCT) are
left on the stack" and restrict the change to `seed` and `rng` for now.

hanno-becker (changes requested):
Thank you @mkannwischer! Good to have that unified for the secret seed and rng buffers. As per the PR description, I think the public, small + statically sized domain separation prefix can be continued to be allocated on the stack for now.

hanno-becker (approved):

PR #1087, opened 2026-05-02
Test: Use Wycheproof signing tests to validate pk_from_sk

### LINKED ISSUE #1078: Test: Use invalid-SK Wycheproof signing tests to validate `pk_from_sk`
This was originally discussed in https://github.com/pq-code-package/mldsa-native/pull/1063

We don't conduct any SK validation as part of signing, and are skipping Wycheproofs tests which exercise invalid SKs.

We do instead implement SK validation as part of `pk_from_sk`. We have an opportunity here to test the coverage of those tests by extracting from the meant-to-be signing Wycheproofs tests some testvectors for `pk_from_sk`.

**Task:** Extend the Wycheproof test to filter the invalid-SK tests from the signing test set, and run a separate test feeding them into `pk_from_sk`.

hanno-becker:
Add a pkFromSk command to the Wycheproof C driver and a corresponding test runner in the Python client. This runs pk_from_sk on every test group from the noseed signing tests:

- InvalidPrivateKey: assert pk_from_sk rejects (non-zero exit)
- IncorrectPrivateKeyLength: assert decode_error from the driver
- Valid SK: assert pk_from_sk succeeds and derived PK matches expected

Also tighten check_sign_result to assert the exact expected failure mode per flag, rather than accepting any error. Unrecognised invalid flags now fail loudly.

Resolves #1078.

### DIFF
--- a/test/wycheproof/wycheproof_client.py
+++ b/test/wycheproof/wycheproof_client.py
@@ -85,16 +85,16 @@ def detect_supported_modes():
             err(
                 f"Warning: {wycheproof_call} failed (rc={result.returncode}), assuming all modes supported"
             )
-            return {"keyGen", "sigGen", "sigVer"}
+            return {"keyGen", "sigGen", "sigVer", "pkFromSk"}
         modes = set()
         for line in result.stdout.splitlines():
             line = line.strip()
-            if line in ("keyGen", "sigGen", "sigVer"):
+            if line in ("keyGen", "sigGen", "sigVer", "pkFromSk"):
                 modes.add(line)
         return modes
     except FileNotFoundError:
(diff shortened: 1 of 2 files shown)


mkannwischer (approved):
Thanks for adding this, @hanno-becker! Sorry that I could not find the time yet to do it.

Only thing that is bothering me a little is that the skipping happpens in the `check_result` step rather than before running the test binary. Do you think we should change that?

hanno-becker:
Thank you for the swift review!

> Only thing that is bothering me a little is that the skipping happpens in the check_result step rather than before running the test binary. Do you think we should change that?

I don't think it really matters, so I'd be inclined to just leave it.

PR #1088, opened 2026-05-03
HOL-Light: Prefix imports to separate mldsa-native from s2n-bignum

### LINKED ISSUE #1086: HOL-Light: Do not use spec files from s2n-bignum
**Context:** mldsa-native and s2n-bignum share proofs and specifications for ML-DSA, but for standalone reproducibility, mldsa-native should only use the generic verification infrastructure from s2n-bignum, while maintaining specs + proofs separately.

**Issue:** https://github.com/pq-code-package/mldsa-native/pull/1069 passed CI even though `arm_mldsa_inverse_ntt` was not defined in mldsa-native's `mldsa_specs.ml`, strongly suggesting that, unintentionally, we import the eponymous homonymous file from s2n-bignum.

**Task:** Propose and implement a robust way to ensure that we never import spec or proof files from s2n-bignum.

hanno-becker:
* Fixes https://github.com/pq-code-package/mldsa-native/issues/1086

**Issue**: every `needs` directive in mldsa-native proof files resolved ambiguously against both $PROOF_DIR and $S2N_BIGNUM_DIR, so a file with the same relative path in either tree could silently shadow the other.

**Proposed resolution:** Every import now declares its source at the call site with a namespace prefix.

```
  needs "mldsa_native/common/mldsa_specs.ml";;   (* local *)
  needs "s2n_bignum/arm/proofs/base.ml";;        (* external *)
  needs "Library/words.ml";;                     (* HOL-Light stdlib *)
```

The nix shellHook creates a temporary .imports directory with symlinks `mldsa_native` and `s2n_bignum` pointing to the local resp. s2n-bignum directories, and HOLLIGHT_LOAD_PATH points at that import directory. The `mldsa_native/` and `s2n_bignum/` namespaces are disjoint by prefix, so path collisions become structurally impossible.

A new scripts/check-hol-light-imports enforces the rule across proofs/hol_light/**/*.ml and is wired into scripts/lint and both build-proof.sh scripts. s2n-bignum imports must additionally appear on proofs/hol_light/s2n_bignum_allowlist.txt, so new cross-repo dependencies surface in review.

A side-benefit of exposing proofs/hol_light/.imports is that it is easier to navigate to the nix-imported s2n-bignum sources for inspection.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -46,6 +46,12 @@
           holLightShellHook = ''
             export PATH=$PWD/scripts:$PATH
             export PROOF_DIR="$PWD/proofs/hol_light"
+            # Namespaced imports root for HOL-Light proofs.
+            # See scripts/check-hol-light-imports for the enforced rule.
+            export IMPORTS_DIR="$PROOF_DIR/.imports"
+            mkdir -p "$IMPORTS_DIR"
+            ln -sfn "$S2N_BIGNUM_DIR" "$IMPORTS_DIR/s2n_bignum"
+            ln -sfn "$PROOF_DIR"      "$IMPORTS_DIR/mldsa_native"
           '';
         in
         {
@@ -87,7 +93,7 @@
           packages.toolchain_aarch64_be = util.toolchain_aarch64_be;
           packages.gcc-arm-embedded = pkgs.gcc-arm-embedded;
 
-          devShells.default = util.mkShell {
+          devShells.default = (util.mkShell {
             packages = builtins.attrValues
               {
                 inherit (config.packages) linters cbmc hol_light s2n_bignum slothy toolchains_native hol_server;
@@ -96,7 +102,7 @@
                   nix-direnv
                   zig_0_13;
               } ++ pkgs.lib.optionals (!pkgs.stdenv.isDarwin) [ config.packages.valgrind_varlat ];
-          };
+          }).overrideAttrs (old: { shellHook = holLightShellHook; });
 
           packages.hol_server = util.hol_server.hol_server_start;
           devShells.hol_light = (util.mkShell {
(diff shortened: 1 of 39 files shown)


flake.nix
```diff
@@ -46,6 +46,12 @@
           holLightShellHook = ''
             export PATH=$PWD/scripts:$PATH
             export PROOF_DIR="$PWD/proofs/hol_light"
+            # Namespaced imports root for HOL-Light proofs.
+            # See scripts/check-hol-light-imports for the enforced rule.
+            export IMPORTS_DIR="$PROOF_DIR/.imports"
+            mkdir -p "$IMPORTS_DIR"
+            ln -sfn "$S2N_BIGNUM_DIR" "$IMPORTS_DIR/s2n_bignum"
+            ln -sfn "$PROOF_DIR"      "$IMPORTS_DIR/mldsa_native"
```
mkannwischer, line 54: This is only hooked up in the `hol_light` shell. Please also add to the default shell.

mkannwischer (changes requested):
Thanks @hanno-becker. I support this approach, but I think it should also work in the default shell.

flake.nix
hanno-becker: Hm yes, but why isn't this already an issue? This hook is pre-existing and not (it appears) applied to the default shell.

mkannwischer: PROOF_DIR indeed seems to be not set in the default shell, but it also does not seem to be needed.

hanno-becker: PROOF_DIR is used in the patches version of `hol.sh`

mkannwischer (approved):
Thanks! Works for me now.

PR #1089, opened 2026-05-03
Use heap allocation + valgrind in backend unit test

hanno-becker:
Port of [mlkem-native#1633 ](https://github.com/pq-code-package/mlkem-native/pull/1633) to mldsa-native.

* Replace aligned_alloc + MLD_ALIGN_UP with posix_memalign in custom_heap_alloc_config.h. Unlike aligned_alloc, posix_memalign does not require the size to be a multiple of the alignment, removing the need for MLD_ALIGN_UP rounding. This ensures that allocations are exact-sized, allowing memory-safety tests like valgrind and ASan to detect overflows at precise buffer boundaries. On Windows, _aligned_malloc is used instead. Also adds the missing configs.yml entry so the file is tracked by autogen.

* Replace all stack-allocated buffers in test_unit.c with heap allocations via MLD_ALLOC/MLD_FREE, using the custom_heap_alloc_config. This enables valgrind to detect buffer overflows in assembly backends, which operate on these buffers.

* Build the unit test objects with custom_heap_alloc_config.h by adding the appropriate -DMLD_CONFIG_FILE, -std=c11, and -D_GNU_SOURCE flags in components.mk, factored through a new UNIT_CFLAGS variable.

* Add unit_valgrind job to ci.yml that runs the unit tests under valgrind on x86_64 and aarch64 runners. This catches buffer overflows in hand-written assembly that ASan cannot detect, since ASan only instruments compiler-generated code.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -466,6 +466,45 @@ jobs:
           examples: false
           stack: true
           check_namespace: false
+  unit_valgrind:
+    name: Unit tests + valgrind (${{ matrix.target.name }}, ${{ matrix.cflags }})
+    strategy:
+      fail-fast: false
+      matrix:
+        external:
+         - ${{ github.repository_owner != 'pq-code-package' }}
+        target:
+          - runner: ubuntu-latest
+            name: x86_64
+          - runner: ubuntu-24.04-arm
+            name: aarch64
+        cflags: ['-O3', '-Os']
+        exclude:
+          - external: true
+    runs-on: ${{ matrix.target.runner }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Unit tests under valgrind
+        uses: ./.github/actions/functest
+        with:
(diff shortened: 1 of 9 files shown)


mkannwischer (changes requested):
Thanks @hanno-becker. The port looks good to me.
Can we also cover the rejction sampling, please?

The commit message could use updating:
```
* Build the unit test objects with custom_heap_alloc_config.h by adding
  the appropriate -DMLD_CONFIG_FILE, -std=c11, and -D_GNU_SOURCE flags
  in components.mk, factored through a new UNIT_CFLAGS variable.
```

hanno-becker:
@mkannwischer This is a pre-existing test gap? It should be addressed, but seems orthogonal to this PR.

hanno-becker:
Opened #1090 and addressed in the follow-up commit. Also fixed the commit message.

mkannwischer (approved):
Thanks @hanno-becker for adding that. LGTM. Happy if CI is.

hanno-becker:
@mkannwischer Multiplexed the unit test for `rej_uniform` to cover `N=1,2,3,4,5` blocks.

mkannwischer (approved):

PR #1095, opened 2026-05-04
Lowram: Refactor verify into per-row loop, drop polyveck w1/t1/h

mkannwischer:
Restructure mld_sign_verify_internal so that w1, t1 and h live as single mld_poly scratch buffers instead of full mld_polyveck vectors. Each row of A * z - c * t1 * 2^d, the use_hint correction and the polyw1 packing are now done one polynomial at a time.

Convert mld_unpack_pk_t1 and mld_sig_unpack_hints to per-row APIs (mld_unpack_pk_t1 takes the row index, mld_sig_unpack_hints decodes and validates one row at a time, with the trailing-zero check folded into the i == K-1 iteration).

Drop the now-unused mld_polyveck_{sub,shiftl,pointwise_poly_montgomery, use_hint} helpers and their CBMC harnesses, and narrow the guards on mld_polyveck_ntt and mld_polyveck_pack_w1 accordingly.


mldsa/src/packing.h
```diff
@@ -240,25 +243,37 @@ __contract__(
...
+ *              across all rows (a count array followed by a single
+ *              index list), so this function performs the validation
+ *              relevant to row i:
+ *                - the i'th hint count is non-decreasing and bounded
+ *                  by MLDSA_OMEGA;
+ *                - the indices for row i are strictly ascending;
+ *                - on i == MLDSA_K - 1, the trailing index slots are
+ *                  zero.
+ *              Callers must invoke this for i = 0, 1, ..., MLDSA_K - 1
+ *              in order; if any call returns non-zero the encoding is
```
hanno-becker: Why is it important that this is called in order?

mkannwischer: Right. It doesn't have to be called in order as long as you call it for every i. Will fix.
If you call it in order it behaves exactly as the reference implementation.

mldsa/src/packing.c
```diff
@@ -172,75 +166,51 @@ void mld_pack_sig_z(uint8_t sig[MLDSA_CRYPTO_BYTES], const mld_poly *zi,
...
+      decreases(MLDSA_OMEGA - j)
+    )
     {
-      const uint8_t this_hint_index = packed_hints[j];
-
-      /* Coefficients must be ordered for strong unforgeability */
-      if (j > old_hint_count && this_hint_index <= packed_hints[j - 1])
+      if (packed_hints[j] != 0)
       {
         return 1;
```
hanno-becker: While at it, can we change the return code to `MLD_ERR_FAIL`?

mkannwischer: Yes!

mldsa/src/packing.c
```diff
@@ -172,75 +166,51 @@ void mld_pack_sig_z(uint8_t sig[MLDSA_CRYPTO_BYTES], const mld_poly *zi,
...
-
-    /* new_hint_count must increase or stay the same, but also remain */
-    /* less than or equal to MLDSA_OMEGA                              */
-    if (new_hint_count < old_hint_count || new_hint_count > MLDSA_OMEGA)
+    if (j > old_hint_count && packed_hints[j] <= packed_hints[j - 1])
     {
-      /* Error - new_hint_count is invalid */
       return 1;
     }
+    h->coeffs[packed_hints[j]] = 1;
```
hanno-becker, line 199: While at it, it may be worth noting that this access is safe by virtue of `coeffs` having `256` entries, which happens to be the upper bound for `uint8_t`. Otherwise, explicit input validation would be needed.

hanno-becker (changes requested):
Some smaller comments, but overall it looks very good to me.

mldsa/src/packing.h
mkannwischer: Fixed.

mldsa/src/packing.c
mkannwischer: Fixed.

mldsa/src/packing.c
mkannwischer: Added a `Safety:` comment.

mldsa/src/sign.c
```diff
@@ -1137,33 +1117,51 @@ int mld_sign_verify_internal(const uint8_t *sig, size_t siglen,
...
+
+    /* w1 = invNTT(w1 - c * t1_i * 2^d) */
+    mld_poly_sub(w1, tmp);
+    mld_poly_reduce(w1);
+    mld_poly_invntt_tomont(w1);
+    mld_poly_caddq(w1);
+
+    /* tmp = h_i (decoded and validated from signature) */
+    ret = mld_sig_unpack_hints(tmp, sig, i);
+    if (ret)
```
hanno-becker: Nit: We have snuck two instances of `if (ret)` in `sign.c` lately, where predominantly we used `if (ret != 0)`. Should be unified. Up to you to do it here or in a follow-up.

hanno-becker (changes requested):
Minor error-check inconsistency, but not blocking. LGTM

mldsa/src/sign.c
mkannwischer: Will fix. Could be linted - I still have on my list to revisit codeql.

hanno-becker: > Will fix. Could be linted - I still have on my list to revisit codeql.

Indeed. Same for safety comments. I'm still keen on CodeQL ;-)

hanno-becker (approved):

PR #1105, opened 2026-05-07
nix: Anchor HOL-Light shellHook to repo root, not $PWD

hanno-becker:
The holLightShellHook set PROOF_DIR (and thus IMPORTS_DIR) from $PWD, which is whatever directory the user ran `nix develop` from. Entering the shell from a subdirectory (e.g. proofs/cbmc/) caused PROOF_DIR to resolve to a non-existent nested path, and the .imports/ directory plus its `mldsa_native` / `s2n_bignum` symlinks were created under that wrong location -- with `mldsa_native` pointing at a path that does not exist.

Resolve the repo root via `git rev-parse --show-toplevel` (matching the pattern already used in nix/hol_light/hol-server.sh), with a $PWD fallback for non-git invocations, so PATH / PROOF_DIR / IMPORTS_DIR are stable regardless of the invocation cwd.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -44,8 +44,13 @@
             ];
           };
           holLightShellHook = ''
-            export PATH=$PWD/scripts:$PATH
-            export PROOF_DIR="$PWD/proofs/hol_light"
+            # Resolve the repo root independently of the directory from which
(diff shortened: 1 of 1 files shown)


mkannwischer (approved):
Thanks @hanno-becker. Works.

PR #1107, opened 2026-05-07
Doc: Improve documentation of externalmu and relax its CBMC contract

mkannwischer:
The externalmu flag on sign_signature_internal/sign_verify_internal was only described as "indicates input message m is processed as mu", which left readers guessing about the meaning of 0 vs 1, the constraints on mlen/pre/prelen, and how to compute mu for the *_extmu wrappers. Expand the doc comments in mldsa_native.h and sign.h to spell out both modes, and describe what mu is (SHAKE256(tr || M', 64)) for the *_extmu APIs.

Also relax the CBMC contract from externalmu == 1 to externalmu != 0 to match the implementation.

Thanks to @nicolabalzano for reporting that the documentation is unclear.

### DIFF
--- a/mldsa/src/sign.h
+++ b/mldsa/src/sign.h
@@ -181,15 +181,28 @@ __contract__(
  *              - size_t *siglen:            pointer to output length of
  *                                           signature
  *              - const uint8_t *m:          pointer to message to be signed
- *              - size_t mlen:               length of message
- *              - const uint8_t *pre:        pointer to prefix string
- *              - size_t prelen:             length of prefix string
+ *                                           (when externalmu == 0), or to a
+ *                                           precomputed message
+ *                                           representative mu (when
+ *                                           externalmu != 0).
+ *              - size_t mlen:               length of m. Must equal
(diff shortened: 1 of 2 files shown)


mldsa/src/sign.h
```diff
@@ -413,8 +440,8 @@ __contract__(
   requires(siglen <= MLD_MAX_BUFFER_SIZE)
   requires(memory_no_alias(sig, siglen))
   requires(memory_no_alias(m, mlen))
-  requires(externalmu == 0 || (externalmu == 1 && mlen == MLDSA_CRHBYTES))
-  requires(externalmu == 1 || prelen == 0 || memory_no_alias(pre, prelen))
+  requires(externalmu == 0 || mlen == MLDSA_CRHBYTES)
+  requires(externalmu != 0 || prelen == 0 || memory_no_alias(pre, prelen))
```
hanno-becker: This is easier to read as an implication

mkannwischer: Fixed.

hanno-becker (approved):

PR #1109, opened 2026-05-08
HOL-Light: Unify *_SAFE proofs to use full_spec directly

mkannwischer:


hanno-becker (changes requested):
Hmm, I am not yet convinced this is an improvement -- spelling out the specification is more accessible for the reader than asking them to look up the `mk_safety_spec`. What was your motivation for this change, @mkannwischer?

mkannwischer:
> Hmm, I am not yet convinced this is an improvement -- spelling out the specification is more accessible for the reader than asking them to look up the `mk_safety_spec`. What was your motivation for this change, @mkannwischer?

Sorry, I was a little too eager in removing elaborate specs. As discussed offline, we should spell out the safety spec fully in `SUBROUTINE_SAFE`. The other safety specs should be shortened or eliminated.

PR #1116, opened 2026-05-12
CBMC: Improve readability of CBMC specs

hanno-becker:
CBMC contracts in this repo currently require one instance of `requires(memory_no_alias(...))` per pointer argument and a
parallel set of `memory_slice(...)` targets in `assigns(...)`. For functions dealing with multiple pointers, this quickly bloats and thereby impedes the readability of the CBMC specs.

Introduce two variadic preprocessor macros in mldsa/src/cbmc.h that unfold a list of regions into the corresponding memory_no_alias / memory_slice calls, joined as required by their containing clause:

```
  requires(disjoint(arg, ...))   -- AND-joined memory_no_alias checks
  assigns(..., slices(arg, ...)) -- comma-joined memory_slice targets
```

Each arg is either a bare object x (size `sizeof(*x)`) or a pair (x, N) where N is an element count (size `sizeof(*x) * N`). The pair form covers byte buffers like `uint8_t seed[CRHBYTES]` and typed arrays where the count comes from a parameter rather than the type.

Implementation relies on some C99 preprocessor magic.

### DIFF
--- a/dev/aarch64_clean/src/arith_native_aarch64.h
+++ b/dev/aarch64_clean/src/arith_native_aarch64.h
@@ -66,11 +66,11 @@ void mld_ntt_aarch64_asm(int32_t *r, const int32_t *zetas_l123456,
 /* This must be kept in sync with the HOL-Light specification
  * in proofs/hol_light/aarch64/proofs/ntt_aarch64_asm.ml */
 __contract__(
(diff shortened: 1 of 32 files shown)


hanno-becker:
@mkannwischer This is _not_ ready for detailed review as is, but I would like to hear your thoughts on the general direction and what specific argument/macro shape you find the most appealing.

PR #1117, opened 2026-05-12
CBMC: Introduce MLD_ANY_ERROR helper for error disjunctions

### LINKED ISSUE #1096: Consider abbreviating enumeration of valid error codes
With the increase in the number of error codes, top-level contracts are becoming harder to read due to repetitive clauses like `ensures(return_value == ERR_A || return_value == ERR_B || ...)`. Consider introducing an abbreviation like `any_error` or similar, so capture the common error enumeration pattern.

hanno-becker:
Top-level contracts that admit every MLD_ERR_* failure code were repeating a 3-line disjunction at each call site. Introduce MLD_ANY_DEFINED_ERROR(err) in common.h (next to the error-code definitions) and apply it to the 6 full-error-set contracts in sign.h / sign.c.

Narrower contracts that only admit a subset of errors are left untouched: their restricted error set is part of their specification.

Fixes #1096.

### DIFF
--- a/mldsa/mldsa_native.c
+++ b/mldsa/mldsa_native.c
@@ -221,6 +221,7 @@
 /* mldsa/src/common.h */
 #undef MLD_ADD_PARAM_SET
 #undef MLD_ALLOC
+#undef MLD_ANY_ERROR
 #undef MLD_APPLY
 #undef MLD_ASM_FN_SIZE
 #undef MLD_ASM_FN_SYMBOL
--- a/mldsa/mldsa_native_asm.S
+++ b/mldsa/mldsa_native_asm.S
@@ -228,6 +228,7 @@
 /* mldsa/src/common.h */
 #undef MLD_ADD_PARAM_SET
 #undef MLD_ALLOC
+#undef MLD_ANY_ERROR
 #undef MLD_APPLY
 #undef MLD_ASM_FN_SIZE
 #undef MLD_ASM_FN_SYMBOL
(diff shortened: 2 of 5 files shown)


hanno-becker:
@mkannwischer This is about your comment in https://github.com/pq-code-package/mldsa-native/pull/1059#discussion_r3188924921

mkannwischer:
I support this change. Thank you @hanno-becker.
What was the reason not to use `MLD_ANY_ERROR` as suggested in the PR title?

hanno-becker:
@mkannwischer I felt `MLD_ANY_ERROR` could be misread as "any non-zero value", while `MLD_ANY_DEFINED_ERROR` seemed to make it more explicit that we want one of a defined list of errors. But it's a minor point.

mkannwischer:
> @mkannwischer I felt `MLD_ANY_ERROR` could be misread as "any non-zero value", while `MLD_ANY_DEFINED_ERROR` seemed to make it more explicit that we want one of a defined list of errors. But it's a minor point.
> @mkannwischer I felt `MLD_ANY_ERROR` could be misread as "any non-zero value", while `MLD_ANY_DEFINED_ERROR` seemed to make it more explicit that we want one of a defined list of errors. But it's a minor point.

Hmm, I prefer `MLD_ANY_ERROR` - we shouldn't have any error codes that are not in that list anyway?

hanno-becker:
> Hmm, I prefer MLD_ANY_ERROR - we shouldn't have any error codes that are not in that list anyway?

Okay.

mkannwischer (changes requested):
`autogen` needs to be run. Rest LGTM. Thanks @hanno-becker.

mkannwischer (approved):

PR #1118, opened 2026-05-12
Speed up C-reference NTT/invNTT with twisted zetas + 2-layer merging

hanno-becker:
Replace the single-layer C-reference forward and inverse NTT in
`mldsa/src/poly.c` with one that merges two layers each.

Also, store each twiddle alongside its precomputed twist, letting
`mld_fqmul(a, b, b_twisted)` drop the multiply with MLDSA_Q^{-1}
that was previously hidden inside `mld_montgomery_reduce`.

Mirrors pq-code-package/mlkem-native/463 (@rod-chapman) and pq-code-package/mlkem-native/683


proofs/cbmc/fqmul/Makefile
rod-chapman, line 29: On my laptop, proof of this new implementation takes 131s with bitwuzla, so I tried z3, which compltes the proof in about 26s. Suggest switch to `CBMCFLAGS=--smt2`

rod-chapman (commented):
Look good. 1 suggestion to improve proof times.

rod-chapman:
22nd June updates.

Proof of mld_fqscale() fails as expected, since it does not meet the requirement to return a value bounded by MLDSA_Q, following the relaxation of the bound returned by mld_fqmul().

All other proofs, including those of merged NTT and INTT, are OK for all parameter sets, with and without MLD_CONFIG_REDUCE_RAM

PR #1122, opened 2026-05-18
CBMC: Add support for using multiple solvers and data models

hanno-becker:
This PR adds infrastructure to run CBMC proofs in any combination of data model and solver configuration.

Each harness declares the `data-model:solver-profile` combinations known to work and identifies one as default:

```make
CBMC_SUPPORTED_CONFIGS = LP64:z3 LP64:bitwuzla
CBMC_DEFAULT_CONFIG = LP64:z3
```

Normal runs remain restricted to these combinations. `--explore` allows to experiment with configurations not known to work.

### Solver IDs

Solver IDs now identify solvers + configurations rather than only solvers. The central registry in `proofs/cbmc/Makefile.common` maps each profile to its backend, executable, and options.

| Profile | Configuration |
|---|---|
| `z3` | CBMC SMT2 backend using Z3 |
| `z3_smt_only` | Z3 with `tactic.default_tactic=smt` |
| `z3_no_bv_extract` | Z3 with `rewriter.bv_le2extract=false` |
| `bitwuzla` | CBMC Bitwuzla backend |
| `cvc5_arrays_exp` | cvc5 with `--arrays-exp` |

The cvc5 wrapper is needed because CBMC's CPROVER library emits `STORE_ALL`, which cvc5 accepts only with experimental array support. cvc5 is added to the CBMC nix environment.

For now, LP64 continues to be the only data model. @rod-chapman is working on 32-bit support. The new structure  merely allows further data models to be added later.

### Selecting a configuration (data model + solver ID)

Both selectors accept `default`, `all`, or a comma-separated list:

```text
--dm default|all|DM[,DM...]
--solver default|all|PROFILE[,PROFILE...]
```

For each proof, the driver:

1. Resolves `default` from that proof's `CBMC_DEFAULT_CONFIG`.
2. Forms candidate combinations from the selected DMs and profiles.
3. Runs candidates present in `CBMC_SUPPORTED_CONFIGS`.
4. Reports other candidates as omitted.
5. Runs undeclared candidates only when `--explore` is present.

This treats `CBMC_SUPPORTED_CONFIGS` as a relation rather than assuming that every selected data model works with every selected profile.

## Usage

Run one proof using its default:

```sh
tests cbmc --mldsa-parameter-set 44 -p ct_abs_i32
```

Select a subset of profiles:

```sh
tests cbmc --mldsa-parameter-set 44 --solver z3,z3_smt_only
```

Normal mode runs only matching supported configurations. Selected combinations not declared by a proof are reported as omitted.

Run every declared configuration, as CI does:

```sh
tests cbmc --mldsa-parameter-set 44 --dm all --solver all
```

Despite the `all` selectors, this does not run unsupported combinations. Currently each proof declares one supported tuple, so CI executes one configuration per proof and reports the other registered candidates as omitted.

Explore an undeclared profile for one proof:

```sh
tests cbmc --mldsa-parameter-set 44 -p ct_abs_i32 --dm LP64 --solver bitwuzla --explore
```

Explore every registered profile:

```sh
tests cbmc --mldsa-parameter-set 44 -p ct_abs_i32  --dm all --solver all --explore
```

## Declaring proof support

A proof supporting more than one profile can declare:

```make
CBMC_SUPPORTED_CONFIGS = LP64:z3 LP64:bitwuzla
CBMC_DEFAULT_CONFIG = LP64:z3
```

`CBMC_DEFAULT_CONFIG` must be a member of `CBMC_SUPPORTED_CONFIGS`. Unknown data models and profile IDs are rejected.

Proof-specific options that are independent of the selected profile remain local:

```make
CBMC_CONFIG_FLAGS = --big-endian
```

Executable and solver-specific options belong in the central profile registry rather than individual proof Makefiles.

## CI behavior

CI omits `--explore`, so only expected-to-succeed configurations execute.


hanno-becker:
@nmouha I'm continuing here with the exploration of adding CVC5 support alongside the existing solvers (bitwuzla, z3) that we use. Once we sort this, we can resume discussion on your proposal in #1114 and the corresponding change in mlkem-native.

rod-chapman:
1. The `--solver` and `--default-solver-only` switches might be replaced with a single
switch.

Something like

  `--solver=default|all|comma-separatest-list-of-solver-ids`

If not given, then `--solver=default` is the default.

2. What is a "solver ID" anyway? Can I use one of the names of the wrapper
scripts that we have in proofs/cbmc/lib ?

For example, can I say
  `--solver=z3_smt_only`

to get proofs/cbmc/lib/z3_smt_only

Perhaps we should have a wrapper script called "z3" in the "lib" directory
that invokes z3 with no options at all - then we can insist that a "solver id"
really is the name of a wrapper script.

Then I could also say

 `--solver=z3,z3_smt_only,z3_something_else`

to experiment with z3 options.

2. In the summary output, I need to know if the selected solver on each line of
output really is the default solver that would have been selected with --solver=default.

I suggest adding a "*" on the end of the solver ID to indicate "this one is the current default"

3. I also need to be able to easily see which solver is fastest, when the number of selected
solvers is >= 2. Combined with the suggestion above, I can easily tell if the current
defaut solver is the fastest of those chosen. Colour highlight the solver-id perhaps to indicate the winner?

4. In the summary output table when numnber of solvers >= 2, could you
add a blank line between the group of results for each function? That makes it easier
for my eye to see the results for each function in groups.

For example, output might look like this:
```
| Proof           | Solver   | Status       | Duration (in s) |
|-----------------|----------|--------------|-----------------|
| unpack_pk_t1    | BITWUZLA | Success      | 13              |
| unpack_pk_t1    | CVC5     | Inconclusive | 2               |
| unpack_pk_t1    | Z3*      | Success      | 3               | // "Z3*" in GREEN for "winner" ???
|                 |          |              |                 |
| unpack_sk       | BITWUZLA | Success      | 230             |
| unpack_sk       | CVC5     | Inconclusive | 5               |
| unpack_sk       | Z3*      | Success      | 3               |
|                 |          |              |                 |
| unpack_sk_s1hat | BITWUZLA | Fail         | 284             |
| unpack_sk_s1hat | CVC5     | Inconclusive | 4               |
| unpack_sk_s1hat | Z3*      | Success      | 3               |
```

hanno-becker:
@rod-chapman I followed your suggestion from https://github.com/pq-code-package/mldsa-native/pull/1122#issuecomment-4543864334 and introduced a notion of solver-ID which combines solver executable and flags.

.github/actions/cbmc/action.yml
```diff
@@ -66,7 +66,7 @@ runs:
             data_dir="dev/cbmc-reduce-ram"
             variant="reduce-ram"
           fi
-          tests cbmc --mldsa-parameter-set ${{ inputs.mldsa_parameter_set }} --per-proof-timeout 1800 --output-result-json ${{ github.workspace }}/cbmc_result_${{ inputs.mldsa_parameter_set }}.json $reduce_ram_flag || cbmc_failed=true
+          tests cbmc --mldsa-parameter-set ${{ inputs.mldsa_parameter_set }} --dm all --solver all --per-proof-timeout 1800 --output-result-json ${{ github.workspace }}/cbmc_result_${{ inputs.mldsa_parameter_set }}.json $reduce_ram_flag || cbmc_failed=true
```
rod-chapman, line 69: So "all" here (twice) really means "everything specified by the Makefile for each function", not "every DM and every prover", right?

proofs/cbmc/lib/cvc5_arrays_exp
```diff
@@ -0,0 +1,13 @@
...
+
+# Enable the experimental array theory in cvc5. CBMC encodes parts of
+# the CPROVER library using STORE_ALL, which cvc5 only accepts under
+# --arrays-exp.
+#
+# `exec` is load-bearing: without it bash sits between cbmc and cvc5,
+# holds the stdout FD open after cvc5 closes it, and cbmc's reader
+# stops mid-stream and treats every property verdict as ERROR even
+# though cvc5 returned `unsat`.
+exec cvc5 --arrays-exp "$@"
```
rod-chapman, line 13: I have also found that "--enum-inst" is essential to get useful results from cvc5 with SMT produced by CBMC. Note that from cvc5 1.4.0 onwards, so-called "safe" and "stable" builds of cvc5 will not allow --arrays-exp to be used at all. For that to work, we need an "unrestricted" build of cvc5.

In the long run, we need "--arrays-exp" to become allowed as a "safe" (and probably default) option in future versions of cvc5.

rod-chapman (commented):
Looks good. 2 comments.

I have not looked at the Python code in detail, so someone else might look at that.

.github/actions/cbmc/action.yml
hanno-becker: Yes. Unless you specify `--explore`, then literally _everything_ is attempted. This would be useful if you want to see what configurations work for a particular function.

proofs/cbmc/lib/cvc5_arrays_exp
hanno-becker: Yes, we can tweak that going forward.

nix/cbmc/default.nix
```diff
@@ -34,9 +35,9 @@ buildEnv {
           hash = "sha256-Lw037Z0t0ySxkgMXkbjNW5CB4QQLRrrSEBsLJqiomZ4=";
         };
       });
-
       inherit
         bitwuzla# 0.9.0
+        cvc5# 1.3.3 (taken from nixpkgs-unstable -- see flake.nix)
```
mkannwischer: This comment looks wrong.
nixpkgs-26.05 and current unstable both ship 1.3.4. Currently you are using the one from nixpkgs-26.05.

.github/workflows/cbmc.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       name: CBMC (ML-DSA-${{ matrix.parameter_set }}${{ matrix.reduce_ram && ', REDUCE_RAM' || '' }})
       ec2_instance_type: c9g.4xlarge
       ec2_ami: ubuntu (aarch64)
-      ec2_volume_size: 20
+      ec2_volume_size: 64
```
mkannwischer: Do we really need 64 GiB here?

.github/actions/cbmc/report.py
```diff
@@ -166,16 +266,19 @@ def sort_key(r):
         f"<!-- {COMMENT_TAG}(start): mldsa-{cfg.param_set}{cfg.tag_suffix} -->",
         f"## CBMC Results (ML-DSA-{cfg.param_set}{cfg.title_suffix})",
         "",
+        "`*` default; `(explore)` undeclared configuration.",
```
mkannwischer: The explore option can never happen in CI, can it? maybe drop it from the comment posted to the PR?

proofs/cbmc/Makefile.common
```diff
@@ -183,22 +167,124 @@ PROOFDIR ?= $(abspath .)
...
+# a known but undeclared configuration to be run experimentally.
+
+CBMC_DMS_ALL := LP64 ILP32
+
+CBMC_DM_LP64_GOTO_CC_FLAG := --64
+CBMC_DM_LP64_CBMC_FLAG := --LP64
+CBMC_DM_LP64_INCLUDE := $(CBMC_ROOT)/cbmc/include/LP64
+
+CBMC_DM_ILP32_GOTO_CC_FLAG := --32
+CBMC_DM_ILP32_CBMC_FLAG := --LP32
```
mkannwischer: Can you elaborate what is the difference between LP32 and ILP32 here?

```
$ cbmc --help 
 --16, --32, --64             equivalent to --LP32, --ILP32, --LP64
```
suggests that LP32 has 16 bit ints.

proofs/cbmc/run-cbmc-proofs.py
```diff
@@ -495,19 +615,87 @@ async def main():  # pylint: disable=too-many-locals
...
+                    default_configs.add(key)
+                if config in supported:
+                    configs_to_run.append((proof_dir, dm, solver, False))
+                elif args.explore:
+                    configs_to_run.append((proof_dir, dm, solver, True))
+                    exploratory_configs.add(key)
+                else:
+                    omitted_configs.add(key)
+
+    if not configs_to_run and not omitted_configs:
```
mkannwischer: Drop the omitted_configs configs here?

.github/actions/cbmc/report.py
```diff
@@ -78,68 +78,166 @@ def fetch_baseline(cfg):
...
+            if base_omitted
+            else "-"
+        )
+        return (
+            ProofResult(name, dm, solver_display, WARN, "?", prev, "inconclusive"),
+            True,
+        )
+
+    # Pair was intentionally not run.
+    if r.get("status") == "omitted":
```
mkannwischer: Can this even be reached? Line 227 filtered omitted results.

mkannwischer (changes requested):
Thanks @hanno-becker. This is a great step towards more portable proofs.

.github/workflows/cbmc.yml
hanno-becker: We can try without, I don't recall.

nix/cbmc/default.nix
hanno-becker: Indeed, thank you.

.github/actions/cbmc/report.py
hanno-becker: Yes indeed, good point.

proofs/cbmc/run-cbmc-proofs.py
hanno-becker: Yes, I think you are right.

.github/actions/cbmc/report.py
hanno-becker: Good spot.

proofs/cbmc/Makefile.common
hanno-becker: Very good spot. cc @rod-chapman

rod-chapman: I understand it that "LP32" means "long and pointers are 32 bit", but "ILP32" means "Int, Long and pointers are 32 bit." Note that the flags are (confusingly) different for goto-cc and cbmc.

mkannwischer (approved):
Thanks @hanno-becker. LGTM now!

PR #1124, opened 2026-05-21
x86_64: 32-byte align Keccak x4 AVX2 stack frame

hanno-becker:
For better performance, align stack to 32-byte in the AVX2 x4 Keccak backend implementation.

Updates HOL-Light proofs accordingly. Unfortunately, no existing tactic for the extension from the 'core' to the 'subroutine' proofs can handle the pattern we use, so we fall back to some ad-hoc tactics. As and when s2n-bignum adds support for stack alignment to its automation, this can hopefully be removed.

We also extend scripts/cfify to track the CFA register. A `mov rsp, %REG` re-anchors the CFA on REG, and subsequent modifications to the RSP do not require CFI directives. We handle this by conditioning the rules subq/addq->cfi_adjust_cfa_offset on the operand being the current CFA reg.


mkannwischer (approved):
Thanks @hanno-becker. LGTM.

PR #1125, opened 2026-05-22
nix: Bump z3 to 4.16.0 from nixpkgs-unstable

mkannwischer:
It is somewhat inconvienent for developers having to compile z3 from source. While this a one-off step, I would still like to avoid it if possible.

This commit sources z3 from nixpkgs-unstable rather than the local override so we pick up the binary cache, and aligns the pinned version with what nixpkgs 26.05 will ship.

Port of https://github.com/pq-code-package/mlkem-native/pull/1702.


jakemas (approved):
LGTM

mkannwischer:
Let's hold this back until we have #1126 has been addressed.

mkannwischer:
superseded by https://github.com/pq-code-package/mldsa-native/pull/1210

PR #1128, opened 2026-05-22
scripts/cfify: don't re-anchor CFA on scratch `movq %rsp, %REG`

hanno-becker:
The original `mov rsp, %REG` -> `.cfi_def_cfa_register %REG` rule fires on every `movq %rsp, %REG`, including scratch base-pointer copies (e.g. the `rep movsb` source setup in mlkem-native's rej_uniform_avx2_asm.S) with no intent to re-anchor. That misclassifies the scratch copy as an alignment anchor, drops the legitimate `.cfi_adjust_cfa_offset` on the matching `addq $N, %rsp`, and emits a spurious `.cfi_def_cfa_register`.

cfify now scans forward to the next ret and only re-anchors when a matching `movq %REG, %rsp` restore is found in the same function body.

Ported from mlkem-native commit 6ac47cb41. mldsa-native has no inputs that hit the bad path today (the only `movq %rsp, %REG` site, in keccak_f1600_x4_avx2_asm.S, has a matching restore), so the change is preventative; regenerated assembly is byte-identical.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -38,16 +38,6 @@ jobs:
           bench_extra_args: ""
           nix_shell: bench
           only_no_opt: false
-        - system: rpi5
-          name: Arm Cortex-A76 (Raspberry Pi 5) benchmarks
-          bench_pmu: PERF
-          archflags: "-mcpu=cortex-a76 -march=armv8.2-a"
-          cflags: "-flto -DMLD_FORCE_AARCH64"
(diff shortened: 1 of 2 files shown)


mkannwischer (approved):
LGTM.

The benchmarking CI was failing due to the A76 runner having a full disk. I can't access that machine anymore, so I have temporarily disabled the benchmarks in a commit stapled onto this PR.

PR #1131, opened 2026-05-26
Relax fqmul output bound to 5/4*Q

hanno-becker:
This commit relaxes the output bound for modular multiplication
from MLDSA_Q to 5/4 * MLDSA_Q. This bound is attained when using
a Barrett multiplication based on MUL+MULH+MLS rather than
MUL+SQRDMULH+MLS.

For now, mldsa-native does not contain any code which requires
the weaker output bound, but this commit gives us flexibility to
add it in the future if we need it. For example, a RV32-IM backend
would likely use a MUL+MULH+MLS based Barrett multiplication which
is subject to the 5/4 * MLDSA_Q bound.

### DIFF
--- a/dev/aarch64_clean/src/arith_native_aarch64.h
+++ b/dev/aarch64_clean/src/arith_native_aarch64.h
@@ -67,6 +67,7 @@ void mld_ntt_aarch64_asm(int32_t *r, const int32_t *zetas_l123456,
  * in proofs/hol_light/aarch64/proofs/ntt_aarch64_asm.ml */
 __contract__(
   requires(memory_no_alias(r, sizeof(int32_t) * MLDSA_N))
+  /* check-magic: 8380417 == MLDSA_Q */
   requires(array_abs_bound(r, 0, MLDSA_N, 8380417))
   requires(zetas_l123456 == mld_aarch64_ntt_zetas_layer123456)
   requires(zetas_l78 == mld_aarch64_ntt_zetas_layer78)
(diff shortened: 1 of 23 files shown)


dev/aarch64_clean/src/arith_native_aarch64.h
```diff
@@ -261,12 +262,14 @@ void mld_poly_pointwise_montgomery_aarch64_asm(int32_t *a, const int32_t *b)
 __contract__(
   requires(memory_no_alias(a, sizeof(int32_t) * MLDSA_N))
   requires(memory_no_alias(b, sizeof(int32_t) * MLDSA_N))
-  /* check-magic: off */
-  requires(array_abs_bound(a, 0, MLDSA_N, 75423753))
-  requires(array_abs_bound(b, 0, MLDSA_N, 75423753))
+  /* Input bound MLD_NTT_BOUND = 9 * MLD_FQMUL_BOUND, the post-condition of
+   * the forward NTT. Hardcoded here to keep this header free of poly.h.
+   * Must be kept in sync with the HOL-Light precondition. */
```
mkannwischer: I don't think we have to duplicate the "keep in sync" comment here.

mldsa/src/polyvec.c
mkannwischer, line 1: Comments in polyvec.h need updating.

dev/aarch64_clean/src/arith_native_aarch64.h
```diff
@@ -261,12 +262,14 @@ void mld_poly_pointwise_montgomery_aarch64_asm(int32_t *a, const int32_t *b)
 __contract__(
   requires(memory_no_alias(a, sizeof(int32_t) * MLDSA_N))
   requires(memory_no_alias(b, sizeof(int32_t) * MLDSA_N))
-  /* check-magic: off */
-  requires(array_abs_bound(a, 0, MLDSA_N, 75423753))
-  requires(array_abs_bound(b, 0, MLDSA_N, 75423753))
+  /* Input bound MLD_NTT_BOUND = 9 * MLD_FQMUL_BOUND, the post-condition of
+   * the forward NTT. Hardcoded here to keep this header free of poly.h.
```
mkannwischer: This is actually incorrect. The postcondition of the AArch64 NTT is < 75423753.

mkannwischer (changes requested):
Thanks @hanno-becker  - quite nice that this all works out.
A couple of nits.

The 36% CBMC performance regression is painful, but I suspect we may have to live with it?

dev/aarch64_clean/src/arith_native_aarch64.h
hanno-becker: The AArch64 NTT has stronger post-conditions, yes, but that's not visible in this spec. The C layer still feeds only the weaker pre-condition.

dev/aarch64_clean/src/arith_native_aarch64.h
hanno-becker: Ok

dev/aarch64_clean/src/arith_native_aarch64.h
hanno-becker: How about
```
/* Input bound MLD_NTT_BOUND = 9 * MLD_FQMUL_BOUND, the guaranteed 
 * bound of any forward NTT implementation. Hardcoded here to keep this header 
 * free of poly.h.
```

hanno-becker: Added the above. WDYT

mldsa/src/polyvec.c
hanno-becker: Right, done.

mkannwischer (approved):
LGTM. Thanks @hanno-becker!

PR #1133, opened 2026-05-27
CI: Add gcc16 tests 

mkannwischer:
- Based on #1132


hanno-becker (approved):
It doesn't work on MacOS, but apparently that's expected and accepted for now. Otherwise LGTM

PR #1139, opened 2026-05-31
CI: Check that all HOL-Light .S files are autogenerated

### LINKED ISSUE #1104: HOL-Light: Check in CI that all .S files are autogenerated
All .S files in `proofs/hol_light/...` should be autogenerated from the corresponding sources in `dev/`. 
We currently do check in CI that all the autogenerated .S files are up to date, but we do not check if all the files are actually autogenerated.
Consequently, it is possible to add a proof with a manually generated .S file. If later the corresponding source asm in `dev/` changes that would not propagate to `proofs/hol_light/...` and proofs would continue to pass even if they should be failing.

That gap should be closed.

mkannwischer:
The .S files under proofs/hol_light/{arch}/mldsa are generated from dev/ sources via the gen_hol_light_asm joblist in scripts/autogen. CI checks that the generated files are up to date, but not that every .S file is actually in the joblist. A proof could therefore be added with a manually written .S file; if the corresponding dev/ source later changed, that would not propagate and the proof would keep passing when it should fail.

Add check_hol_light_asm_autogenerated to close that gap: it flags any .S file under proofs/hol_light that is not produced by the joblist.

- Resolves https://github.com/pq-code-package/mldsa-native/issues/1104


scripts/autogen
```diff
@@ -2790,8 +2796,33 @@ def gen_hol_light_asm():
...
+        for infile, _indir, _cflags, arch in hol_light_asm_joblist()
+    }
+    stray = sorted(
+        os.path.join(d, fn)
+        for arch in ("aarch64", "x86_64")
+        for d in [os.path.join("proofs/hol_light", arch, "mldsa")]
+        if os.path.isdir(d)
+        for fn in os.listdir(d)
+        if fn.endswith(".S") and os.path.join(d, fn) not in expected
+    )
```
hanno-becker, line 2819: Bit clumsy, but nevermind

hanno-becker (approved):

PR #1140, opened 2026-06-01
ACVP: Allow skipping pre-hash modes unsupported by Python's hashlib

### LINKED ISSUE #719: Python 3.7 on Amazon Linux 2 lacks `sha512_224` in hashlib
- During #665 , we found that the following Amazon Linux 2 containers failed ACVP tests due to missing sha512_224 support:
  - amazonlinux-2-aarch:base
  - amazonlinux-2-aarch:gcc-7x
  - amazonlinux-2-aarch:clang-7x

Error is reported during `make quickcheck`:
```
  File "/usr/lib64/python3.7/hashlib.py", line 150, in __hash_new
    return _hashlib.new(name, data)
ValueError: unsupported hash type

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./test/acvp_client.py", line 469, in <module>
    test(args.prompt, args.expected, args.output, args.version)
  File "./test/acvp_client.py", line 437, in test
    runTest(data, output)
  File "./test/acvp_client.py", line 417, in runTest
    runTestSingle(promptName, prompt, expectedResultName, expectedResult, output)
  File "./test/acvp_client.py", line 379, in runTestSingle
    result = run_sigGen_test(tg, tc)
  File "./test/acvp_client.py", line 206, in run_sigGen_test
    ph = compute_hash(tc["message"], tc["hashAlg"])
  File "./test/acvp_client.py", line 161, in compute_hash
    return hashlib.new("sha512_224", msg_bytes).hexdigest()
  File "/usr/lib64/python3.7/hashlib.py", line 156, in __hash_new
    return __get_builtin_constructor(name)(data)
  File "/usr/lib64/python3.7/hashlib.py", line 113, in __get_builtin_constructor
    raise ValueError('unsupported hash type ' + name)
ValueError: unsupported hash type sha512_224
```
This failure is unrelated to test correctness; it is solely due to missing hash support required for computing the verification value. 
- This issue is created to track and record the problem for a future fix.

mkannwischer:
Some Python/OpenSSL builds lack the truncated SHA-512 variants
(SHA2-512/224, SHA2-512/256) needed for pre-hash ACVP test cases. Add
--skip-unsupported to drop the affected cases from both prompt and
expected results; without it the client errors with guidance on the
flag. Expose it through make via ACVP_OPTIONS.

- Resolves https://github.com/pq-code-package/mldsa-native/issues/719

### DIFF
--- a/test/acvp/acvp_client.py
+++ b/test/acvp/acvp_client.py
@@ -202,18 +202,26 @@ def compute_hash(msg, alg):
     return h.hexdigest() if xof_len is None else h.hexdigest(xof_len)
 
 
+def _available_hash_algs():
+    """Determine hashes supported by the underlying hashlib implementation"""
+    available = set()
+    for alg, (name, _) in HASH_ALG_TO_HASHLIB.items():
+        try:
+            hashlib.new(name)
+            available.add(alg)
+        except (ValueError, TypeError):
+            pass
+    return available
+
+
+AVAILABLE_HASH_ALGS = _available_hash_algs()
+
+
 def hashlib_can_compute(hashAlg):
     # SHAKE-256 pre-hash is computed inside the ACVP binary, not via hashlib.
     if hashAlg in (None, "none", "SHAKE-256"):
         return True
-    name, _ = HASH_ALG_TO_HASHLIB.get(hashAlg, (None, None))
-    if name is None:
-        return False
-    try:
-        hashlib.new(name)
-        return True
-    except (ValueError, TypeError):
-        return False
+    return hashAlg in AVAILABLE_HASH_ALGS
 
 
 def unwrap_acvts(data):
--- a/test/acvp/acvp_client.py
+++ b/test/acvp/acvp_client.py
@@ -177,35 +177,80 @@ def run_keyGen_test(tg, tc):
     return results
 
 
-def compute_hash(msg, alg):
-    msg_bytes = bytes.fromhex(msg)
+# ACVP hashAlg -> (hashlib name, XOF output length in bytes or None).
+HASH_ALG_TO_HASHLIB = {
+    "SHA2-224": ("sha224", None),
+    "SHA2-256": ("sha256", None),
+    "SHA2-384": ("sha384", None),
+    "SHA2-512": ("sha512", None),
+    "SHA2-512/224": ("sha512_224", None),
+    "SHA2-512/256": ("sha512_256", None),
+    "SHA3-224": ("sha3_224", None),
+    "SHA3-256": ("sha3_256", None),
+    "SHA3-384": ("sha3_384", None),
+    "SHA3-512": ("sha3_512", None),
+    "SHAKE-128": ("shake_128", 32),
+    "SHAKE-256": ("shake_256", 64),
+}
 
-    if alg == "SHA2-224":
-        return hashlib.sha224(msg_bytes).hexdigest()
-    elif alg == "SHA2-256":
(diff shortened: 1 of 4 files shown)


jakemas (approved):
Looks good to me! Shall we squash the PR into a single commit?

hanno-becker (changes requested):
I think this could be a bit simpler and more generic.

Can we have a single function which takes a filter on `tc,tg` and trims down `promptData` and `expectedData` in one go, based on the filter?

The filter should be configurable and either pass or return a string explaining why the test is dropped. That string is then simply printed out by the overall filter.

This makes it easier to add other sources of unsupported tests in the future (or to deliberately trim down the ACVP test set for other reasons).

WDYT

mkannwischer:
> I think this could be a bit simpler and more generic.
> 
> Can we have a single function which takes a filter on `tc,tg` and trims down `promptData` and `expectedData` in one go, based on the filter?
> 
> The filter should be configurable and either pass or return a string explaining why the test is dropped. That string is then simply printed out by the overall filter.
> 
> This makes it easier to add other sources of unsupported tests in the future (or to deliberately trim down the ACVP test set for other reasons).
> 
> WDYT

I have pushed a new version. I don't think this qualifies as "simpler", but it should be easier to extend.

test/acvp/acvp_client.py
```diff
@@ -177,35 +177,80 @@ def run_keyGen_test(tg, tc):
...
+                reason = should_drop(tg, tc)
+                if reason is not None:
+                    drop[tg["tgId"], tc["tcId"]] = reason
+        for data in (promptData, expectedData):
+            if data is None:
+                continue
+            for tg in unwrap_acvts(data).get("testGroups", []):
+                tg["tests"] = [
+                    tc for tc in tg["tests"] if (tg["tgId"], tc["tcId"]) not in drop
+                ]
```
hanno-becker, line 259: Why do we need those loops separately? Why can we not just filter directly in the second loop?

test/acvp/acvp_client.py
```diff
@@ -177,35 +177,80 @@ def run_keyGen_test(tg, tc):
...
-        return hashlib.shake_256(msg_bytes).hexdigest(64)
-    else:
+    if alg not in HASH_ALG_TO_HASHLIB:
         raise ValueError(f"Unsupported hash algorithm: {alg}")
+    name, xof_len = HASH_ALG_TO_HASHLIB[alg]
+    h = hashlib.new(name, bytes.fromhex(msg))
+    return h.hexdigest() if xof_len is None else h.hexdigest(xof_len)
+
+
+def hashlib_can_compute(hashAlg):
```
hanno-becker, line 220: Nit: Why don't we filter this just once in the beginning rather than with every test?

test/acvp/acvp_client.py
hanno-becker: I thought about this -- probably longer than merited -- and came to the conclusion that the current code is the cleanest.

test/acvp/acvp_client.py
hanno-becker: I added this in 0c7a141ce6e1286ed84b9d180107d004963b0834. @mkannwischer please have a quick look.

hanno-becker (approved):

test/acvp/acvp_client.py
mkannwischer: LGTM. Thanks!

PR #1141, opened 2026-06-02
CI: Move RISC-V runner jobs out of base into a dedicated workflow

mkannwischer:
- Ported from https://github.com/pq-code-package/mlkem-native/pull/1708


hanno-becker (changes requested):
Same as in mlkem-native: Seeing that there aren't that many RISC-V runners available, we should not unnecessarily exercise them; let's only run the RV workflow once the base workflow passes.

hanno-becker (approved):

Issue #536, opened 2025-10-08
Split up poly_kl.c to reduce code duplication in multilevel build


mkannwischer:
poly_kl.c contains poly functions that are parameter set specific.
Currently in a multilevel build each function in that file would be built once for each parameter set.

This can be improved, e.g., 
 - ML-DSA-65 and ML-DSA-87 share the same mld_poly_decompose
 - ML-DSA-44 and ML-DSA-87 share the same mld_rej_eta
 - etc. 

Once implemented, this comment should be adjusted: 

https://github.com/pq-code-package/mldsa-native/blob/db475c1ac5882c12eb1c1a5906458b7f2614b68c/mldsa/src/config.h#L57-L78

PR #1142, opened 2026-06-02
Deduplicate polyw1_pack in multilevel builds

mkannwischer:
polyw1_pack depends only on GAMMA2, so ML-DSA-65 and -87 use identical code. Move it into the shared poly.c as mld_polyw1_pack_{88,32}, compiled once, with a thin dispatcher in poly_kl.h, so a multilevel build shares a single copy instead of emitting one per parameter set. Split the CBMC proof into per-variant proofs plus a dispatcher proof.

- Towards #536


mkannwischer:
@hanno-becker, could you take a look if you agree with the general approach taken here? Then I'll apply the same to the other functions that are currently duplicated in poly_kl.c

mldsa/src/poly.h
```diff
@@ -378,4 +378,49 @@ __contract__(
   ensures((return_value == 0) == array_abs_bound(a->coeffs, 0, MLDSA_N, B))
 );
 
+#if !defined(MLD_CONFIG_NO_SIGN_API) || !defined(MLD_CONFIG_NO_VERIFY_API)
+#if defined(MLD_CONFIG_MULTILEVEL_WITH_SHARED) || MLD_CONFIG_PARAMETER_SET == 44
+#define mld_polyw1_pack_88 MLD_NAMESPACE(polyw1_pack_88)
+/**
+ * Bit-pack polynomial w1 for GAMMA2 = (MLDSA_Q-1)/88 (coefficients in [0, 43]).
+ * Input coefficients are assumed to be standard representatives.
```
hanno-becker: It is unclear what this means

mldsa/src/poly.h
```diff
@@ -378,4 +378,49 @@ __contract__(
...
+);
+#endif /* MLD_CONFIG_MULTILEVEL_WITH_SHARED || MLD_CONFIG_PARAMETER_SET == 44 \
+        */
+
+#if defined(MLD_CONFIG_MULTILEVEL_WITH_SHARED) || \
+    (MLD_CONFIG_PARAMETER_SET == 65 || MLD_CONFIG_PARAMETER_SET == 87)
+#define mld_polyw1_pack_32 MLD_NAMESPACE(polyw1_pack_32)
+/**
+ * Bit-pack polynomial w1 for GAMMA2 = (MLDSA_Q-1)/32 (coefficients in [0, 15]).
+ * Input coefficients are assumed to be standard representatives.
```
hanno-becker: Same here

hanno-becker (approved):
LGTM, but we might use the opportunity to improve some pre-existing yet unclear documentation; see comment.

hanno-becker:
@mkannwischer `poly_ntt_c` exhibits a significant performance regression which we may want to fix before merging.

mkannwischer:
> @mkannwischer `poly_ntt_c` exhibits a significant performance regression which we may want to fix before merging.

I've spent quite some time trying to fix this, but I couldn't find flags that restore the old performance with z3.
To summarize: This is platform-dependent code-location-dependent CBMC flakiness. It's not reproducible on x86_64-linux or MacOS. Tiny changes to poly.c make it go away. On main the proof takes seconds. With tiny unrelated changes to poly.c it goes to TIMEOUT in CI. I have reproduced the behavior on the same machine as CI is using (c7g.4xlarge). Our usual z3 flags do not help. That is why I switched to bitwuzla which at least seems to produce stable results in 2 minutes. 

As discussed offline, we'll merge the switch to bitwuzla, then reproduce the problem in the main branch and open an issue to investigate further.

mldsa/src/poly.h
mkannwischer: Fixed.

mldsa/src/poly.h
mkannwischer: Fixed.

PR #1146, opened 2026-06-03
mldsa: annotate functions with FIPS 204 algorithm references

### LINKED ISSUE #696: Add ` Specification:` annotations linking functions to FIPS204
mlkem-native has ` Specification:` that links individual functions to Algorithms in the specification, e.g.,

https://github.com/pq-code-package/mlkem-native/blob/d5d13e1c9e42b15b21d1f6a3b27831573d75dc83/mlkem/src/kem.h#L180-L200

We should add such annotations to the mldsa-native code base, too.

mkannwischer:
Add `@spec{Implements @[FIPS204, Algorithm N, Name]}` annotations linking functions to their FIPS 204 algorithms.

Also normalize the existing sign.h `@spec` references from `@[FIPS204 Algorithm N (Name)]` to the comma style used throughout.

Covers Algorithms 1-8, 16-36, 39-42, 44, 45, 48, 49. The remaining algorithms have no dedicated function: 9-15 (IntegerToBits/.../CoeffFromHalfByte), 37/38 (HighBits/LowBits, subsumed in Decompose), 43 (BitRev8, precomputed zetas), and 46/47 (AddVectorNTT/ScalarVectorNTT, applied inline in sign.c as loops over AddNTT/MultiplyNTT).

 - Resolves https://github.com/pq-code-package/mldsa-native/issues/696


mldsa/src/packing.h
```diff
@@ -80,6 +99,10 @@ __contract__(
 /**
  * Compute hints from (w0, w1) and pack them into the hint section of sig.
  *
+ * @spec{Implements @[FIPS204, Algorithm 20, HintBitPack], computing the hints
```
hanno-becker: The relation to `MakeHint` isn't readily clear and should be documented while at it. Generally, it is difficult to reconcile `attempt_signature_generation` with the flow of `ML-DSA.Sign_internal` because the variables names in the spec and in the source don't line up.

If possible, I'd prefer to improve the documentation of `attempt_signature_generation` while at it, so it gets easier to see why the way we compute signature and hints is indeed what's written in the spec.

hanno-becker: I'll take care of that.

jakemas:
Did a scan and looks good to me!

jakemas (approved):

mldsa/src/poly.h
```diff
@@ -169,7 +170,9 @@ __contract__(
  * Input coefficients need to be less than MLDSA_Q in absolute value and
  * output coefficients are bounded by MLD_INTT_BOUND.
  *
- * @spec{Implements @[FIPS204, Algorithm 42, NTT^{-1}].}
+ * @spec{Implements @[FIPS204, Algorithm 42, NTT^{-1}], up to the extra factor
+ * of 2^{32}: the output is in Montgomery form, i.e. NTT^{-1}(a) * 2^{32} mod q.
+ * This is the Montgomery representation sanctioned by @[FIPS204, Appendix A].}
```
mkannwischer: I don't think this is true, is it?
Isn't it that in the pointwise multiplication we Montgomery reduce even though the inputs are not in Montgomery domain, so in the intt we have to correct the R^-1 factor from there?

mldsa/src/poly.h
```diff
@@ -190,7 +193,9 @@ __contract__(
  * multiplication of resulting polynomial by 2^{-32}. Destructive in the first
  * argument.
  *
- * @spec{Implements @[FIPS204, Algorithm 45, MultiplyNTT].}
+ * @spec{Implements @[FIPS204, Algorithm 45, MultiplyNTT] on inputs in
+ * Montgomery form: the extra factor of 2^{-32} produces the Montgomery-form
+ * product a * b * 2^{-32} mod q, as sanctioned by @[FIPS204, Appendix A].}
```
mkannwischer: Inputs aren't in Montgomery form here, are they?

mldsa/src/poly.h
hanno-becker: You are correct. The _input_ is twisted by `R^{-1}`, the _output_ isn't anymore.

mldsa/src/poly.h
hanno-becker: Inputs are in normal domain. Outputs are twisted by `R^{-1}` (which is not Montgomery domain).

hanno-becker: Fixed

mldsa/src/poly.h
hanno-becker: Fixed

mldsa/src/poly.h
```diff
@@ -170,9 +190,12 @@ __contract__(
     defined(MLD_CONFIG_REDUCE_RAM) || defined(MLD_UNIT_TEST)
 #define mld_poly_pointwise_montgomery MLD_NAMESPACE(poly_pointwise_montgomery)
 /**
- * Pointwise multiplication of polynomials in NTT domain representation and
- * multiplication of resulting polynomial by 2^{-32}. Destructive in the first
- * argument.
+ * Pointwise multiplication of polynomials. Destructive in the first argument.
+ *
+ * @spec{Implements @[FIPS204, Algorithm 45, MultiplyNTT], up to scaling: The
+ * input is in normal domain, the ouptut is scaled by 2^{-32} as a result of
```
mkannwischer:
```suggestion
 * input is in normal domain, the output is scaled by 2^{-32} as a result of
```

hanno-becker: Fixed

hanno-becker (approved):

PR #1148, opened 2026-06-04
dev: Document ABI and callee-saved register handling at function entry

hanno-becker:
Add a uniform comment block at the entry of every aarch64 and x86_64 assembly function under dev/, naming the calling convention (AAPCS or System V AMD64), listing which callee-saved registers (if any) are saved/restored, and giving the function signature with per-register argument descriptions.

### DIFF
--- a/dev/aarch64_clean/src/intt_aarch64_asm.S
+++ b/dev/aarch64_clean/src/intt_aarch64_asm.S
@@ -228,6 +228,18 @@
         .global MLD_ASM_NAMESPACE(intt_aarch64_asm)
         .balign 4
 MLD_ASM_FN_SYMBOL(intt_aarch64_asm)
+        // ABI: AArch64 AAPCS. Callee-saved vector registers v8-v15 are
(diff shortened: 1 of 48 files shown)


dev/aarch64_clean/src/intt_aarch64_asm.S
```diff
@@ -228,6 +228,18 @@
...
+        // stack via push_stack/pop_stack. No callee-saved GPRs are
+        // touched.
+        //
+        // Function signature:
+        //   void mld_intt_aarch64_asm(int32_t *in,
+        //                             const int32_t *r78,
+        //                             const int32_t *r123456)
+        //   x0: pointer to in/out polynomial
+        //   x1: pointer to layer 7-8 twiddles
+        //   x2: pointer to layer 1-6 twiddles
```
mkannwischer, line 242: Why are we not following what we are doing in mlkem-native?

hanno-becker:
Let's defer this til after the merge and port of the ABI checker from mlkem-native.

PR #1136, opened 2026-05-29
HOL-Light: Consolidate AArch64 poly_use_hint_32 _CORRECT theorems

jakemas:
## Summary

Consolidates the `poly_use_hint_32` AArch64 HOL-Light proofs to use the same `_CORRECT` / `_SUBROUTINE_CORRECT` shape as every other proof in the codebase, addressing #1085.

The `_CORRECT_CODE` / `_CORRECT_BOUND_CODE` / `_SUBROUTINE_CORRECT_CODE` intermediates are removed; the public `_CORRECT` is now stated directly in FIPS 204 terms (`mldsa_use_hint_32`) with the `< 16` bound as a corollary, following the pattern outlined in https://github.com/pq-code-package/mldsa-native/pull/1037#issuecomment-4363057304:

- The FIPS 204 / code-aligned equivalence section (`MLDSA_USE_HINT_32_EQUIV`) moves above the correctness proof.
- `_CORRECT` uses `ENSURES_STRENGTHEN_POST` to bridge the FIPS-aligned postcondition to the code-aligned form before symbolic execution; the strengthening branch closes via `IMP_REWRITE_TAC[MLDSA_USE_HINT_32_EQUIV]` and a short bound discharge.
- `_SUBROUTINE_CORRECT` derives directly from `_CORRECT` via `ARM_ADD_RETURN_NOSTACK_TAC`.

The `_88` variant will follow in a separate PR.

Validated end-to-end against the equivalent file in s2n-bignum (`arm/proofs/mldsa_poly_use_hint_32.ml`), where the same refactor proves in 25:10 wall time.

Closes part of #1085. If we are happy with `_32` I can make the same changes to `_88`

Issue #1085, opened 2026-05-02
HOL-Light: Consolidate AArch64 poly_use_hint proofs


mkannwischer:
The poly_use_hint proofs from #1037 currently has the following specs:

```
POLY_USE_HINT_32_AARCH64_ASM_CORRECT_CODE
POLY_USE_HINT_32_AARCH64_ASM_SUBROUTINE_CORRECT_CODE
POLY_USE_HINT_32_AARCH64_ASM_SUBROUTINE_CORRECT
```

No other proof in mlkem-native and mldsa-native has this `_CORRECT` and `_CORRECT_CODE` split. It would be nice if we can consolidate this to follow the other proofs.

@hanno-becker outlined how this could be done in https://github.com/pq-code-package/mldsa-native/pull/1037#issuecomment-4363057304.

jakemas:
Do the linked PRs meet the requirements of this issue?

Issue #995, opened 2026-03-16
HOL-Light: Unify "pre-conditions as post-conditions" style


hanno-becker:
Pre-conditions can often be transformed into premises of the post-condition. While largely equivalent, this can lead to slightly stronger theorems: For example, if an NTT zeta table is fixed in the pre-condition, we have no guarantee of memory-safety if corrupted constants are passed. If instead, the values of the zeta pointer are constrained only in the post-condition, then the safety spec entails that even with a corrupted constant table, safety (albeit not correctness, of course) is retained.

**Task:**
* Audit the mlkem-native and mldsa-native for the use of this specification pattern
* Agree on a set of preconditions for which there is added value in expressing them as premises of the post-condition
* Adjust all specifications and proofs as needed

hanno-becker:
Mirroring  https://github.com/pq-code-package/mlkem-native/issues/1627:

I propose to pivot this to the exact opposite. The safety proofs already ensure that programs execute and terminate normally, without specific constraints on the contents of pointers. The premise-to-postcondition pattern in the correctness spec would only be useful in the absence of separate safety proofs. It also makes the specs significantly harder to read -- in a world where proofs are increasingly AI generated, communicating what has been proved and making it as accessible as possible is crucial.

So, I propose to instead remove all uses of the premise-to-postcondition pattern and go back to a plain precondition.

PR #1152, opened 2026-06-04
HOL-Light: Consolidate AArch64 poly_use_hint_88 _CORRECT theorems

jakemas:
Mirrors the poly_use_hint_32 consolidation in #1136. Closes part of #1085.

### DIFF
--- a/proofs/hol_light/aarch64/proofs/poly_use_hint_88_aarch64_asm.ml
+++ b/proofs/hol_light/aarch64/proofs/poly_use_hint_88_aarch64_asm.ml
@@ -291,8 +291,6 @@ let BARRETT_EQUIV_88 = prove(
     | [] -> failwith "empty" in
   cascade 0 intervals);;
 
-
-
 let WORD_SUB_SIGN_88 = BITBLAST_RULE
   `!a:int32 b:int32. val b <= 8189952 /\ val a <= 8285184 ==>
    ((bit 31 (word_sub a b) \/ word_sub a b = word 0) <=> val a <= val b)`;;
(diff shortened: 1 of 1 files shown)


proofs/hol_light/aarch64/proofs/poly_use_hint_88_aarch64_asm.ml
```diff
@@ -930,50 +768,165 @@ let MLDSA_USE_HINT_88_EQUIV = prove(
...
+          (\s. aligned_bytes_loaded s (word pc) poly_use_hint_88_aarch64_asm_mc /\
+               read PC s = word pc /\
+               C_ARGUMENTS [a; h] s /\
+               (!i. i < 256 ==>
+                 read(memory :> bytes32(word_add a (word(4 * i)))) s = x i) /\
+               (!i. i < 256 ==>
+                 read(memory :> bytes32(word_add h (word(4 * i)))) s = y i))
+          (\s. read PC s = word(pc + LENGTH poly_use_hint_88_aarch64_asm_mc - 4) /\
+               ((!i. i < 256 ==> val(x i:int32) < 8380417) /\
+                (!i. i < 256 ==> val(y i:int32) <= 1)
```
hanno-becker, line 793: When we discussed this internally with @jargh recently, we came to the consensus that in the context of a separate safety specification, there is no added value to moving those preconditions into premises of the post-condition (without a separate safety spec, however, there would be).

So I don't think this move should be made. However, we have already merged the corresponding change for use_hint_32, so let's uniformly change it in another round of rework at a later stage.

hanno-becker: cf https://github.com/pq-code-package/mldsa-native/issues/995 (note that the issue started out proposing the opposite)

hanno-becker (approved):
We should not move preconditions into premises of postconditions anymore, cf https://github.com/pq-code-package/mldsa-native/issues/995, but we have already done the same in use_hint_32, so let's keep things uniform and switch jointly at a later point.

Other than that, the proof cleanup looks great, thank you @jakemas for taking the time to do this!

@mkannwischer Can please also take a quick look?

mkannwischer (approved):

PR #1158, opened 2026-06-07
Tests: Add AVR baremetal tests

### LINKED ISSUE #1157: Consider AVR baremetal tests
mlkem-native runs functional/KAT/ACVP tests on a baremetal AVR target (test/baremetal/platform/avr/) using avr-gcc and a patched simavr (ATmega128RFR2 bumped to 32 KB RAM), primarily to test correctness on a 16-bit platform.

With the recent memory optimizations (via `REDUCE_RAM`), we should be able to at least run ML-DSA-44 tests on that platform. We may even be able to extend the RAM more and run more parameter sets.

mkannwischer:
- Resolves #1157

### DIFF
--- a/.github/workflows/baremetal.yml
+++ b/.github/workflows/baremetal.yml
@@ -24,6 +24,8 @@ jobs:
            acvp: true
            wycheproof: false
            alloc: true
+           unit: true
+           rng_fail: true
            bench: true
            opt: all
          - runner: ubuntu-latest
@@ -35,7 +37,23 @@ jobs:
            acvp: true
            wycheproof: false
            alloc: true
+           unit: true
+           rng_fail: true
            bench: true
+           opt: no_opt
+         - runner: ubuntu-latest
+           name: 'AVR ATmega128RFR2 (modified for 63.5K RAM)'
+           makefile: test/baremetal/platform/avr/platform.mk
+           nix-shell: cross-avr
+           func: true
+           acvp: true
+           # Other tests are disabled to keep the CI runtime manageable.
+           kat: false
+           wycheproof: false
+           alloc: false
+           unit: false
+           rng_fail: false
+           bench: false
            opt: no_opt
     runs-on: ${{ matrix.target.runner }}
     steps:
@@ -55,7 +73,8 @@ jobs:
           examples: false
           stack: false
           alloc: ${{ matrix.target.alloc }}
-          rng_fail: true
+          unit: ${{ matrix.target.unit }}
+          rng_fail: ${{ matrix.target.rng_fail }}
       - name: Baremetal bench
         if: ${{ matrix.target.bench }}
         uses: ./.github/actions/bench
--- a/mldsa/src/poly.c
+++ b/mldsa/src/poly.c
@@ -217,7 +217,7 @@ static MLD_INLINE void mld_ntt_butterfly_block(int32_t r[MLDSA_N],
                                                const int32_t zeta,
                                                const unsigned start,
                                                const unsigned len,
-                                               const unsigned bound)
+                                               const uint32_t bound)
 __contract__(
   requires(start < MLDSA_N)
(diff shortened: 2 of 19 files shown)


mkannwischer:
The ACVP tests for verification require the return value of the main function to properly test invalid signatures. Currently the AVR harness does not propagate the return codes. That needs to be fixed first.

mkannwischer:
> The ACVP tests for verification require the return value of the main function to properly test invalid signatures. Currently the AVR harness does not propagate the return codes. That needs to be fixed first.

That is now added in. Also the UBSan tests were ported from mlkem-native

mldsa/src/poly.c
```diff
@@ -285,7 +285,8 @@ __contract__(
     decreases(MLDSA_N - start))
   {
     int32_t zeta = mld_zetas[k++];
-    mld_ntt_butterfly_block(r, zeta, start, len, layer * MLD_FQMUL_BOUND);
+    mld_ntt_butterfly_block(r, zeta, start, len,
+                            layer * (uint32_t)MLD_FQMUL_BOUND);
```
hanno-becker: Hm, I'm surprised this is even necessary and the type is not 32-bit from the definition of the macros themselves.

But if it isn't, I'd rather force the type explicitly in those macro definitions, not here?

mkannwischer: this was a leftover from an earlier attempt. Removed now.

.github/workflows/baremetal.yml
```diff
@@ -35,8 +37,23 @@ jobs:
...
            alloc: true
+           unit: true
+           rng_fail: true
            bench: true
            opt: no_opt
+         - runner: ubuntu-latest
+           name: 'AVR ATmega128RFR2 (modified for 63.5K RAM)'
+           makefile: test/baremetal/platform/avr/platform.mk
+           nix-shell: cross-avr
+           func: false
```
hanno-becker: The commit message says that functional tests pass, yet they are disabled here.

mkannwischer: They passed in CI before if I remember correctly, but runtime for the test overall was too long. We could add a func test with a reduced number of tests if you think it is useful. My priority is ACVP.

test/baremetal/platform/avr/exec_wrapper.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+def create_eeprom_hex(args, output_file):
+    """
+    Create EEPROM hex file from command line arguments.
+
+    EEPROM layout: 2 bytes block base address (little-endian), followed by
+    the argc/argv block to be copied to that address:
+    argc (2 bytes) + argv array (len(args) * 2 bytes) + packed strings.
+    """
```
hanno-becker, line 41: Will we be able to port this approach to mlkem-native? It would be nice to use the same scripts in both repositories.

.github/workflows/baremetal.yml
hanno-becker: Similar for `rng_fail`

mkannwischer: Same case - runtime is the issue.

hanno-becker: Ah, understood. Then we could just leave a comment for the time being.

hanno-becker (approved):
Thank you very much @mkannwischer, this is looking great.

It seems that the in-place hex-decoding and the flexible placement of arguments would benefit mlkem-native as well. Could you consider a port?

mkannwischer:
> Thank you very much @mkannwischer, this is looking great.
> 
> It seems that the in-place hex-decoding and the flexible placement of arguments would benefit mlkem-native as well. Could you consider a port?

Thanks!
I've opened https://github.com/pq-code-package/mlkem-native/issues/1755 for porting it to mlkem-native.

PR #1169, opened 2026-06-11
Add clang-tidy to lint with pedantic config

mkannwischer:
Run clang-tidy as part of scripts/lint. clang-tidy needs a C compiler to resolve system headers, so the lint CI job moves from the compiler-less 'linter' shell to 'ci' (linters + native toolchain); the now-unused 'linter' devShell is removed.

Pre-suppress style checks hostile to this codebase; fix or suppress the substantive findings, including breaking the common.h <-> ct.h include cycle.

Ports pq-code-package/mlkem-native#1737.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,48 @@
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+#
+# clang-tidy configuration for mldsa-native
+#
+# We enable whole C-relevant families and disable individual checks with
+# rationale provided inline below.
+Checks:
+  - bugprone-*
+  - cert-*
+  - clang-analyzer-*
+  - concurrency-*
+  - misc-*
+  - performance-*
+  - portability-*
+  - readability-*
+  # Spec/maths use 1-letter names (a, b, r, f, t, k).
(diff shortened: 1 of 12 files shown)


mldsa/src/rounding.h
```diff
@@ -223,19 +223,13 @@ __contract__(
...
 #else  /* MLD_CONFIG_PARAMETER_SET == 44 */
   if (a0 > 0)
   {
     return (a1 + 1) & 15;
   }
-  else
-  {
-    return (a1 - 1) & 15;
-  }
+  return (a1 - 1) & 15;
```
hanno-becker: I'm not sure I agree with this being an improvement. Can we consider disabling this check?

hanno-becker (commented):
Thank you, @mkannwischer! For some changes, I'm not sure if they really improve the code -- see comment. WDYT

mldsa/src/rounding.h
mkannwischer: Agree. I was strictly following what you had in mlkem-native, but that rule doesn't help us.

hanno-becker (approved):
LGTM

PR #1177, opened 2026-06-14
Simplify decompose C reference to a single high multiplication

### LINKED ISSUE #652: Consider simplifying `decompose` C implementation
Currently decompose implements `round(a / 2*GAMMA2)` using a two steps: For ML-DSA-44 (round(a / 190464) this looks as follows (but similar for the other parameter sets)
```c
*a1 = (a + 127) >> 7;
  *a1 = (*a1 * 11275 + (1 << 23)) >> 24;
```

In the AArch64 backend we show how this can be done using a single high multiplication:
```asm
        // Compute a1 = round-(a / 190464) ≈ round(a * 1477838209 /
        // 2^48), where round-() denotes "round half down". This is
        // exact for 0 <= a < Q. Note that half is rounded down since
        // 1477838209 / 2^48 ≲ 1 / 190464.
        sqdmulh \a1\().4s, \a\().4s, barrett_const.4s
        srshr \a1\().4s, \a1\().4s, #17
```

We should consider revisiting decompose to also do this in a single step.

mkannwischer:
Replace the two-step Barrett division (ceil(a/128) then Barrett-divide
by 2*GAMMA2/128) with a direct high multiplication by floor(2^N /
2*GAMMA2), mirroring the AArch64 backend. For ML-DSA-44 this is
`(a * 1477838209 + 2^47) >> 48`; for ML-DSA-65/87 it is
`(a * 1074791425 + 2^48) >> 49`. Both constants strictly
under-approximate 1/(2*GAMMA2), so half-points round down, matching
the original round-half-down semantics, and the result is exact for
all 0 <= a < Q.

Update the Isabelle attribution in compress/ML-DSA_Compress.thy and
neon_ntt/Barrett_Division_Even.thy.

- Resolves #652

### DIFF
--- a/mldsa/src/rounding.h
+++ b/mldsa/src/rounding.h
@@ -91,78 +91,34 @@ __contract__(
 )
 {
   /*
-   * The goal is to compute f1 = round-(f / (2*GAMMA2)), which can be computed
-   * alternatively as round-(f / (128B)) = round-(ceil(f / 128) / B) where
-   * B = 2*GAMMA2 / 128. Here round-() denotes "round half down".
-   *
-   * The equality round-(f / (128B)) = round-(ceil(f / 128) / B) can deduced
-   * as follows. Since changing f to align-up(f, 128) can move f onto but not
-   * across a rounding boundary for division by 128*B (note that we need B to be
-   * even for this to work), and round- rounds down on the boundary, we have
-   *
-   *   round-(f / (128B)) = round-(align-up(f, 128) / (128B))
-   *                      = round-((align-up(f, 128) / 128) / B)
-   *                      = round-(ceil(f / 128) / B).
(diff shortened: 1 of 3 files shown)


hanno-becker (approved):
This looks right to me.

hanno-becker:
Functionally, I am fine with this, but the impact on 32-bit platforms is unclear. We should do some measurements on Armv8-M and RV32 first before proceeding with this PR.

mkannwischer:
> Functionally, I am fine with this, but the impact on 32-bit platforms is unclear. We should do some measurements on Armv8-M and RV32 first before proceeding with this PR.

Agreed, I was still planning to investigate more before marking this ready as review.  I won't find time to look into it before next month - feel free to take over.

PR #1187, opened 2026-06-16
x86_64: Replace rej_uniform_eta2/eta4 intrinsics with hand-written assembly

jakemas:
## Summary

Draft — opened while the HOL-Light proofs are still in progress.

Replaces the AVX2 intrinsics implementations of `rej_uniform_eta2` and `rej_uniform_eta4` with hand-written x86_64 assembly, following the same approach as #1014:

- Table passed as a parameter (consistent with the aarch64 approach), avoiding external symbol references for simpasm compatibility.
- All constants constructed from immediates (no `.rodata`), enabling future HOL-Light formal verification.
- `__contract__` annotations on the asm entry points (CBMC), to be kept in sync with the HOL-Light specs.
- `scripts/autogen` and the x86_64 HOL-Light `Makefile` register the eta2/eta4 bytecode dump targets.
- Adds a `poly_uniform_eta_4x` component benchmark.

## Scope of this draft

This draft contains **assembly + build/bytecode infrastructure only**. It intentionally **excludes the HOL-Light `.ml` proofs**, which are still being developed. The proofs will be added before this is marked ready for review.

## Notes / TODO before ready-for-review

- [ ] `meta.h` currently wires only **eta4** to the new asm; **eta2** asm is in place but `mld_rej_uniform_eta2_native` still calls the intrinsic — wire it once verified.
- [ ] Add HOL-Light CORRECT (and MEMSAFE) proofs for eta2/eta4.
- [ ] Performance numbers.

### DIFF
--- a/dev/x86_64/src/arith_native_x86_64.h
+++ b/dev/x86_64/src/arith_native_x86_64.h
@@ -85,10 +85,42 @@ MLD_MUST_CHECK_RETURN_VALUE
 unsigned mld_rej_uniform_eta2_avx2(
     int32_t *r, const uint8_t buf[MLD_AVX2_REJ_UNIFORM_ETA2_BUFLEN]);
 
+#define mld_rej_uniform_eta2_avx2_asm MLD_NAMESPACE(rej_uniform_eta2_avx2_asm)
+/* This contract must be kept in sync with the HOL-Light specification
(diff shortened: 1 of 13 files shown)


jakemas:
Superseded by a branch pushed directly to the upstream repo so CI runs (fork PRs skip the self-hosted/EC2 test+bench jobs).

PR #940, opened 2026-02-03
x86_64: Eliminate use_hint 32/88 intrinsics

willieyz:
- Resolves: #484
In this PR, we replace the AVX2 intrinsics implementation of poly_use_hint_32 and poly_use_hint_88 with a x86_64 assembly version, this is part of the effort to enable HOL-Light proofs.

We also tried unrolling the loops: `mld_poly_use_hint_88_avx2_loop` and `mld_poly_use_hint_32_avx2_loop`
in both files. However, the benchmark results showed that this did not provide any performance benefit, so we decided to keep the current version.

- bench components
  - Δ (%) = (asm − AVX2) / AVX2 × 100

| Component            | Implementation      | Build               |  ML-DSA-44 |  ML-DSA-65 |  ML-DSA-87 |   Notes   |
|----------------------|---------------------|---------------------|------------|------------|------------|-----------|
| mld_poly_caddq <br>(avg) | AVX2 intrinsics     | no-opt              |     821    |     781    |    789     |           |
|                      | x86_64 asm          | no-opt              |     847    |     786    |    787     |           |
|                      | Δ (%)               | no-opt              |    +3.17%	|    +0.64%  |   -0.25%   |           |
| mld_poly_caddq <br>(avg) | AVX2 intrinsics     | opt                 |     210    |     147    |    143     |           |
|                      | x86_64 asm          | opt                 |     220    |     153    |    155     |           |
|                      | x86_64 asm <br>(unroll) | opt                 |     273    |     154    |    156     |unroll by 4|
|                      | **Δ (%)**           | **opt**             | **+4.76%** |  **+4.08%**| **+8.39%** |           |
|                      | **Δ (%) (unroll)**  | **opt**             | **+30.00%**|  **+4.76%**| **+9.09%** |unroll by 4|
- bench
  - Δ (%) = (asm − AVX2) / AVX2 × 100

| Component               | Implementation      | Build               |  ML-DSA-44 |  ML-DSA-65 |  ML-DSA-87 |      Notes     |
|-------------------------|---------------------|---------------------|------------|------------|------------|----------------|
| keypair cycles <br>(avg)    | AVX2 intrinsics     | no-opt              |   127436   |   218610   |    360739  | baseline (main)|
|                         | x86_64 asm          | no-opt              |   127459   |   217604   |    367118  |                |
|                         | Δ (%)               | no-opt              |   +0.02%   |  -0.46%    |   +1.77%   |                |
[...]

PR #1189, opened 2026-06-18
x86_64 + HOL-Light: Replace poly_use_hint AVX2 intrinsics with hand-written assembly and HOL-Light proofs

### LINKED ISSUE #484: AVX2: Replace intrinsics implementation of `poly_use_hint` with assembly

### LINKED ISSUE #485: HOL-Light: Prove AVX2 `poly_use_hint`

### LINKED ISSUE #1074: Potentially wrong value for q_bound in poly_use_hint_32_avx2.c

jakemas:
Resolves #484
Resolves #485

This is the only remaining intrinsic to be picked up, and proved. I know @willieyz has a PR in flight, I've taken it, brought it up to main, fixed conflicts, adjusted asm, fixed #1074/#1076, added hol-light CBMC proofs and supporting infra.

Replaces the `poly_use_hint_32` (ML-DSA-65/87) and `poly_use_hint_88` (ML-DSA-44) AVX2 intrinsics with hand-written assembly, and adds HOL-Light functional-correctness and constant-time / memory-safety proofs for both x86_64 routines.

Builds on @willieyz's original assembly (#940). Changes from that code:
- Reworked to the in-place 2-arg signature `(int32_t *a, const int32_t *h)` instead of the 3-arg out-of-place form.
- Renamed to the `_asm` convention and wired the native backend to the assembly (not the `.c` intrinsic).
- Fixed the `poly_use_hint_32` compare bound `0x15BA900` (87·GAMMA2) → `0x7BE100` (31·GAMMA2), per #1074/#1076.

### Performance

The performance degradation we previously saw in willieyz's PR is fixed.
`poly_use_hint` component benchmark, median cycles on Intel Xeon Platinum 8175M, `OPT=1 CYCLES=PMU`:

| Variant | AVX2 intrinsics (baseline, `main`) | @willieyz asm (#940) | Hand-written AVX2 asm (this PR) |
|---|---|---|---|
| `use_hint_32` (ML-DSA-65/87) | ~314 | ~300 | ~303 |
| `use_hint_88` (ML-DSA-44) | ~377 | ~351 | ~356 |


proofs/hol_light/x86_64/proofs/poly_use_hint_32_avx2_asm.ml
```diff
@@ -0,0 +1,1326 @@
...
+let POLY_USE_HINT_32_AVX2_ASM_EXEC =
+  X86_MK_CORE_EXEC_RULE poly_use_hint_32_avx2_asm_tmc;;
+
+(* ------------------------------------------------------------------------- *)
+(* Numeric (code-aligned) form of UseHint, matching the Barrett computation  *)
+(* performed by the assembly. Bridged to the FIPS 204 spec                   *)
+(* (mldsa_use_hint_32 in mldsa_specs.ml) by MLDSA_USE_HINT_32_EQUIV below.   *)
+(* The decompose helper lemmas (A1_BOUND, A1_WRAP, BARRETT_INTERVAL_32) and  *)
+(* the DIV/MOD equivalence tactics are arch-independent and shared with the  *)
+(* AArch64 proof of the same routine.                                        *)
```
mkannwischer: If they are shared, can we move them to a shared file, please?

dev/x86_64/src/poly_use_hint_32_avx2_asm.S
```diff
@@ -0,0 +1,148 @@
...
+ * - Our implementation of decompose() is slightly different from that in
+ *   @[REF_AVX2]. See poly_decompose_32_avx2.c for more information.
+ */
+
+// a aliased with a0
+.macro decompose32_avx2 a1, a, temp1, temp2, temp3
+// Compute a1 = round-(a / 523776) ≈ round(a * 1074791425 /
+// 2^49), where round-() denotes "round half down". This is
+// exact for 0 <= a < Q. Note that half is rounded down since
+// 1074791425 / 2^49 ≲ 1 / 523776.
```
mkannwischer: These comments seem to describe the AArch64 code, not the x86_64 code. Can you please adjust them to describe what this code is doing?

dev/x86_64/src/poly_use_hint_32_avx2_asm.S
```diff
@@ -0,0 +1,148 @@
...
+vpsubd  \temp1, \a, \a
+
+// If wrap-around is required, adjust a0 by -1
+vpaddd  \temp2, \a, \a
+.endm
+
+/* Reference: The reference avx2 implementation checks a0 >= 0, which is
+ * different from the specification and the reference C implementation. We
+ * follow the specification and check a0 > 0.
+ */
```
mkannwischer: This is no longer true since 3 weeks ago: https://github.com/pq-crystals/dilithium/commit/bba9534a3c5cbef66481cd6518d786e23ae601b5. We should reference that commit here

proofs/hol_light/x86_64/proofs/poly_use_hint_32_avx2_asm.ml
```diff
@@ -0,0 +1,1326 @@
...
+
+let BARRETT_INTERVAL_32 = prove(
+  `!a lo hi k.
+    lo <= a /\ a <= hi /\
+    k * 262144 <= (2 * lo * 1074791425) DIV 4294967296 + 131072 /\
+    (2 * hi * 1074791425) DIV 4294967296 + 131072 < (k + 1) * 262144 /\
+    k * 4194304 <= (lo + 127) DIV 128 * 1025 + 2097152 /\
+    (hi + 127) DIV 128 * 1025 + 2097152 < (k + 1) * 4194304
+    ==> ((2 * a * 1074791425) DIV 4294967296 + 131072) DIV 262144 = k /\
+        ((a + 127) DIV 128 * 1025 + 2097152) DIV 4194304 = k`,
```
mkannwischer: I find this rather confusing - this seems to mix the AArch64 branches (`k * 262144 <= (2 * lo * 1074791425) DIV 4294967296 + 131072`) and the x86 branches (` k * 4194304 <= (lo + 127) DIV 128 * 1025 + 2097152 `). But the AArch64 branches are dead in this proof.
It seems easiest to remove the dead branches and not claim that this theorem is shared with AArch64.

mkannwischer (changes requested):
Thanks @jakemas for moving this forward!

proofs/hol_light/x86_64/proofs/poly_use_hint_32_avx2_asm.ml
jakemas: Done

dev/x86_64/src/poly_use_hint_32_avx2_asm.S
jakemas: Done

dev/x86_64/src/poly_use_hint_32_avx2_asm.S
jakemas: Done, ty!

proofs/hol_light/x86_64/proofs/poly_use_hint_32_avx2_asm.ml
jakemas: ok, done!

mldsa/src/native/x86_64/src/poly_use_hint_32_avx2_asm.S
```diff
@@ -0,0 +1,130 @@
...
+      size_bytes: 1024
+      permissions: read-only
+      c_parameter: const int32_t *h
+      description: Hint polynomial (256 x int32_t)
+*/
+
+#include "../../../common.h"
+
+#if defined(MLD_ARITH_BACKEND_X86_64_DEFAULT) &&   \
+    !defined(MLD_CONFIG_MULTILEVEL_NO_SHARED) &&   \
```
mkannwischer, line 50: This is missing the 
```
 !defined(MLD_CONFIG_NO_VERIFY_API) && 
``` 
conditional.

mldsa/src/native/x86_64/src/poly_use_hint_88_avx2_asm.S
```diff
@@ -0,0 +1,133 @@
...
+      size_bytes: 1024
+      permissions: read-only
+      c_parameter: const int32_t *h
+      description: Hint polynomial (256 x int32_t)
+*/
+
+#include "../../../common.h"
+
+#if defined(MLD_ARITH_BACKEND_X86_64_DEFAULT) &&   \
+    !defined(MLD_CONFIG_MULTILEVEL_NO_SHARED) &&   \
```
mkannwischer, line 49: Same here.

mkannwischer (changes requested):
Thanks @jakemas for the updates.
I have one last comment, then this can be merged imo.

mldsa/src/native/x86_64/src/poly_use_hint_32_avx2_asm.S
jakemas: added!

mldsa/src/native/x86_64/src/poly_use_hint_88_avx2_asm.S
jakemas: added!

jakemas:
> Thanks @jakemas for the updates. I have one last comment, then this can be merged imo.

Thank you! Addressed now :)

mkannwischer (approved):
Thanks @jakemas. LGTM.

@hanno-becker, could you take another look?

jakemas:
remove double description per Hanno request

mldsa/src/native/x86_64/src/arith_native_x86_64.h
```diff
@@ -143,11 +143,33 @@ __contract__(
...
+void mld_poly_use_hint_32_avx2_asm(int32_t *a, const int32_t *h)
+/* This must be kept in sync with the HOL-Light specification
+ * in proofs/hol_light/x86_64/proofs/poly_use_hint_32_avx2_asm.ml */
+__contract__(
+  requires(memory_no_alias(a, sizeof(int32_t) * MLDSA_N))
+  requires(memory_no_alias(h, sizeof(int32_t) * MLDSA_N))
+  requires(array_bound(a, 0, MLDSA_N, 0, MLDSA_Q))
+  requires(array_bound(h, 0, MLDSA_N, 0, 2))
+  assigns(memory_slice(a, sizeof(int32_t) * MLDSA_N))
+  ensures(array_bound(a, 0, MLDSA_N, 0, (MLDSA_Q - 1) / (2 * MLDSA_GAMMA2)))
```
hanno-becker: We can't use level-dependent constants in this file, so need to spell this out unfortunately.

mldsa/src/native/x86_64/src/arith_native_x86_64.h
```diff
@@ -143,11 +143,33 @@ __contract__(
...
+void mld_poly_use_hint_88_avx2_asm(int32_t *a, const int32_t *h)
+/* This must be kept in sync with the HOL-Light specification
+ * in proofs/hol_light/x86_64/proofs/poly_use_hint_88_avx2_asm.ml */
+__contract__(
+  requires(memory_no_alias(a, sizeof(int32_t) * MLDSA_N))
+  requires(memory_no_alias(h, sizeof(int32_t) * MLDSA_N))
+  requires(array_bound(a, 0, MLDSA_N, 0, MLDSA_Q))
+  requires(array_bound(h, 0, MLDSA_N, 0, 2))
+  assigns(memory_slice(a, sizeof(int32_t) * MLDSA_N))
+  ensures(array_bound(a, 0, MLDSA_N, 0, (MLDSA_Q - 1) / (2 * MLDSA_GAMMA2)))
```
hanno-becker: We can't use level-dependent constants in this file, so need to spell this out unfortunately.

hanno-becker: @mkannwischer We need to do something about this... this is far too easy to get wrong.

hanno-becker (changes requested):
We need to remove level-dependent constants from the backend spec (and ideally find a way to detect this in CI).

Otherwise, this looks great so far

mldsa/src/native/x86_64/src/arith_native_x86_64.h
hanno-becker: On the other hand, the CBMC specs are never used in a multi-level build setting, and only in that case would this matter.

mkannwischer: Good catch. It is still confusing. We should ban level dependent macros in all level-independent files.
Maybe a linter is sufficient?

mldsa/src/native/x86_64/src/arith_native_x86_64.h
jakemas: thanks, fixed!

mldsa/src/native/x86_64/src/arith_native_x86_64.h
jakemas: fixed!

jakemas:
> We need to remove level-dependent constants from the backend spec (and ideally find a way to detect this in CI).
> 
> Otherwise, this looks great so far

Thank you, fixed and rebased on main

hanno-becker (approved):

PR #1190, opened 2026-06-19
ppc64: POWER8 VSX backend, little- and big-endian (caddq, pointwise/acc, NTT/inverse-NTT)

Scottcjn:
This adds a **POWER8 (ppc64le) VSX backend** to mldsa-native. The repository
previously had aarch64 and x86_64 native backends; this is the first ppc64le
one. It is selected automatically via `MLD_SYS_PPC64LE` and replaces the
following primitives:

| primitive | approach |
|-----------|----------|

### Montgomery reduction
The int32 reduction is done in the word domain: even/odd 32×32→64 products via
`vmulesw`/`vmulosw`, `m = low32(prod * QINV)` via `vmuluwm`, then
`(prod - m*q) >> 32` taken as the high words. Low words are recombined with
`vmrgow` and high words with `vmrgew` — the same even/odd convention on both
ends, a closed permutation back to input coefficient order, so the kernels do
not depend on a global VSX lane-order assumption.

### NTT / inverse NTT
Natural coefficient order, matching the C reference exactly. The outer six
layers (len ≥ 4) are vectorized four coefficients at a time with a per-block
broadcast twiddle; the inner two layers (len = 2, 1) have within-vector
partners and run scalar. Twiddles are streamed from a natural-order table
(one per block, `k` ascending for the forward transform, descending + negated
for the inverse), so twiddle addressing is a single moving pointer.

### Notes
- The backend files are currently hand-maintained: `scripts/cfify` does not yet
  emit CFI for ppc64le and the autogen has no ppc64le simpasm job, so `dev/` and
  `mldsa/src/native/ppc64le/` are kept in sync directly. Happy to wire up
  autogen support for ppc64le as a follow-up if that is preferred.
- Inner NTT layers use a scalar fallback for now; vectorizing them is a natural
  follow-up optimization.

### Testing
Validated on an IBM POWER8 S824 (Ubuntu 20.04, gcc, `-mcpu=power8`):
- `tests unit` — `poly_ntt`/`poly_invntt_tomont` match the C reference
  (zeros / single / random inputs).
- `tests func` and `tests kat` — ML-DSA-44 / 65 / 87 pass with the native
  backend active.


hanno-becker:
@Scottcjn Thank you for the work. We don't currently accept contributions from people we don't know. If you would like to become a contributor for mldsa-native, please reach out to myself and @mkannwischer directly, e.g. via e-mail or discord. Until then, closing this PR.

Scottcjn:
Pushed an additional commit that also enables this backend on **big-endian** POWER (ppc64 BE). The kernels needed no changes — they rely only on architectural (endian-invariant) VSX element numbering and a closed even/odd permutation in the Montgomery reduction, so they are bit-exact on both endians. The commit is just plumbing: `MLD_SYS_PPC64BE` detection in `sys.h`, selecting the backend for it in `native/meta.h`, and a big-endian `powerpc64` cross prefix in `test/mk/auto.mk`.

Validated big-endian under `qemu-ppc64` (`powerpc64-linux-gnu`, `-mcpu=power8`): `poly_ntt`/`poly_invntt_tomont` unit tests match the C reference and ML-DSA-44/65/87 pass func + KAT. The little-endian native path on a POWER8 S824 is unchanged and still passes. Happy to split this into a separate PR if you'd prefer to land the little-endian backend first.

PR #1191, opened 2026-06-19
ppc64: POWER8 VSX backend, little- and big-endian (caddq, pointwise/acc, NTT/inverse-NTT)

Scottcjn:
This adds a **POWER8 (ppc64) VSX backend** to mldsa-native, working on **both
little- and big-endian** POWER. The repository previously had aarch64 and
x86_64 native backends; this is the first ppc64 one. It is selected
automatically via `MLD_SYS_PPC64LE` / `MLD_SYS_PPC64BE` and replaces:

| primitive | approach |
|-----------|----------|

### Montgomery reduction
int32 reduction in the word domain: even/odd 32×32→64 products via
`vmulesw`/`vmulosw`, `m = low32(prod * QINV)` via `vmuluwm`, then
`(prod - m*q) >> 32` taken as the high words. Low words are recombined with
`vmrgow` and high words with `vmrgew` — the same even/odd convention on both
ends, a closed permutation back to input coefficient order. Because this relies
only on the **architectural (endian-invariant) VSX element numbering** and never
on where a load physically places coefficients, the kernels are bit-exact on
**both endians with no code differences**.

### NTT / inverse NTT
Natural coefficient order, matching the C reference exactly. The outer six
layers (len ≥ 4) are vectorized four coefficients at a time with a per-block
broadcast twiddle; the inner two layers (len = 2, 1) have within-vector
partners and run scalar. Twiddles stream from a natural-order table (one per
block, `k` ascending for the forward transform, descending + negated for the
inverse), so twiddle addressing is a single moving pointer.

### Notes
- The backend files are hand-maintained: `scripts/cfify` does not yet emit CFI
  for ppc64 and the autogen has no ppc64 simpasm job, so `dev/` and
  `mldsa/src/native/ppc64le/` are kept in sync directly. Happy to wire up
  autogen support as a follow-up.
- Inner NTT layers use a scalar fallback for now; vectorizing them is a natural
  follow-up optimization.

### Testing
- **Little-endian** on an IBM POWER8 S824 (`-mcpu=power8`): `tests unit` matches
  the C reference for `poly_ntt`/`poly_invntt_tomont`; `tests func` and
  `tests kat` pass for ML-DSA-44 / 65 / 87.
- **Big-endian** under `qemu-ppc64` (`powerpc64-linux-gnu`, `-mcpu=power8`):
  same unit / func / KAT suites pass with the native backend active.

(Supersedes #1190, which got wedged closed by a stuck head-ref sync after a
follow-up push; same branch, now with the big-endian commit included.)


hanno-becker:
@Scottcjn Thank you for the work. We don't currently accept contributions from people we don't know. If you would like to become a contributor for mldsa-native, please reach out to myself and @mkannwischer directly, e.g. via e-mail or discord. Until then, closing this PR.

PR #1192, opened 2026-06-20
CI: Trim unnecessary fetch-depth from checkout steps

mkannwischer:
Only the jobs that use tj-actions/changed-files need git history. Drop fetch-depth: 0 from the checkout steps that don't and document why the changed-files jobs keep full history.

### DIFF
--- a/.github/workflows/hol_light.yml
+++ b/.github/workflows/hol_light.yml
@@ -52,8 +52,6 @@ jobs:
     if: github.repository_owner == 'pq-code-package' && !github.event.pull_request.head.repo.fork
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/hol_light.yml
```diff
@@ -133,6 +129,7 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
+          # Full history needed by changed-files (rebase-merge pushes can be multi-commit)
```
hanno-becker: Can one be pragmatic here and check out, say, the last 100 commits?

mkannwischer: We could do that, yes.

mkannwischer: Done.

hanno-becker (approved):

PR #1195, opened 2026-06-23
Add ABI checker for AArch64, x86_64 (SysV), Armv8.1-M+MVE

hanno-becker:
* Port of https://github.com/pq-code-package/mlkem-native/pull/1135

Verify that each assembly kernel preserves the callee-saved registers its platform calling convention requires:

- AArch64 (AAPCS64):   x19-x28, x29/FP, lower 64 bits of d8-d15.
- x86_64 (System V):   rbx, rbp, r12-r15. No SIMD register callee-saved.
- Armv8.1-M (AAPCS32): r4-r11 plus MVE Q4-Q7 (= D8-D15).

A per-arch assembly call stub loads a random register state, calls the kernel, and captures the result; a checker then confirms the callee-saved set is intact. Pointer arguments are backed by correctly-sized buffers whose layout comes from a per-kernel YAML block; scripts/autogen turns that YAML into the per-kernel checks. A self-test of hand-written corrupters confirms the checker actually fires before kernel verdicts are trusted.

Run via `make run_abicheck OPT=1` or `scripts/tests abicheck`. It needs no library build, so `scripts/tests all` runs it by default.

Ported from mlkem-native. Notable deviations:
- No PowerPC64/ELFv2 backend: mldsa-native ships no ppc64le assembly, so that arch and its callstub/selftest/registry are dropped.
- mldsa kernel set and signatures (int32_t coefficients, qdata/zeta tables, polyvecl_acc l4/l5/l7, decompose/use_hint 32/88, polyz_unpack 17/19, rej_uniform eta2/eta4) get fresh ABI YAML blocks.
- Fixed the FIPS202 Keccak YAML Name fields, which did not match the exported symbols (e.g. keccak_f1600_x1_scalar_asm vs the actual keccak_f1600_x1_scalar_aarch64_asm); the generator derives the C symbol as mld_<Name>, so this was a latent mismatch.

Also brings mldsa/src/sys.h into line with mlkem-native by adding MLD_SYS_CAP_MVE to the capability enum (and the autogen syscap allowlist): mlkem-native already carries MLK_SYS_CAP_MVE, but mldsa's enum only had AVX2/SHA3. The Armv8.1-M abicheck needs it for its runtime capability check.

See test/abicheck/README.md for details.


mkannwischer (approved):
Thanks @hanno-becker. As far as I can tell this is faithful port.

PR #1206, opened 2026-06-26
Config: Remove legacy configuration

### LINKED ISSUE #1202: Config: Remove legacy configuration
mldsa-native still carries the the legacy configuration options:

https://github.com/pq-code-package/mldsa-native/blob/13977d0a2b613e7f1ea2c011d4e0c760ff8992ff/mldsa/mldsa_native.h#L34-L50

As far as I know, none of our consumers is still using that configuration. I suggest removing the configuration prior to a non-beta release. mlkem-native will also remove that config in v2.

mkannwischer:
Drop the deprecated MLD_CONFIG_API_* configuration path from the public header. Builds now exclusively use the MLD_CONFIG_* options from the configuration file.

Consumers that set the legacy MLD_CONFIG_API_* options directly before including mldsa_native.h should instead configure the build through the configuration file (mldsa_native_config.h, or MLD_CONFIG_FILE), renaming:

  MLD_CONFIG_API_PARAMETER_SET    -> MLD_CONFIG_PARAMETER_SET
  MLD_CONFIG_API_NAMESPACE_PREFIX -> MLD_CONFIG_NAMESPACE_PREFIX
  MLD_CONFIG_API_NO_SUPERCOP      -> MLD_CONFIG_NO_SUPERCOP
  MLD_CONFIG_API_CONSTANTS_ONLY   -> MLD_CONFIG_CONSTANTS_ONLY
  MLD_CONFIG_API_QUALIFIER        -> MLD_CONFIG_EXTERNAL_API_QUALIFIER

For a multi-level build, set MLD_CONFIG_MULTILEVEL_BUILD; the parameter set is now appended to MLD_CONFIG_NAMESPACE_PREFIX automatically and no longer needs to be baked into the prefix.

 - Resolves #1202


mldsa/mldsa_native.h
```diff
@@ -31,45 +31,6 @@
...
- *   but only constants for the sizes of key material, set
- *   MLD_CONFIG_API_CONSTANTS_ONLY. In this case, you don't need to set
- *   MLD_CONFIG_API_PARAMETER_SET or MLD_CONFIG_API_NAMESPACE_PREFIX,
- *   nor include a configuration.
- *
- * - MLD_CONFIG_API_QUALIFIER [optional]
- *
- *   Qualifier to apply to external API.
- *
  ******************************************************************************/
```
hanno-becker: Nit, pre-existing: This initial comment block is inconsistently delimited: A plain `/* ..` at the beginning, but `*****/` at the end. More prominent now that the legacy config doc is gone.

hanno-becker:
```suggestion
 */
```

mldsa/mldsa_native.h
```diff
@@ -144,62 +105,38 @@
...
 #define MLD_API_CONCAT_UNDERSCORE(x, y) MLD_API_CONCAT(MLD_API_CONCAT(x, _), y)
 
-#if !defined(MLD_CONFIG_API_PARAMETER_SET)
-/* Recommended configuration via same config file as used for the build. */
-
-/* For now, we derive the legacy API configuration MLD_CONFIG_API_XXX from
- * the config file. In mldsa-native-v2, this will be removed and we will
- * exclusively work with MLD_CONFIG_XXX. */
-
 /* You need to make sure the config file is in the include path. */
```
hanno-becker: Pre-existing: This is really a comment for `mldsa_native_config.h`.

mldsa/mldsa_native.h
```diff
@@ -144,62 +105,38 @@
 #define MLD_API_CONCAT(x, y) MLD_API_CONCAT_(x, y)
 #define MLD_API_CONCAT_UNDERSCORE(x, y) MLD_API_CONCAT(MLD_API_CONCAT(x, _), y)
```
hanno-becker: Pre-existing: This blob of macros and namespacing might be better placed in a separate `Namespacing and Qualifiers` section, so that when the user scrolls to `Function API`, they actually see the functions immediately.

mldsa/mldsa_native.h
hanno-becker: Not changing this here.

mldsa/mldsa_native.h
hanno-becker: Fixed in follow-up

mldsa/mldsa_native.h
hanno-becker: Fixed in follow-up

hanno-becker (approved):
I'm very happy to see this go. LGTM

PR #1209, opened 2026-06-26
CI: Apply clang 22 macOS sanitizer fix to build_kat too

hanno-becker:
Same macOS 26 ASan startup deadlock as the preceding backend_tests fix. Set the macos-latest target's nix_shell to 'clang22', and pass nix-shell to build_kat's sanitizer step uniformly with the other steps (it omitted it and fell back to the action default). Other targets are unchanged.


mkannwischer (approved):
Thanks!

PR #1220, opened 2026-06-26
Reject MLD_PREHASH_NONE in prehash APIs

### LINKED ISSUE #1215: Prehash API accepts `MLD_PREHASH_NONE` and ignores the digest
The public unstable HashML-DSA APIs `signature_pre_hash_internal` and
`verify_pre_hash_internal` accept `MLD_PREHASH_NONE` even though their supported
prehash algorithm list only includes real hash/XOF algorithms. When `hashalg` is
`MLD_PREHASH_NONE`, the supplied prehash buffer `ph` and `phlen` are not
validated or incorporated into the formatted message. A signature produced over
one digest therefore verifies against a different digest as long as both calls use
`MLD_PREHASH_NONE`.

This breaks message binding for callers that accidentally pass the exported
`MLD_PREHASH_NONE` constant to the prehash API, for example through an
uninitialized/default algorithm selector or a configuration mapping bug.

fegge:
## Summary
- Reject `MLD_PREHASH_NONE` in the unstable pre-hash signing and verification APIs.
- Document that `MLD_PREHASH_NONE` is not supported by these HashML-DSA entry points.
- Add functional coverage for the rejected pre-hash algorithm.

## Validation
- `clang-format -i mldsa/src/sign.c mldsa/src/sign.h mldsa/mldsa_native.h test/src/test_mldsa.c`: passed
- `git diff --check -- mldsa/src/sign.c mldsa/src/sign.h mldsa/mldsa_native.h test/src/test_mldsa.c`: passed
- `make run_func -j4`: passed

Fixes #1215

---

### DIFF
--- a/mldsa/mldsa_native.h
+++ b/mldsa/mldsa_native.h
@@ -637,6 +637,7 @@ int MLD_API_NAMESPACE(open)(
  *   MLD_PREHASH_SHA2_512, MLD_PREHASH_SHA2_512_224, MLD_PREHASH_SHA2_512_256,
  *   MLD_PREHASH_SHA3_224, MLD_PREHASH_SHA3_256, MLD_PREHASH_SHA3_384,
  *   MLD_PREHASH_SHA3_512, MLD_PREHASH_SHAKE_128, MLD_PREHASH_SHAKE_256.
+ * MLD_PREHASH_NONE is rejected by this HashML-DSA API.
  *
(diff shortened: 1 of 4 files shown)


fegge:
Reopened from an upstream repository branch with DCO-signed commits so full CI can run: https://github.com/pq-code-package/mldsa-native/pull/1228

PR #1221, opened 2026-06-26
Document signing secret-key validity precondition

### LINKED ISSUE #1216: Signing accepts invalid secret keys that `pk_from_sk` rejects
The signing APIs decode the secret key and use `s1`, `s2`, and `t0` without
validating that the decoded secret-key components are canonical and within the
ML-DSA coefficient bounds. The separate `pk_from_sk` API performs those checks
and rejects the same malformed keys.

Wycheproof marks these test keys with `InvalidPrivateKey`. The normal
Wycheproof signing harness skips those signing tests because signing currently
does not validate secret keys, then tests them only through `pkFromSk`. When run
directly, `sigGenDeterministic` succeeds for the invalid private keys and the
resulting signatures verify under the corresponding public keys.

fegge:
## Summary
- Document that signing APIs require a valid secret key.
- Point callers to `pk_from_sk` when they need to validate a serialized secret key before signing.
- Add a Wycheproof client comment explaining why `InvalidPrivateKey` signing vectors are skipped while `pkFromSk` still tests those keys.

## Validation
- `clang-format -i mldsa/src/sign.h mldsa/mldsa_native.h`: passed
- `git diff --check -- mldsa/src/sign.h mldsa/mldsa_native.h test/wycheproof/wycheproof_client.py`: passed
- `make run_func -j4`: passed
- `python3 test/wycheproof/wycheproof_client.py --data-dir <fresh-temp-dir>`: passed

Fixes #1216

---

### DIFF
--- a/mldsa/mldsa_native.h
+++ b/mldsa/mldsa_native.h
@@ -315,7 +315,11 @@ int MLD_API_NAMESPACE(keypair)(
  * @param      prelen     Length of prefix string. Ignored when
  *                        externalmu != 0.
  * @param[in]  rnd        Random seed.
- * @param[in]  sk         Bit-packed secret key.
+ * @param[in]  sk         Bit-packed valid secret key. Signing does not perform
+ *                        full serialized secret-key validation; callers
+ *                        importing serialized keys can use
+ *                        crypto_sign_pk_from_sk to validate them before
+ *                        signing.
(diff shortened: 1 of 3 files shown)


fegge:
Reopened from an upstream repository branch with DCO-signed commits so full CI can run: https://github.com/pq-code-package/mldsa-native/pull/1229

PR #1222, opened 2026-06-26
Preserve in-place message on signing failure

### LINKED ISSUE #1217: In-place signing failure wipes the input message
The public `crypto_sign` API documents that the signed-message output buffer `sm` can be equal to the message input buffer `m`. In that in-place mode, `mld_sign` first copies the message from the beginning of `sm` to `sm + MLDSA_CRYPTO_BYTES` at `mldsa/src/sign.c:1156` through `mldsa/src/sign.c:1164`. It then calls `mld_sign_signature`.

If signing fails after this copy, for example because the context length is greater than 255 and domain-prefix construction returns zero at `mldsa/src/sign.c:1610`, `mld_sign_signature` clears the detached-signature buffer at `mldsa/src/sign.c:1094` through `mldsa/src/sign.c:1095`. In in-place mode, that signature buffer is the start of the caller's original message buffer. The call returns `MLD_ERR_FAIL`, sets `smlen` to zero, and leaves the caller-visible original message wiped even though the message was an input.

fegge:
## Summary
- Preserve the caller's in-place message buffer when `crypto_sign` fails before producing a signature.
- Document the failure behavior for signed-message APIs.
- Add regression coverage for in-place signing failure and RNG-failure paths.

## Validation
- `clang-format -i mldsa/src/sign.c mldsa/src/sign.h mldsa/mldsa_native.h test/src/test_mldsa.c test/src/test_rng_fail.c`: passed
- `git diff --check -- mldsa/src/sign.c mldsa/src/sign.h mldsa/mldsa_native.h test/src/test_mldsa.c test/src/test_rng_fail.c`: passed
- `make run_func run_rng_fail -j4`: passed

Fixes #1217

---

### DIFF
--- a/mldsa/mldsa_native.h
+++ b/mldsa/mldsa_native.h
@@ -444,6 +444,9 @@ int MLD_API_NAMESPACE(signature_extmu)(
  *                     MLD_CONFIG_CONTEXT_PARAMETER is defined; type set by
  *                     MLD_CONFIG_CONTEXT_PARAMETER_TYPE.
  *
+ * On failure, *smlen is set to 0. No partial signed-message output is
+ * preserved, but if sm == m, the input message bytes remain unchanged.
+ *
  * @retval 0                               Success.
  * @retval MLD_ERR_OUT_OF_MEMORY           MLD_CONFIG_CUSTOM_ALLOC_FREE was
  *                                         used and an allocation via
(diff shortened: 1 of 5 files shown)


fegge:
Reopened from an upstream repository branch with DCO-signed commits so full CI can run: https://github.com/pq-code-package/mldsa-native/pull/1230

PR #1229, opened 2026-06-26
Document signing secret-key validity precondition

### LINKED ISSUE #1216: Signing accepts invalid secret keys that `pk_from_sk` rejects

fegge:
## Summary
- Document that signing APIs require a valid secret key.
- Point callers to `pk_from_sk` when they need to validate a serialized secret key before signing.
- Add a Wycheproof client comment explaining why `InvalidPrivateKey` signing vectors are skipped while `pkFromSk` still tests those keys.

## Validation
- `clang-format -i mldsa/src/sign.h mldsa/mldsa_native.h`: passed
- `git diff --check -- mldsa/src/sign.h mldsa/mldsa_native.h test/wycheproof/wycheproof_client.py`: passed
- `make run_func -j4`: passed
- `python3 test/wycheproof/wycheproof_client.py --data-dir <fresh-temp-dir>`: passed

Fixes #1216

---

### DIFF
--- a/mldsa/src/sign.h
+++ b/mldsa/src/sign.h
@@ -180,6 +180,10 @@ __contract__(
  *
  * @spec{Implements @[FIPS204, Algorithm 7, ML-DSA.Sign_internal].}
  *
+ * @warning This function does not perform secret key validation.
+ *          Callers importing serialized keys can use mld_sign_pk_from_sk
+ *          to validate them before signing.
+ *
  * @param[out] sig        Output signature.
  * @param[out] siglen     Pointer to output length of signature.
  * @param[in]  m          Pointer to message to be signed (when
@@ -192,7 +196,7 @@ __contract__(
  * @param      prelen     Length of prefix string. Ignored when
  *                        externalmu != 0.
  * @param[in]  rnd        Random seed.
- * @param[in]  sk         Bit-packed secret key.
+ * @param[in]  sk         Bit-packed secret key; assumed to be valid.
  * @param      externalmu 0: m/mlen is the raw message; mu = H(tr, pre, m) is
  *                        computed internally.
  *                        non-zero: m points to a precomputed mu of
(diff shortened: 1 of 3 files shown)


mldsa/src/sign.h
```diff
@@ -194,7 +194,10 @@ __contract__(
  * @param      prelen     Length of prefix string. Ignored when
  *                        externalmu != 0.
  * @param[in]  rnd        Random seed.
- * @param[in]  sk         Bit-packed secret key.
+ * @param[in]  sk         Bit-packed valid secret key. Signing does not perform
+ *                        full serialized secret-key validation; callers
```
hanno-becker: Should we make this more prominent? E.g.

```
Bit-packed secret key; assumed to be valid.
```

and further above/below

```
@warning This function does not perform secret key validation.
         Callers importing serialized keys can use mld_sign_pk_from_sk
         to validate them before signing.
```

@mkannwischer wdyt

It also sheds light that we might want to provide an explicit `mld_sign_sk_validate` or similar (even if it's just a dummy around `mld_sign_pk_from_sk`). Proposing to call `mld_sign_pk_from_sk` for key validation and throwing away `pk` is, at best, unintuitive.

hanno-becker (changes requested):
Thank you @fegge, I support this change. One could even make the comment more prominent, see comment. @mkannwischer wdyt

hanno-becker (approved):
LGTM

mkannwischer (approved):

PR #1230, opened 2026-06-26
Preserve in-place message on signing failure

### LINKED ISSUE #1217: In-place signing failure wipes the input message

fegge:
## Summary
- Preserve the caller's in-place message buffer when `crypto_sign` fails before producing a signature.
- Document the failure behavior for signed-message APIs.
- Add regression coverage for in-place signing failure and RNG-failure paths.

## Validation
- `clang-format -i mldsa/src/sign.c mldsa/src/sign.h mldsa/mldsa_native.h test/src/test_mldsa.c test/src/test_rng_fail.c`: passed
- `git diff --check -- mldsa/src/sign.c mldsa/src/sign.h mldsa/mldsa_native.h test/src/test_mldsa.c test/src/test_rng_fail.c`: passed
- `make run_func run_rng_fail -j4`: passed

Fixes #1217

---

### DIFF
--- a/mldsa/src/sign.c
+++ b/mldsa/src/sign.c
@@ -1151,7 +1151,16 @@ int mld_sign(uint8_t *sm, size_t *smlen, const uint8_t *m, size_t mlen,
              MLD_CONFIG_CONTEXT_PARAMETER_TYPE context)
 {
   int ret;
+  uint8_t sig[MLDSA_CRYPTO_BYTES];
+  size_t siglen;
   size_t i;
+
+  ret = mld_sign_signature(sig, &siglen, m, mlen, ctx, ctxlen, sk, context);
+  if (ret != 0)
+  {
+    *smlen = 0;
+    goto cleanup;
+  }
 
   for (i = 0; i < mlen; ++i)
   __loop__(
@@ -1162,12 +1171,12 @@ int mld_sign(uint8_t *sm, size_t *smlen, const uint8_t *m, size_t mlen,
   {
     sm[MLDSA_CRYPTO_BYTES + mlen - 1 - i] = m[mlen - 1 - i];
   }
-  ret = mld_sign_signature(sm, smlen, sm + MLDSA_CRYPTO_BYTES, mlen, ctx,
-                           ctxlen, sk, context);
-  if (ret == 0)
-  {
-    *smlen += mlen;
-  }
+
+  mld_memcpy(sm, sig, MLDSA_CRYPTO_BYTES);
+  *smlen = siglen + mlen;
+
+cleanup:
+  mld_zeroize(sig, sizeof(sig));
   return ret;
 }
 #endif /* !MLD_CONFIG_NO_RANDOMIZED_API */
--- a/mldsa/src/sign.h
+++ b/mldsa/src/sign.h
@@ -259,6 +259,9 @@ __contract__(
  *                     MLD_CONFIG_CONTEXT_PARAMETER is defined; type set by
  *                     MLD_CONFIG_CONTEXT_PARAMETER_TYPE.
  *
+ * On failure, *smlen is set to 0. No partial signed-message output is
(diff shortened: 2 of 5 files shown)


mldsa/src/sign.h
```diff
@@ -259,6 +259,9 @@ __contract__(
  *                     MLD_CONFIG_CONTEXT_PARAMETER is defined; type set by
  *                     MLD_CONFIG_CONTEXT_PARAMETER_TYPE.
  *
+ * On failure, *smlen is set to 0. No partial signed-message output is
```
hanno-becker, line 262: The parameter is called `sig` here. Can we align this across `sign.h`, `api.h`, and `sign.c`, while at it?

I think `sig` and `siglen` are more intuitive than `sm/smlen`. @mkannwischer wdyt

mldsa/src/sign.c
```diff
@@ -1151,8 +1151,17 @@ int mld_sign(uint8_t *sm, size_t *smlen, const uint8_t *m, size_t mlen,
...
+  size_t siglen;
   size_t i;
 
+  ret = mld_sign_signature(sig, &siglen, m, mlen, ctx, ctxlen, sk, context);
+  if (ret != 0)
+  {
+    *smlen = 0;
+    goto cleanup;
+  }
+
```
hanno-becker, line 1164: Hmm, I'm not too happy with this additional stack allocation (NB: large allocations should come from `MLD_ALLOC/FREE`, but that is a simple change).

I agree, though, that's unintuitive that the message is corrupted upon error.

Alternative: If signing fails and `sm == m` (the only interesting case), copy the message from the tail of `sm` to `m`. This isn't pretty, but it does the job, because that tail of `sm` is guaranteed not to be touched.

Leveling up, though: Why do we need this API in the first place? @mkannwischer can you share some perspectives?

hanno-becker (changes requested):
I agree that the function's behavior is not ideal and should, at the very least, be documented if not changed.

However, I'm hesitant about the additional allocation of `MLDSA_CRYPTO_BYTES` bytes here.

I see three alternative options; @mkannwischer please chime in:
1. Just document and warn about the behavior.
2. For `sm == m`, restore the message upon failure.
3. Remove the API altogether.

mkannwischer:
> I agree that the function's behavior is not ideal and should, at the very least, be documented if not changed.
> 
> However, I'm hesitant about the additional allocation of `MLDSA_CRYPTO_BYTES` bytes here.
> 
> I see three alternative options; @mkannwischer please chime in:
> 
> 1. Just document and warn about the behavior.
> 2. For `sm == m`, restore the message upon failure.
> 3. Remove the API altogether.

I propose to just remove the the API altogether. Consumers needing the signed message API can straightforwardly construct it themselves from the other APIs. I have prepared a draft removing the API here: https://github.com/pq-code-package/mldsa-native/pull/1236

mkannwischer:
superseded by https://github.com/pq-code-package/mldsa-native/pull/1236. 

Thanks @fegge for initiating the discussion!

PR #1231, opened 2026-06-26
Reject invalid API inputs before unsafe work

### LINKED ISSUE #1212: Public APIs perform unsafe work before validating available input lengths
Several public API entry points receive caller-controlled lengths that are
sufficient to reject malformed input, but they perform memory access or
allocator-controlled work before validating those lengths. Attackers who
control message lengths, signature lengths, context lengths, or low-level
external-mu mode lengths can trigger out-of-bounds access, undefined behavior,
null pointer dereferences, or allocator-dependent error paths before the API
rejects the request.

The signed-message API is the highest-impact instance. It receives `mlen`, then
uses `MLDSA_CRYPTO_BYTES + mlen` in indexed memory accesses before checking
whether that addition fits in `size_t` or whether the implementation supports
such a large input. A malformed `mlen` near `SIZE_MAX` can make the output index
wrap and write outside the caller-provided output buffer.

```c
  for (i = 0; i < mlen; ++i)
  __loop__(
    assigns(i, object_whole(sm))
    invariant(i <= mlen)
    decreases(mlen - i)
  )
  {
    sm[MLDSA_CRYPTO_BYTES + mlen - 1 - i] = m[mlen - 1 - i];
  }
  ret = mld_sign_signature(sm, smlen, sm + MLDSA_CRYPTO_BYTES, mlen, ctx,
                           ctxlen, sk, context);
```
Figure X.1: The signed-message API uses `mlen` in output indexing before checking length arithmetic ([mldsa-native/mldsa/src/sign.c#L1156-L1166](https://github.com/pq-code-package/mldsa-native/blob/7750d741d57c0a459f59dae6dd961309f5f70a58/mldsa/src/sign.c#L1156-L1166))

The low-level signing and verification APIs also receive an explicit `mlen`
argument in external-mu mode. Their public contracts state that `mlen` must
equal `MLDSA_CRHBYTES` when `externalmu != 0`, but the implementation copies
`MLDSA_CRHBYTES` bytes from `m` without first enforcing that mode-specific
length rule.

```c
  else
  {
    /* mu has been provided directly (external-mu variant; line 6 done by the
     * caller in a separate cryptographic module). */
    mld_memcpy(mu, m, MLDSA_CRHBYTES);
  }
```
Figure X.2: The external-mu signing path copies 64 bytes before enforcing `mlen == MLDSA_CRHBYTES` ([mldsa-native/mldsa/src/sign.c#L948-L953](https://github.com/pq-code-package/mldsa-native/blob/7750d741d57c0a459f59dae6dd961309f5f70a58/mldsa/src/sign.c#L948-L953))

```c
  else
  {
    /* mu has been provided directly */
    mld_memcpy(mu, m, MLDSA_CRHBYTES);
  }
```
Figure X.3: The external-mu verification path has the same unchecked fixed-size copy ([mldsa-native/mldsa/src/sign.c#L1235-L1239](https://github.com/pq-code-package/mldsa-native/blob/7750d741d57c0a459f59dae6dd961309f5f70a58/mldsa/src/sign.c#L1235-L1239))

Other retained cases have the same property: the API has the length needed to
reject the input before the risky operation. Detached signing allocates scratch
buffers before checking `ctxlen <= 255`, detached verification allocates all
verifier scratch buffers before checking `siglen == MLDSA_CRYPTO_BYTES`, and
the public prefix helper documents `ctx` as nullable but copies from it whenever
`ctxlen > 0`.

```c
  MLD_ALLOC(pre, uint8_t, MLD_DOMAIN_SEPARATION_MAX_BYTES, context);
  MLD_ALLOC(rnd, uint8_t, MLDSA_RNDBYTES, context);

  if (pre == NULL || rnd == NULL)
  {
    ret = MLD_ERR_OUT_OF_MEMORY;
    goto cleanup;
  }

  /* Prepare domain separation prefix for pure ML-DSA */
  pre_len = mld_prepare_domain_separation_prefix(pre, NULL, 0, ctx, ctxlen,
                                                 MLD_PREHASH_NONE);
```
Figure X.4: Detached signing allocates before rejecting overlong contexts ([mldsa-native/mldsa/src/sign.c#L1055-L1070](https://github.com/pq-code-package/mldsa-native/blob/7750d741d57c0a459f59dae6dd961309f5f70a58/mldsa/src/sign.c#L1055-L1070))

```c
  MLD_ALLOC(buf, uint8_t, (MLDSA_K * MLDSA_POLYW1_PACKEDBYTES), context);
  MLD_ALLOC(mu, uint8_t, MLDSA_CRHBYTES, context);
  MLD_ALLOC(c, uint8_t, MLDSA_CTILDEBYTES, context);
  MLD_ALLOC(c2, uint8_t, MLDSA_CTILDEBYTES, context);
  MLD_ALLOC(z, mld_polyvecl, 1, context);
  MLD_ALLOC(cp, mld_poly, 1, context);
  MLD_ALLOC(mat, mld_polymat, 1, context);
  MLD_ALLOC(w1, mld_poly, 1, context);
  MLD_ALLOC(tmp, mld_poly, 1, context);

  if (buf == NULL || mu == NULL || c == NULL || c2 == NULL || z == NULL ||
      cp == NULL || mat == NULL || w1 == NULL || tmp == NULL)
  {
    ret = MLD_ERR_OUT_OF_MEMORY;
    goto cleanup;
  }

  if (siglen != MLDSA_CRYPTO_BYTES)
```
Figure X.5: Verification allocates scratch buffers before checking the caller-provided signature length ([mldsa-native/mldsa/src/sign.c#L1190-L1208](https://github.com/pq-code-package/mldsa-native/blob/7750d741d57c0a459f59dae6dd961309f5f70a58/mldsa/src/sign.c#L1190-L1208))

```c
  /* Common prefix: 0x00/0x01 || ctxlen || ctx */
  prefix[0] = (hashalg == MLD_PREHASH_NONE) ? 0 : 1;
  prefix[1] = (uint8_t)ctxlen;
  if (ctxlen > 0)
  {
    mld_memcpy(prefix + 2, ctx, ctxlen);
  }
```
Figure X.6: The prefix helper copies from `ctx` whenever `ctxlen` is nonzero ([mldsa-native/mldsa/src/sign.c#L1624-L1629](https://github.com/pq-code-package/mldsa-native/blob/7750d741d57c0a459f59dae6dd961309f5f70a58/mldsa/src/sign.c#L1624-L1629))

fegge:
## Summary
- Reject malformed public API inputs before allocation, copying, or length arithmetic that depends on those inputs.
- Tighten external-mu, context, signature-length, and signed-message overflow checks.
- Add allocator and functional regression coverage for the rejected inputs.

## Validation
- `clang-format -i mldsa/src/sign.c mldsa/src/sign.h mldsa/mldsa_native.h test/src/test_alloc.c test/src/test_mldsa.c`: passed
- `git diff --check -- mldsa/src/sign.c mldsa/src/sign.h mldsa/mldsa_native.h test/src/test_alloc.c test/src/test_mldsa.c`: passed
- `make run_alloc run_func -j4`: passed

Fixes #1212

---

### DIFF
--- a/mldsa/src/sign.c
+++ b/mldsa/src/sign.c
@@ -915,6 +915,14 @@ int mld_sign_signature_internal(uint8_t sig[MLDSA_CRYPTO_BYTES], size_t *siglen,
   uint8_t *rho, *tr, *key, *mu, *rhoprime;
   uint16_t nonce = 0;
   const uint16_t nonce_limit = mld_get_max_signing_attempts();
+
+  if (externalmu && mlen != MLDSA_CRHBYTES)
+  {
+    *siglen = 0;
+    mld_memset(sig, 0, MLDSA_CRYPTO_BYTES);
+    return MLD_ERR_FAIL;
+  }
+
   MLD_ALLOC(seedbuf, uint8_t,
             2 * MLDSA_SEEDBYTES + MLDSA_TRBYTES + 2 * MLDSA_CRHBYTES, context);
   MLD_ALLOC(mat, mld_polymat, 1, context);
@@ -1052,6 +1060,14 @@ int mld_sign_signature(uint8_t sig[MLDSA_CRYPTO_BYTES], size_t *siglen,
 {
   size_t pre_len;
   int ret;
+
+  if (ctxlen > 255 || (ctx == NULL && ctxlen != 0))
+  {
+    *siglen = 0;
+    mld_memset(sig, 0, MLDSA_CRYPTO_BYTES);
+    return MLD_ERR_FAIL;
+  }
+
   MLD_ALLOC(pre, uint8_t, MLD_DOMAIN_SEPARATION_MAX_BYTES, context);
   MLD_ALLOC(rnd, uint8_t, MLDSA_RNDBYTES, context);
 
@@ -1153,6 +1169,12 @@ int mld_sign(uint8_t *sm, size_t *smlen, const uint8_t *m, size_t mlen,
   int ret;
   size_t i;
 
+  if (mlen > SIZE_MAX - MLDSA_CRYPTO_BYTES)
+  {
+    *smlen = 0;
+    return MLD_ERR_FAIL;
+  }
+
   for (i = 0; i < mlen; ++i)
   __loop__(
     assigns(i, object_whole(sm))
@@ -1187,6 +1209,11 @@ int mld_sign_verify_internal(const uint8_t *sig, size_t siglen,
   int ret, cmp;
   unsigned int i;
 
+  if (siglen != MLDSA_CRYPTO_BYTES || (externalmu && mlen != MLDSA_CRHBYTES))
+  {
+    return MLD_ERR_FAIL;
+  }
+
   MLD_ALLOC(buf, uint8_t, (MLDSA_K * MLDSA_POLYW1_PACKEDBYTES), context);
   MLD_ALLOC(mu, uint8_t, MLDSA_CRHBYTES, context);
   MLD_ALLOC(c, uint8_t, MLDSA_CTILDEBYTES, context);
@@ -1201,12 +1228,6 @@ int mld_sign_verify_internal(const uint8_t *sig, size_t siglen,
       cp == NULL || mat == NULL || w1 == NULL || tmp == NULL)
   {
     ret = MLD_ERR_OUT_OF_MEMORY;
-    goto cleanup;
-  }
-
-  if (siglen != MLDSA_CRYPTO_BYTES)
-  {
-    ret = MLD_ERR_FAIL;
     goto cleanup;
   }
 
@@ -1607,7 +1628,7 @@ size_t mld_prepare_domain_separation_prefix(
     uint8_t prefix[MLD_DOMAIN_SEPARATION_MAX_BYTES], const uint8_t *ph,
     size_t phlen, const uint8_t *ctx, size_t ctxlen, int hashalg)
 {
-  if (ctxlen > 255)
+  if (ctxlen > 255 || (ctx == NULL && ctxlen != 0))
   {
     return 0;
   }
--- a/mldsa/src/sign.c
+++ b/mldsa/src/sign.c
@@ -916,13 +916,6 @@ int mld_sign_signature_internal(uint8_t sig[MLDSA_CRYPTO_BYTES], size_t *siglen,
   uint16_t nonce = 0;
   const uint16_t nonce_limit = mld_get_max_signing_attempts();
 
-  if (externalmu && mlen != MLDSA_CRHBYTES)
-  {
-    *siglen = 0;
-    mld_memset(sig, 0, MLDSA_CRYPTO_BYTES);
-    return MLD_ERR_FAIL;
-  }
-
   MLD_ALLOC(seedbuf, uint8_t,
             2 * MLDSA_SEEDBYTES + MLDSA_TRBYTES + 2 * MLDSA_CRHBYTES, context);
   MLD_ALLOC(mat, mld_polymat, 1, context);
@@ -934,6 +927,12 @@ int mld_sign_signature_internal(uint8_t sig[MLDSA_CRYPTO_BYTES], size_t *siglen,
       s2hat == NULL)
   {
     ret = MLD_ERR_OUT_OF_MEMORY;
+    goto cleanup;
+  }
+
+  if (externalmu && mlen != MLDSA_CRHBYTES)
+  {
+    ret = MLD_ERR_FAIL;
     goto cleanup;
   }
 
@@ -1061,19 +1060,18 @@ int mld_sign_signature(uint8_t sig[MLDSA_CRYPTO_BYTES], size_t *siglen,
   size_t pre_len;
   int ret;
 
-  if (ctxlen > 255 || (ctx == NULL && ctxlen != 0))
-  {
-    *siglen = 0;
-    mld_memset(sig, 0, MLDSA_CRYPTO_BYTES);
-    return MLD_ERR_FAIL;
-  }
-
   MLD_ALLOC(pre, uint8_t, MLD_DOMAIN_SEPARATION_MAX_BYTES, context);
   MLD_ALLOC(rnd, uint8_t, MLDSA_RNDBYTES, context);
 
   if (pre == NULL || rnd == NULL)
   {
     ret = MLD_ERR_OUT_OF_MEMORY;
+    goto cleanup;
+  }
+
+  if (ctxlen > 255 || (ctx == NULL && ctxlen != 0))
+  {
+    ret = MLD_ERR_FAIL;
     goto cleanup;
   }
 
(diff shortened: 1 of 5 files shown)


mldsa/src/sign.c
```diff
@@ -1052,6 +1060,14 @@ int mld_sign_signature(uint8_t sig[MLDSA_CRYPTO_BYTES], size_t *siglen,
...
+
+  if (ctxlen > 255 || (ctx == NULL && ctxlen != 0))
+  {
+    *siglen = 0;
+    mld_memset(sig, 0, MLDSA_CRYPTO_BYTES);
+    return MLD_ERR_FAIL;
+  }
+
   MLD_ALLOC(pre, uint8_t, MLD_DOMAIN_SEPARATION_MAX_BYTES, context);
   MLD_ALLOC(rnd, uint8_t, MLDSA_RNDBYTES, context);
```
hanno-becker, line 1064: The check must happen after the allocations for strict C90 compliance. Which has the benefit that we can reuse the cleanup section.

mldsa/src/sign.c
```diff
@@ -1153,6 +1169,12 @@ int mld_sign(uint8_t *sm, size_t *smlen, const uint8_t *m, size_t mlen,
   int ret;
   size_t i;
 
+  if (mlen > SIZE_MAX - MLDSA_CRYPTO_BYTES)
+  {
+    *smlen = 0;
+    return MLD_ERR_FAIL;
+  }
+
```
hanno-becker: I'm not sure about this change.

If a user wants to call us with bad parameters, they can. They can just pass a nonsense (non-NULL) address, and we'll crash.

So, if we do want to conduct input validation, I think it should be input validation for the sake of preventing _accidental_ misuse, not deliberate/adversarial one. The above check doesn't fall under that category, IMO, so I don't think we should add it.

mldsa/src/sign.c
hanno-becker: Same as above: It's for C90 compliance that we need to conduct the declarations/allocations upfront.

mldsa/src/sign.c
```diff
@@ -1607,7 +1628,7 @@ size_t mld_prepare_domain_separation_prefix(
     uint8_t prefix[MLD_DOMAIN_SEPARATION_MAX_BYTES], const uint8_t *ph,
     size_t phlen, const uint8_t *ctx, size_t ctxlen, int hashalg)
 {
-  if (ctxlen > 255)
+  if (ctxlen > 255 || (ctx == NULL && ctxlen != 0))
```
hanno-becker, line 1642: I'd prefer input validation to happen at the top-level API, even if it means a bit of code-duplication.

mldsa/src/sign.h
```diff
@@ -281,7 +281,7 @@ __contract__(
   requires(memory_no_alias(sig, MLDSA_CRYPTO_BYTES))
   requires(memory_no_alias(siglen, sizeof(size_t)))
   requires(memory_no_alias(m, mlen))
-  requires(ctxlen <= MLD_MAX_BUFFER_SIZE)
+  requires(ctxlen <= 255)
```
hanno-becker: It's unclear if this should be changed. There are two valid purposes for the CBMC specs:
1. MUST: Document the conditions under which the function is safe to be called and has well-defined behavior. `ctxlen < = 255` is not part of that.
2. MAY: Document the success conditions for the function. `ctx <= 255` is (now) part of that.

The CBMC spec differs in this regard from the natural language documentation, which should be focused on 2. -- the success case.

I'm thus leaning towards changing `Should be <= 255` to `Must be <= 255` in the dcoumentation, but keep the CBMC contract as-is.

mldsa/src/sign.h
```diff
@@ -342,9 +342,10 @@ __contract__(
  *                     MLDSA_CRYPTO_BYTES + mlen bytes); can be equal to m.
  * @param[out] smlen   Pointer to output length of signed message.
  * @param[in]  m       Pointer to message to be signed.
- * @param      mlen    Length of message.
- * @param[in]  ctx     Pointer to context string.
- * @param      ctxlen  Length of context string.
+ * @param      mlen    Length of message. Must be <= SIZE_MAX -
+ *                     MLDSA_CRYPTO_BYTES.
```
hanno-becker: I'd suggest to drop `Must be <= SIZE_MAX - MLDSA_CRYPTO_BYTES` -- this is already implied by the validity assumption and the fact that `sm` has size `MLDSA_CRYPTO_BYTES + mlen`.

test/src/test_alloc.c
```diff
@@ -336,6 +336,29 @@ void custom_free(test_ctx_t *ctx, void *p, size_t sz, const char *file,
     }                                                                          \
   } while (0)
 
+#define TEST_NO_ALLOC_FAILURE(test_name, call, expected_rc)                  \
```
hanno-becker: This is no longer needed once we reorder the checks past allocation.

hanno-becker (changes requested):
Thank you for initiating a discussion on input validation, @fegge.

So far, we have deliberately not conducted input validation, but instead simply _assumed_ that the user passes correct arguments. As mentioned in the comments, we ultimately _have_ to do that, because true pre-conditions such as pointer-validity cannot be checked. For that reason, I think the only input validation one should consider is that aiming to catch _accidental_ misuse.

I'd be OK with consistently adding `len == 0 || ptr != NULL` checks as an safe-guard against accidental passage of bad length or NULL pointers. If we decide to do so, this should be applied to all top-level APIs and all pointers consistently, perhaps even through some shared macro.

@mkannwischer wdyt

mldsa/src/sign.c
fegge: Ok. I think that makes sense.

fegge:
@hanno-becker I updated the PR to address your immediate comments. I think a shared macro for this type of validation makes sense. Just ping me if you'd rather go down that route.

mldsa/src/sign.c
```diff
@@ -1061,6 +1069,12 @@ int mld_sign_signature(uint8_t sig[MLDSA_CRYPTO_BYTES], size_t *siglen,
     goto cleanup;
   }
 
+  if (ctxlen > 255 || (ctx == NULL && ctxlen != 0))
```
mkannwischer, line 1072: Please remove the `ctx == NULL` check throughout.
We are not checking for NULL pointers anywhere else in the code.
I'll separately open a PR to document the convention to not check for NULL.

mkannwischer (changes requested):
Apologies for the delay @fegge! Could you please rebase this on top of main?
There have been some API changes.

mldsa/src/sign.h
```diff
@@ -371,7 +371,7 @@ __contract__(
   requires(memory_no_alias(smlen, sizeof(size_t)))
   requires(m == sm || memory_no_alias(m, mlen))
   requires(ctxlen <= MLD_MAX_BUFFER_SIZE)
-  requires(memory_no_alias(ctx, ctxlen))
+  requires(ctxlen == 0 || memory_no_alias(ctx, ctxlen))
```
mkannwischer, line 374: I am not sure if this needed. My understanding was that memory_no_alias() does permit NULL if the length is zero bytes.
@hanno-becker?

mkannwischer (changes requested):

mkannwischer:
I believe this PR is mostly made obsolete by #1236, #1240, #1241, #1295. Removing from v1 for now.
@fegge, could you check if any change should still be made and either close this PR or rebase it? Thanks!

fegge:
I agree @mkannwischer. Closing this.

PR #1232, opened 2026-06-26
Fail HOL-Light tests on unsupported arch

### LINKED ISSUE #1214: `scripts/tests hol_light` silently succeeds on unsupported host architectures
`scripts/tests hol_light` selects the proof architecture from
`platform.machine()` at `scripts/tests:1031` through `scripts/tests:1034`. If
the host is not `arm64`, `aarch64`, or `x86_64`, the function returns at
`scripts/tests:1035` through `scripts/tests:1036` without printing a diagnostic,
listing proofs, or calling `check_fail()`.

The proof of concept monkeypatches `platform.machine()` to return `riscv64` and
executes the real `scripts/tests hol_light` entry point. The command exits zero
with no output.

Run:

```sh
bash ptp/pocs/poc-hol-light-unsupported-arch-exits-zero.sh
```

Observed output:

```text
run_rc=0
stdout_bytes=0
stderr_bytes=0
VULNERABLE: HOL-Light tests silently passed on unsupported arch
```

fegge:
## Summary
- Make the HOL-Light test command fail on unsupported host architectures instead of reporting success.
- Keep supported architecture mapping explicit before selecting proof functions.

## Validation
- `ruff format scripts/tests`: passed
- `ruff check scripts/tests`: passed
- `python3 -m py_compile scripts/tests`: passed
- `git diff --check -- scripts/tests`: passed
- `python3 scripts/tests hol_light --list-functions`: passed
- Unsupported-architecture probe with `platform.machine()` mocked to `riscv64`: failed as expected with `HOL-Light unsupported architecture: riscv64`

Fixes #1214

---

### DIFF
--- a/scripts/tests
+++ b/scripts/tests
@@ -1049,12 +1049,14 @@ class Tests:
         self.check_fail()
 
     def hol_light(self):
-        if platform.machine().lower() in ["arm64", "aarch64"]:
+        machine = platform.machine().lower()
+        if machine in ["arm64", "aarch64"]:
             arch = "aarch64"
-        elif platform.machine().lower() in ["x86_64"]:
+        elif machine in ["x86_64"]:
             arch = "x86_64"
         else:
-            return
+            self.fail(f"HOL-Light unsupported architecture: {machine}")
+            self.check_fail()
 
         def list_proofs(arch):
             cmd_str = ["./proofs/hol_light/" + arch + "/list_proofs.sh"]


scripts/tests
```diff
@@ -1049,12 +1049,14 @@ class Tests:
...
+        machine = platform.machine().lower()
+        if machine in ["arm64", "aarch64"]:
             arch = "aarch64"
-        elif platform.machine().lower() in ["x86_64"]:
+        elif machine in ["x86_64"]:
             arch = "x86_64"
         else:
-            return
+            self.fail(f"HOL-Light unsupported architecture: {machine}")
+            self.check_fail()
```
hanno-becker, line 1059: Pre-existing: Looking at the script, it seems that we almost always call `check_fail` at the end of each individual test. Why don't we just call `check_fail()` once at the very end of `main`?

hanno-becker (approved):
I wonder if we should take the opportunity to hoist `check_fail` to the bottom of the script instead invoking it in many disparate places. But this is a pre-existing issue and need not necessarily be addressed here.

scripts/tests
mkannwischer: Please open an issue.

mkannwischer (approved):

PR #1235, opened 2026-06-26
Support randomized Wycheproof signing vectors

### LINKED ISSUE #1219: Wycheproof randomized signing vectors are not covered
The Wycheproof client pins C2SP/Wycheproof to commit
`4f5e05f71e6b724c20e2c1b6934c7bd7ef6d89e7`, which predates the upstream
randomized ML-DSA signing vectors merged on June 6, 2026. The newer vectors add
valid signing cases with a nonzero `rnd` field, but the local client never
passes that field to the C driver.

The C driver hard-codes signing randomness to all zeroes in the deterministic
seed and no-seed signing paths. As a result, forcing a current randomized vector
through the client produces a signature mismatch. The default `run_wycheproof`
target therefore reports success without covering the current randomized signing
cases.

The proof of concept downloads the pinned current Wycheproof vector file
`mldsa_44_sign_seed_test.json` from commit
`6d7cccd0fcb1917368579adeeac10fe802f1b521`, confirms that it contains a
`Randomized` valid test case with nonzero `rnd`, builds the real
`wycheproof_mldsa44` driver, and runs the real client against that vector. The
client fails on `tcId=90` because the generated signature uses zero randomness
instead of the vector's `rnd` value.

Run:

```sh
bash ptp/pocs/poc-wycheproof-randomized-signing-vectors-uncovered.sh
```

Observed output:

```text
randomized_tc=90
randomized_rnd_prefix=60879ebd4f33a5d8
randomized_result=valid
client_rc=1
signature_mismatch_tc90=1
VULNERABLE: Wycheproof randomized signing vector is not supported by the client
```

fegge:
## Summary
- Update the pinned Wycheproof commit to include randomized ML-DSA signing vectors.
- Refresh cached vector files when the pinned commit changes.
- Pass optional vector-provided `rnd` values to the local signing driver and validate malformed `rnd` inputs.

## Validation
- `clang-format -i test/wycheproof/wycheproof_mldsa.c`: passed
- `git diff --check -- test/wycheproof/wycheproof_client.py test/wycheproof/wycheproof_mldsa.c`: passed
- `make run_wycheproof -j4`: passed, including refreshed randomized signing vectors such as ML-DSA-44 `tcId=90`
- Direct driver probes for valid `rnd` and malformed `rnd` across `sigGenSeedDeterministic`, `sigGenDeterministic`, and `sigGenInternalDeterministic`: passed

Fixes #1219

---

### DIFF
--- a/test/wycheproof/wycheproof_client.py
+++ b/test/wycheproof/wycheproof_client.py
@@ -20,8 +20,8 @@ from pathlib import Path
 exec_prefix = os.environ.get("EXEC_WRAPPER", "")
 exec_prefix = exec_prefix.split(" ") if exec_prefix != "" else []
 
-# Pinned to a specific commit (2026-06-04).
-WYCHEPROOF_COMMIT = "4f5e05f71e6b724c20e2c1b6934c7bd7ef6d89e7"
+# Pinned to a specific commit (2026-06-06).
+WYCHEPROOF_COMMIT = "6d7cccd0fcb1917368579adeeac10fe802f1b521"
 WYCHEPROOF_BASE_URL = (
     "https://api.github.com/repos/C2SP/wycheproof/contents/testvectors_v1"
 )
@@ -88,6 +88,10 @@ def download_file(url, dest, token):
             time.sleep(wait)
 
 
+def get_wycheproof_data_dir(data_dir):
+    return Path(data_dir) / WYCHEPROOF_COMMIT
+
+
 def download_wycheproof_files(data_dir):
     """Download Wycheproof test vector files if not present."""
     data_dir = Path(data_dir)
@@ -106,7 +110,8 @@ def download_wycheproof_files(data_dir):
                     json.load(f)
             except (json.JSONDecodeError, Exception) as e:
                 print(f"Error downloading {filename}: {e}", file=sys.stderr)
-                local_file.unlink(missing_ok=True)
+                if local_file.exists():
+                    local_file.unlink()
                 return False
     return True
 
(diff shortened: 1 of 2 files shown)


test/wycheproof/wycheproof_client.py
```diff
@@ -53,6 +53,14 @@ def download_wycheproof_files(data_dir):
...
     data_dir = Path(data_dir)
     data_dir.mkdir(parents=True, exist_ok=True)
+    commit_file = data_dir / ".wycheproof-commit"
+
+    if (
+        not commit_file.exists()
+        or commit_file.read_text(encoding="utf-8").strip() != WYCHEPROOF_COMMIT
+    ):
+        for filename in WYCHEPROOF_FILES:
+            (data_dir / filename).unlink(missing_ok=True)
```
mkannwischer: This is a very good catch and it should be ported to mlkem-native

mkannwischer: I have opened https://github.com/pq-code-package/mlkem-native/issues/1764 so we don't forget

test/wycheproof/wycheproof_client.py
```diff
@@ -63,6 +63,14 @@ def download_wycheproof_files(data_dir):
...
```
mkannwischer: `missing_ok` got introduced in Python 3.8 - some of our CI tests run an older Python.
Maybe a simpler and more robust way is to use the same approach used by the ACVP tests: Just make the version part of the file path, i.e., cache files under `.wycheproof-data/{commit}/`.

mkannwischer (changes requested):
Thanks @fegge - this is a nice addition and the core looks correct to me.
Can you please get it through CI?

test/wycheproof/wycheproof_client.py
fegge: Sounds good. I'll go ahead and make that update.

mkannwischer (changes requested):
Thanks @fegge.

The changes look good to me now! Could you kindly combine your two commits? We are using `Rebase and merge` only and would like to keep the history cleaned up.

mkannwischer (approved):
I took the liberty to rebase your changes on top of main and squash your second commit into the first one. 

Thanks again for this improvement!

PR #1236, opened 2026-06-28
Remove SUPERCOP signed-message API (crypto_sign/crypto_sign_open)

mkannwischer:
crypto_sign() and crypto_sign_open(), and the underlying mld_sign / mld_sign_open, produce and consume the SUPERCOP signed-message format (signature || message).
This commit removes those APIs, as presently no consumer is using them and they can be rebuilt on top of the other APIs straightforwardly.


hanno-becker (approved):
I support this removal. The code adds complexity while not serving any known use case, so there's no point keeping it.

PR #1237, opened 2026-06-28
Add signing hooks for observing signature attempts and implementing restartable signing

hanno-becker:
* Based on #1238

```
sign: Add configurable signing hooks (resume/attempt/finish)

The runtime of ML-DSA signatures is theoretically unbounded. Even the
bound of max 814 signature attempts -- approved by NIST as lowering the
chance of signing failure below 2^{-256} -- leaves a performance
distribution whose peak percentiles can be unacceptable for constrained
systems with strict timing requirements.

The library already provides MLD_CONFIG_MAX_SIGNING_ATTEMPTS to cap the
number of signing attempts. However, FIPS 204 forbids lowering that bound
below 814, rendering the option non-compliant in constrained environments.

As a remedy, this commit introduces three optional 'signing hooks' around
the core rejection-sampling loop, keyed by the attempt counter (the loop
iteration; the spec counter is kappa = attempt*MLDSA_L):

- `attempt = sign_resume()`: Return the attempt to resume the operation
  from. `0` for an initial / non-restartable signing operation.
- `sign_attempt(attempt)`: Called before each attempt. Returns 0 to
  proceed, or non-zero to pause: signing then returns MLD_ERR_SIGNING_PAUSED
  with `attempt` as the resume point.
- `sign_finish(attempt)`: Called on success with the succeeding attempt.

Together, resume + attempt make signing _restartable_: a caller can pause
after a bounded number of attempts and later resume from the recorded
attempt, bounding per-call runtime while keeping a FIPS-compliant overall
bound and reproducing the same signature as an uninterrupted run.

This only works if the consumer drives the deterministic internal API: the
randomized API would choose a fresh seed on each (re-)start, so the resume
hook requires MLD_CONFIG_NO_RANDOMIZED_API and is incompatible with
MLD_CONFIG_KEYGEN_PCT (whose one-shot internal signature cannot be resumed).
The attempt and finish hooks also serve standalone uses with any API, e.g.
logging or benchmarking the per-signature attempt distribution.

Pausing is independent of MLD_CONFIG_MAX_SIGNING_ATTEMPTS, the upper bound on
the total number of attempts (>= 814 for FIPS 204 compliance) after which
signing gives up with the distinct error MLD_ERR_SIGN_ATTEMPTS_EXHAUSTED; it
must not be lowered as a restart mechanism.

sign_{resume|attempt|finish} can be implemented on mutable globals or on a
context, so they are introduced under the existing MLD_CONFIG_CONTEXT_PARAMETER
wrapper, which lets functions be called with or without an
application-specific context pointer.

Concretely, this commit:

- Adds independent MLD_CONFIG_SIGN_HOOK_{RESUME,ATTEMPT,FINISH} options. When
  set, the integration provides, respectively:

  - `attempt = mld_sign_hook_resume([ctx])`: provide initial/resume attempt
  - `mld_sign_hook_attempt(attempt [, ctx])`: per-attempt hook; pause via
    non-zero return
  - `mld_sign_hook_finish(attempt [, ctx])`: note the succeeding attempt

  The unset dummies leave the default build (and its proofs) unchanged.

- Adds the MLD_ERR_SIGNING_PAUSED error code returned when the attempt hook
  pauses signing.

- Adds CBMC proofs for the three hooks and threads them through the existing
  signing proofs.

- Adds test/src/test_sign_hook.c (a new `sign_hook` test), which checks that
  restartable signing -- single-step and with random attempts-per-call --
  reproduces one-shot signing exactly, and gathers an attempts-per-signature
  distribution that it checks against the expected geometric mean per
  parameter set.


```


mldsa/src/sign.c
```diff
@@ -997,14 +999,27 @@ int mld_sign_signature_internal(uint8_t sig[MLDSA_CRYPTO_BYTES], size_t *siglen,
...
+    }
+
     ret = mld_attempt_signature_generation(sig, mu, rhoprime, kappa, mat, s1hat,
                                            s2hat, t0hat, context);
     if (ret == 0)
     {
       *siglen = MLDSA_CRYPTO_BYTES;
+      /* Signing succeeded: record the attempt that succeeded. No-op in the
+       * default build. */
+      mld_sign_finish(attempt, context);
```
mkannwischer, line 1027: What is `mld_sign_finish` needed for? It looks like the same information is passed to the last `mld_sign_attempt`

mkannwischer (commented):
Thanks @hanno-becker. I have done a design review for now and have some questions - a full review will follow later. 

Could you elaborate why you went for hooks to implement this approach rather than adding a `_restartable()` API that gets a context argument?

It would be good to add an example of how the new hooks can be used. We'll also need to add documentation.

mldsa/src/sign.c
hanno-becker: It's for ease of use and a symmetric API. For example, if you only care about pairing your benchmarks with information of how many signing attempts were performanced, `finish` gives you that information most directly.

But you are right, in _theory_ one can get away without it. I'd prefer not to.

hanno-becker:
> Could you elaborate why you went for hooks to implement this approach rather than adding a _restartable() API that gets a context argument?

Hooks are more flexible and less work for us. The details of how to build a restartable API on top of them are left to the user -- by design. For example, you'd probably also want randomized signing to be restartable; that's something you can easily build yourself. If, on the other hand, we were to do it ourselves, we'd have to define some kind of restart context, which then clashes with the customer's context and is also hard to keep stable if we introduce, say, structured keys. The hooks are the core mechanism.

And the hooks are useful for statistics and benchmarking too, not just for restartability.

hanno-becker:
> It would be good to add an example of how the new hooks can be used. We'll also need to add documentation.

What documentation do you want, beyond the extensive one in `mldsa_native_config.h`?

~I can add an example.~ Added an example.

mldsa/mldsa_native_config.h
```diff
@@ -684,24 +684,103 @@
...
+ * `struct my_ctx *`.
  *
- * This is only relevant if MLD_CONFIG_CONTEXT_PARAMETER is set.
+ * This option must be defined if and only if MLD_CONFIG_CONTEXT_PARAMETER is
+ * defined; defining one without the other is a compile-time error.
  */
 /* #define MLD_CONFIG_CONTEXT_PARAMETER_TYPE void* */
 
+/**
+ * Signing hooks: MLD_CONFIG_SIGN_HOOK_RESUME / _ATTEMPT / _FINISH
```
mkannwischer, line 727: Shall we mark the feature as experimental allowing us to still tweak it after v1 is out?

mldsa/src/params.h
```diff
@@ -70,6 +70,12 @@
 #define MLDSA_POLYT0_PACKEDBYTES 416
 #define MLDSA_POLYVECH_PACKEDBYTES (MLDSA_OMEGA + MLDSA_K)
 
+/* Sampling y from counter kappa uses nonces kappa, ..., kappa+L-1, which fit in
+ * uint16_t iff kappa <= UINT16_MAX - MLDSA_L. With kappa = attempt*MLDSA_L this
+ * bounds the number of signing attempts by MLD_MAX_KAPPA / MLDSA_L; see
+ * MLD_MAX_SIGNING_ATTEMPTS in sign.c. */
+#define MLD_MAX_KAPPA (UINT16_MAX - MLDSA_L)
```
mkannwischer, line 77: We need to include stdint.h ... this only works out of luck. 
(but the `int32_t` casts would fail if we would be out of luck)

test/src/test_sign_hook.c
```diff
@@ -0,0 +1,487 @@
...
+           MLD_SIGN_HOOK_EXPECTED_RATIO_X100 / 100,
+           MLD_SIGN_HOOK_EXPECTED_RATIO_X100 % 100);
+
+    /* With enough samples, check that ratio against the expected mean. All
+     * integer: compare 100*attempts against EXPECTED_RATIO_X100*signatures,
+     * allowing a +-TOL_PCT% relative band. */
+    if (iterations >= MLD_SIGN_HOOK_MIN_CHECK_N)
+    {
+      uint64_t lhs = 100u * a;
+      uint64_t mid = (uint64_t)MLD_SIGN_HOOK_EXPECTED_RATIO_X100 * s;
```
mkannwischer: This never gets executed in CI if I see correctly. Should we add a separate job for it?

mldsa/src/context.h
```diff
@@ -60,6 +60,15 @@
...
 #endif /* !MLD_CONFIG_CONTEXT_PARAMETER */
 
+/* Consume a context parameter carried only for the integration's benefit,
+ * avoiding -Wunused-parameter; expands to nothing when no context is
+ * configured. */
+#if defined(MLD_CONFIG_CONTEXT_PARAMETER)
+#define MLD_CONTEXT_UNUSED(context) ((void)(context))
+#else
+#define MLD_CONTEXT_UNUSED(context) ((void)0)
+#endif
```
mkannwischer, line 75: This could be ported to mlkem-native.

mldsa/src/sign.c
```diff
@@ -586,9 +583,11 @@ __contract__(
...
 }
 
-/* User-facing bound on signing attempts. See MLD_CONFIG_MAX_SIGNING_ATTEMPTS
- * in mldsa_native_config.h. Default is chosen so that failure probability
- * is < 2^{-256}, that is, signatures will practically always succeed. */
+/* Effective bound on signing attempts: the configured bound
+ * MLD_CONFIG_MAX_SIGNING_ATTEMPTS (see mldsa_native_config.h) if set, otherwise
+ * the hard type-safety bound MLD_MAX_KAPPA / MLDSA_L (see MLD_MAX_KAPPA in
+ * params.h). The default is chosen so that the failure probability is
+ * < 2^{-256}, that is, signatures will practically always succeed. */
```
mkannwischer: This may read confusing to the reader.
If the default is chosen such that the failure probability is < 2^{-256}, you would use the NIST bound of 814.

mldsa/mldsa_native_config.h
```diff
@@ -702,6 +702,64 @@
...
+ *   point (needs MLD_CONFIG_SIGN_HOOK_RESUME to resume; otherwise just aborts).
+ *   Always returning 0 makes it a logging/benchmarking hook.
+ *
+ * - MLD_CONFIG_SIGN_HOOK_FINISH: void mld_sign_hook_finish(attempt[, ctxt])
+ *   Called on success with the succeeding attempt. Observe-only.
+ *
+ * When an option is unset, the hook is a no-op (resume to 0, attempt proceeds),
+ * i.e. ordinary one-shot signing.
+ *
+ * Independent of MLD_CONFIG_MAX_SIGNING_ATTEMPTS (the hard attempt bound, >=
```
mkannwischer: Do we still need `MLD_CONFIG_MAX_SIGNING_ATTEMPTS`? May be worth removing to reduce complexity for v1.

mldsa/src/sign.h
mkannwischer, line 1: `keypair` and `keypair_internal` can now also return `MLD_ERR_SIGNING_PAUSED` in case `PCT` and `HOOK_ATTEMPT` is configured. That should be documented.

mkannwischer (changes requested):
Thanks @hanno-becker for all the work!
Found a few small issues - otherwise I'm very happy with this feature! 
Zephyr CI is having some issues, could you fix that?

> What documentation do you want, beyond the extensive one in mldsa_native_config.h?

Maybe a paragraph in the root README like we have for `DMLD_CONFIG_REDUCE_RAM`? We may want to revisit the README before v1 anyway though.

mldsa/src/sign.h
hanno-becker: No, that's not allowed -- there would be no way to restart the PCT. The combination of hooks and PCT is checked for.

mldsa/mldsa_native_config.h
hanno-becker: One could emulate this using an attempt-hook, if we gave the attempt-hook the possibility to inject a _failure_, not just a _pause_. But I think the added complexity of this approach -- both on our side, and on the user's side -- outweighs the benefit of removing this option.

One could consider defaulting to 814, but that's slightly odd seeing FIPS204 says one "SHOULD NOT" do this.

Setting this to 814 is quite natural, IMO, to improve observability in case of a faulty implementation looping; esp. on low-end systems.

Altogether, I prefer leaving this option.

mldsa/src/sign.c
hanno-becker: Any bound >= 814 has that probability. I deliberately omitted the specific value here because I felt it's irrelevant to the user. But point taken, if you're confused, so will be others. I'll reconsider.

mldsa/src/context.h
hanno-becker: Yes

mldsa/mldsa_native_config.h
mkannwischer: I'm not in favor of defaulting to 814 due to the `SHOULD NOT`.
Still I don't see why _any_ consumer would set `MLD_CONFIG_MAX_SIGNING_ATTEMPTS` to anything other than 814 or the max value. 

What do you think about making it a binary option to use 814 instead of the default?

mldsa/src/sign.h
mkannwischer: Can you please double check this? The way I read the config (+ config validation) is that `PCT` is incompatible with `RESUME`, but not with `ATTEMPT`. Am I getting something wrong here?

mldsa/src/context.h
mkannwischer: I'll probably forget this, so I opened an issue: https://github.com/pq-code-package/mlkem-native/issues/1795

mldsa/mldsa_native_config.h
hanno-becker: I'm in favor of keeping it as it is. It gives the consumer more flexibility, if only for testing purposes, without making things any more difficult on our end.

mldsa/src/sign.h
hanno-becker: You spotted a real issue there, thank you. Initially, `attempt` couldn't signal pause (pause was handled via MAX_SIGNING_ATTEMPTS), in which case it was safe to use with PCT. Now, it is indeed ATTEMPT which cannot be used with PCT, while RESUME and FINISH, in theory, could.

mldsa/src/sign.c
hanno-becker: I'll just drop "The default is chosen so that the failure probability is * < 2^{-256}, that is, signatures will practically always succeed." here.

test/src/test_sign_hook.c
hanno-becker: This is a good point. I'll switch to 2000 iterations by default at a 10% interval (according to Claude, covering >99.999% of the cases), as well as a single 100k iteration run with a narrower band.

mldsa/src/params.h
hanno-becker: This doesn't really matter because this is lazily evaluated. Also cf. `MLDSA_GAMMA1` from https://github.com/pq-code-package/mldsa-native/commit/50a5796c5d80bb3b1456621aa0f298fa500cf0f0.

mldsa/mldsa_native_config.h
hanno-becker: Ok

hanno-becker:
@mkannwischer Thank you for the review!

I changed the distribution test so that a) by default, 2000 iterations are run and a wide error margin allowed, b) a dedicated CI job exercises a larger number of iterations and accordingly checks for a narrower error.

mkannwischer (approved):
Thanks @hanno-becker for adding this sizeable feature. This looks great to me now.
I appreciate the example!

PR #1238, opened 2026-06-29
common: Split context-parameter machinery into context.h

hanno-becker:
Move the MLD_CONTEXT_PARAMETERS_n macros and the associated configuration check out of common.h into a new, level-generic header context.h, included by common.h at the point the material previously sat. No behavioral change.


mkannwischer (approved):
Thanks @hanno-becker. LGTM.

PR #1240, opened 2026-06-29
Remove siglen from signing and verification APIs

### LINKED ISSUE #789: `siglen` parameter in `crypto_sign_signature` is prone to API misuse

mkannwischer:
I would like initiate the discussion on the following breaking change to our API. Please share your thoughts if this change should be made prior to releasing v1.0.0. I have updated the AWS-LC, Pavona, and liboqs integrations (see the patches).

ML-DSA signatures have a fixed size, so siglen was redundant: an always-MLDSA_BYTES output for signing and an input used only for a fixed-size check on verification. Drop it from all signing and verification functions (internal, randomized, extmu, and pre-hash variants). Verification now takes the signature as a fixed-size array.

- Resolves #789

### DIFF
--- a/integration/aws-lc/post_import.patch
+++ b/integration/aws-lc/post_import.patch
@@ -1,5 +1,12 @@
 # Copyright (c) The mldsa-native project authors
 # SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+#
+# Adapt AWS-LC's ML-DSA wrapper to the mldsa-native API, where signing and
+# verification omit the signature-length argument (ML-DSA signatures have a
+# fixed length). For signing, the wrapper sets the output length itself; for
+# verification, it validates the input length before calling, since
+# mldsa-native expects a fixed-length signature.
+
 diff --git a/crypto/fipsmodule/ml_dsa/mldsa_native_config.h b/crypto/fipsmodule/ml_dsa/mldsa_native_config.h
 --- a/crypto/fipsmodule/ml_dsa/mldsa_native_config.h
 +++ b/crypto/fipsmodule/ml_dsa/mldsa_native_config.h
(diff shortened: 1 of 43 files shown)


mkannwischer:
@jakemas @bhess @rod-chapman @hanno-becker, WDYT about this proposed change?
My view is that it's a useful simplification for most future consumers, and the required changes for existing  consumers are acceptable.

integration/liboqs/ML-DSA-44_META.yml
```diff
@@ -26,36 +26,36 @@ implementations:
   folder_name: .
   compile_opts: -DMLD_CONFIG_PARAMETER_SET=44 -DMLD_CONFIG_FILE="../../integration/liboqs/config_c.h"
   signature_keypair: PQCP_MLDSA_NATIVE_MLDSA44_C_keypair
-  signature_signature: PQCP_MLDSA_NATIVE_MLDSA44_C_signature
-  signature_verify: PQCP_MLDSA_NATIVE_MLDSA44_C_verify
+  signature_signature: PQCP_MLDSA_NATIVE_MLDSA44_C_signature_oqs
+  signature_verify: PQCP_MLDSA_NATIVE_MLDSA44_C_verify_oqs
   signature_signature_extmu: PQCP_MLDSA_NATIVE_MLDSA44_C_signature_extmu
   signature_verify_extmu: PQCP_MLDSA_NATIVE_MLDSA44_C_verify_extmu
```
mkannwischer: We may want to wrap the `signature_extmu` and `verify_extmu` functions as well - but they are currently unused by oqs anyway.

mkannwischer: I added it for completeness.

integration/liboqs/sig_glue.c
```diff
@@ -0,0 +1,81 @@
...
+ * fixed-length one, and are referenced from the per-level META.yml files.
+ *
+ * The shims are compiled once per build (parameter set and backend), so the
+ * MLD_NAMESPACE_KL-derived symbol names match the names liboqs links against.
+ */
+
+#include <stddef.h>
+#include <stdint.h>
+
+#include "mldsa/src/sign.h"
```
hanno-becker: Why is this working with `sign.h` and not with `mldsa_native.h`?

mldsa/mldsa_native.h
```diff
@@ -229,13 +229,12 @@ int MLD_API_NAMESPACE(keypair)(
 /**
  * Compute signature using a caller-supplied random seed and prefix.
  *
- * If the returned value is non-zero, then the values of *sig and *siglen
+ * If the returned value is non-zero, then the value of *sig
  * should not be referenced.
  *
  * @spec{Implements @[FIPS204, Algorithm 7, ML-DSA.Sign_internal].}
  *
  * @param[out] sig        Output signature.
```
hanno-becker: The expected length should be specified here. While at it, the text should also be aligned with the other parameters, e.g. `Pointer to buffer to hold the generated signature.`

mldsa/mldsa_native.h
```diff
@@ -368,7 +365,6 @@ int MLD_API_NAMESPACE(signature_extmu)(
  * @spec{Implements @[FIPS204, Algorithm 8, ML-DSA.Verify_internal].}
  *
  * @param[in] sig        Pointer to input signature.
```
hanno-becker: Similar here, the length should be documented not only in the signature, but also in the documentation.

hanno-becker (changes requested):
Overall I support this change. Two questions/requests:
1. Ideally, no consumer should work with `sign.h`, because this is internal. They should be working with the public header `mldsa_native.h` only. Is there a reason why that's not possible?
2. Now that the siglen parameter is dropped, the buffer length assumption should be documented.

integration/liboqs/sig_glue.c
mkannwischer: OQS doesn't include any headers at all. I used sign.h here because it's a little easier to use and arguably `sig_glue.c` itself is internal as well.
But we can switch to mldsa_native.h without much complication.
Pushed.

mldsa/mldsa_native.h
mkannwischer: Added

mldsa/mldsa_native.h
mkannwischer: Added

mldsa/src/sign.h
```diff
@@ -175,13 +175,12 @@ __contract__(
 /**
  * Compute signature using internal randomness.
  *
- * If the returned value is non-zero, then the values of *sig and *siglen
+ * If the returned value is non-zero, then the value of *sig
  * should not be referenced.
```
hanno-becker: Pre-existing: This is a bit cryptic. What does "should not be referenced" mean?

We should either drop the line and say nothing, or specify that the sig buffer will be zeroized upon error.

mldsa/src/sign.h
```diff
@@ -480,7 +458,6 @@ __contract__(
  * a stable API use mld_sign_signature_pre_hash_shake256.
  *
  * @param[out] sig     Output signature.
```
hanno-becker, line 479: While at it, unify the language and add "Pointer to ...". Similar in the other APIs.

integration/aws-lc/post_import.patch
```diff
@@ -12,3 +19,288 @@ diff --git a/crypto/fipsmodule/ml_dsa/mldsa_native_config.h b/crypto/fipsmodule/
...
++  }
+ 
+   FIPS_service_indicator_unlock_state();
+   if (ret == 0) {
+@@ -386,7 +409,10 @@ int ml_dsa_extmu_65_verify(const uint8_t *public_key /* IN */,
+ 
+   // mu_len is ignored - extmu always uses MLDSA_CRHBYTES (64 bytes)
+   (void)mu_len;
+-  int ret = mldsa65_verify_extmu(sig, sig_len, mu, public_key);
++  int ret = -1;
```
hanno-becker, line 181: There's a switch from a named error (previously returned from `verify`) to a hardcoded `-1` here. But it doesn't matter seeing that only `== 0` vs. `!= 0` ultimately matters for the behavior of the function.

hanno-becker (approved):
This is a welcome simplification, thank you @mkannwischer.

I left some observations for how we can improve documentation at the same time, but they are not blocking.

mldsa/src/sign.h
mkannwischer: Clarified it. We also need to consolidate the zeroization in other places, see https://github.com/pq-code-package/mldsa-native/pull/1242

mldsa/src/sign.h
mkannwischer: done

jakemas:
> @jakemas @bhess @rod-chapman @hanno-becker, WDYT about this proposed change? My view is that it's a useful simplification for most future consumers, and the required changes for existing consumers are acceptable.

While I can’t argue with the simplification of the API, I do want to put down some thoughts. It kind of goes against aws-LC design guidelines. I believe, across awslc we intend to include len (and length checks) to buffers such as this, so I’m not sure the change is positive for aws-LC. Can’t currently check, but IIRC it’s part of the EVP API so we are always going to have those values as call time. We can of course do anything outside of the mldsa API and within EVP itself, but these are consequences of this change. 

I do recall some parts in FIPS204 that says about length checks are a requirement (I’m not sure that extends as far as an API requirement, but perhaps worth a check over FIPS 204 on exact language here). 

See perhaps Section 3.6.2:

>If an implementation can accept inputs for σ or pk of any other length, it shall return false whenever the lengths … differ from their lengths specified in this standard.

siglen allows the caller to define this exactly, and when accepting buffers from potentially untrusted sources, not a bad call for API security in depth. Are we making the API more or less misuse resistant here, are realistic FIPS callers going to be making this check anyway, do we like the control the caller has to enforce len at call time. Would we have to add a caveat that to have FIPS204 one would need to validate the sig length before calling our API? Does this make usage actually more difficult and confusing?

jakemas (dismissed):
Discussion of comment above of potential downstream consequences and FIPS considerations before my approval.

hanno-becker:
Many thanks @jakemas for chiming in!

I think you raise very good points. For higher level APIs such as provided by AWS-LC, it might make sense to always pair buffers with their lengths, and include lengths-checks as required.

If we look at the lower-level API of mlkem/mldsa-native, however, this is not the case: Where buffers are statically sized by the spec, we pass them as statically sized buffers. Good examples are `pk`, `sk`, `ss`, `ct`. Dropping `siglen` and working with fixed-size signature buffers seems to add consistency to the mlx-native API design.

I believe there is no right or wrong here: Each library has an API style to choose and a suitable error checking regime to implement. If you pass buffers + lengths, and the length is subject to constraints, you may want to check them. If you merely pass buffers, you may or may not want to include at least rudimentary `!= NULL` checks -- this is actually still an open question for the mlx-native API, as we currently don't do this. If you decide not to include those checks, I'd argue it's not a bug, it's a choice: So long as you are clear about the assumptions of an API (in our case formally reflected in the CBMC preconditions), it's OK to drop checks in the implementations.

Generally, the higher-level your API gets, the less you want to assume and the more checks you want to include. So, at the level of the EVP API it makes sense to check everything, IMO. But equally I believe that at the lowest level we are looking at here, not passing redundant information, and limiting parameter validation checks, is a defensible choice, and it remains my preference.

Note that there is no need to demand that in stack of APIs, the APIs all follow the same choice -- much the contrary. As you move from one layer to another, you'll discharge the conditions that the higher layer does not assume, but the lower layer does. We see this in @mkannwischer's patch to the LC integration layer: The siglen checks are still present in LC, they have just been hoisted out. There has not been a change in behavior for LC, overall.

Overall, I believe that this is a good change which removes an outlier to what is otherwise a consistent low-level API-style pursued in mlx-native. But Jake is right that lengths checks definitely have a place and the higher levels, and we need to make sure there's no change in behavior for our integrations -- which is the case, as far as I see.

Thanks again @jakemas for joining in and let us know further thoughts 🙏

mkannwischer:
Thanks @jakemas for taking a look at this change and the impact on AWS-LC. 
@hanno-becker's comment captures my motivation for this change well.
In my view, the removal of the siglen unifies our API with how we treat all other input buffers. Length checking should happen on a higher level.

@jakemas, I'm assuming your 👍 means you agree with the change. If not, could you please open a new issue so we can continue the discussion there.

PR #1242, opened 2026-06-29
Clear ML-DSA outputs on signing failure

tob-joe:
## Summary

- clear caller-visible ML-DSA keypair outputs when key generation fails
- clear signature and signed-message outputs, and reset output lengths, when
  signing fails
- add RNG/allocation/PCT failure regression test coverage for stale output
  buffers
- keep verify-only reduced-api allocation and RNG-failure tests buildable by
  compiling the zeroization helper only when keypair or signing tests are
  enabled

## Notes

The reduced-api verify-only configuration disables both keypair and sign APIs.
Those verify tests intentionally keep no-op output checks because verification
does not expose a caller-owned output buffer that needs zeroing. Without the
API guard, the shared zeroization helper was compiled but unused in that mode,
which fails under `-Werror=unused-function`.

## Testing

- `make run_rng_fail`
- `make run_alloc`
- `make run_unit`
- `make run_func`
- focused reduced-api alloc/rng-fail with ASan/UBSan flags
- focused reduced-api-reduce-ram alloc/rng-fail with ASan/UBSan flags
- `git diff --check`

This work was completed by Trail of Bits as part of the Patch The Planet
project in collaboration with OpenAI. The issue was identified primarily by
the Codex coding agent, and manually reviewed before submission.

### DIFF
--- a/mldsa/src/sign.c
+++ b/mldsa/src/sign.c
@@ -391,14 +391,21 @@ cleanup:
   MLD_FREE(inbuf, uint8_t, MLDSA_SEEDBYTES + 2, context);
   MLD_FREE(seedbuf, uint8_t, 2 * MLDSA_SEEDBYTES + MLDSA_CRHBYTES, context);
 
-  if (ret != 0)
-  {
-    return ret;
-  }
-
   /* Pairwise Consistency Test (PCT) @[FIPS140_3_IG, p.87] */
   /* Do this after freeing all temporaries. */
-  return mld_check_pct(pk, sk, context);
+  if (ret == 0)
+  {
+    ret = mld_check_pct(pk, sk, context);
+  }
+
+  if (ret != 0)
+  {
+    /* Clear caller outputs on failure. */
+    mld_zeroize(pk, MLDSA_CRYPTO_PUBLICKEYBYTES);
+    mld_zeroize(sk, MLDSA_CRYPTO_SECRETKEYBYTES);
+  }
+
+  return ret;
 }
 
 #if !defined(MLD_CONFIG_CORE_API_ONLY)
--- a/mldsa/src/sign.c
+++ b/mldsa/src/sign.c
@@ -1050,7 +1050,7 @@ cleanup:
   if (ret != 0)
   {
     /* To be on the safe-side, we zeroize the signature buffer. */
-    mld_memset(sig, 0, MLDSA_CRYPTO_BYTES);
+    mld_zeroize(sig, MLDSA_CRYPTO_BYTES);
   }
 
   /* @[FIPS204, Section 3.6.3] Destruction of intermediate values. */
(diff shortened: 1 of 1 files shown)


mkannwischer (dismissed):
Thanks @tob-joe! Generally, I do support this change of clearing buffers on failures.
I'm not very happy about the added complexity to the `test_alloc.c` and `test_rng_fail.c`.
Can't we just modify the existing tests to init with with fixed values, and check it's zero in the end?
Please also rebase on top of main.

jakemas:
Is there a performance penalty to this? Added benchmark label to see if it will pick up anything. Did you do any locally testing on performance?

mkannwischer:
I took the liberty to rebase this on top of main. 
I have also removed the tests for now (let's get the code change merged first and discuss separately how this is best tested).
Also I have switched from `memset` to `zeroize` which stops the compiler from optimizing it out.
I also moved the zeroize in the keypair to after the PCT.

Now I am happy with this change.
@hanno-becker, you too?

We should still clean up the commit history afterwards.

mkannwischer:
This PR should now adhere to what we document in #1295.

mldsa/src/sign.c
```diff
@@ -429,6 +436,14 @@ int mld_sign_keypair(uint8_t pk[MLDSA_CRYPTO_PUBLICKEYBYTES],
 cleanup:
   /* @[FIPS204, Section 3.6.3] Destruction of intermediate values. */
   MLD_FREE(seed, uint8_t, MLDSA_SEEDBYTES, context);
+
+  if (ret != 0)
+  {
+    /* Clear caller outputs on failure. */
+    mld_zeroize(pk, MLDSA_CRYPTO_PUBLICKEYBYTES);
+    mld_zeroize(sk, MLDSA_CRYPTO_SECRETKEYBYTES);
```
hanno-becker: I don't think this is necessary anymore as per the convention that output is unchanged or zero'ed.

mldsa/src/sign.h
```diff
@@ -91,6 +91,8 @@
  * When MLD_CONFIG_KEYGEN_PCT is set, performs a Pairwise Consistency Test
  * (PCT) as required by FIPS 140-3 IG.
  *
+ * On error (non-zero return value), the output buffers pk and sk are zeroized.
+ *
```
hanno-becker: This is superseded by our recently agreed convention that output is either left unchanged, or zero'ed.

hanno-becker (changes requested):
As per the convention that output is unmodified or zero'ed, we can remove some of the zeroizations, see comments.

mldsa/src/sign.c
mkannwischer: Removed.

mldsa/src/sign.h
mkannwischer: Removed.

mkannwischer:
> As per the convention that output is unmodified or zero'ed, we can remove some of the zeroizations, see comments.

Adjusted. Please take another look.

hanno-becker (approved):

PR #1243, opened 2026-06-29
Make ML-DSA SHAKE finalization idempotent

tob-joe:
## Summary

- make consecutive ML-DSA SHAKE128/SHAKE256 finalize calls harmless before squeezing
- tighten the internal Keccak finalize precondition and use MLD_MAX_BUFFER_SIZE for XOF proof-contract bounds
- add regression test coverage for one-finalize versus two-finalize SHAKE output across boundary absorb lengths

## Rationale

This is narrow misuse resistance for the built-in incremental FIPS202 implementation. The SHAKE helpers are internal, but they have external linkage by default and sit behind helper boundaries where names such as `*_absorb_once` already include finalization before later squeezing. A maintainer adding an explicit finalize at such a boundary should not silently corrupt the state by applying the Keccak delimiter twice.

This does not change the intended absorb/finalize/squeeze lifecycle, does not make absorb-after-finalize valid, and does not claim to cover finalizing again after squeezing has begun.

## Testing

- `make clean run_unit`
- `nix develop --command python3 scripts/autogen --dry-run`

### DIFF
--- a/mldsa/src/fips202/fips202.c
+++ b/mldsa/src/fips202/fips202.c
@@ -111,7 +111,7 @@ __contract__(
 static void keccak_finalize(uint64_t s[MLD_KECCAK_LANES], unsigned int pos,
                             unsigned int r, uint8_t p)
 __contract__(
-  requires(pos <= r && r < sizeof(uint64_t) * MLD_KECCAK_LANES)
+  requires(pos < r && r < sizeof(uint64_t) * MLD_KECCAK_LANES)
   requires((r / 8) >= 1)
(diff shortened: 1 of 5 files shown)


mkannwischer (commented):
Thanks @tob-joe. Could you elaborate what is the motivation for this change?
This is an internal function that we are not exposing to consumers.
I'm not convinced that we need to make internal functions misuse resistant.

mkannwischer:
> Thanks @tob-joe. Could you elaborate what is the motivation for this change? This is an internal function that we are not exposing to consumers. I'm not convinced that we need to make internal functions misuse resistant.

Closing this as I didn't get a response. These are internal functions that shouldn't be used by consumers directly.

PR #1244, opened 2026-06-29
Enforce ML-DSA x86 native assumptions

tob-joe:
# Enforce ML-DSA x86 native assumptions

## Summary

- Add assembler-time guards that fail the AVX2 NTT/INTT assembly build if the qdata table offsets drift away from the hard-coded displacements.
- Strengthen the native API, dev-source declarations, and generated x86 implementation contracts for AVX2 polynomial routines that require 32-byte-aligned polynomial buffers.
- Keep the public native wrapper contracts, implementation declarations, and proof-facing assumptions aligned for `poly_caddq`, `poly_use_hint`, and `polyz_unpack`.

## Why

The x86_64 NTT/INTT assembly uses fixed qdata displacements after simplification. The dev source includes the symbolic layout, but the checked-in simplified assembly has numeric offsets; the new assembler guards make a future qdata layout mismatch fail at build time instead of silently preserving stale displacements.

Several AVX2 polynomial routines also use aligned loads or stores, and the HOL/CBMC proof assumptions already model aligned polynomial buffers. The in-tree typed callers pass `mld_poly` storage, which is declared with the project alignment attribute; this patch makes that precondition explicit at the native API and raw x86 contract boundary.

## Verification

- `./scripts/tests cbmc -kl ALL -p poly_use_hint_native --single-step -j1`
- `./scripts/tests cbmc -kl ALL -p poly_caddq_native_x86_64 --single-step -j1`
- `./scripts/tests cbmc -kl 44 -p polyz_unpack_native_x86_64 --single-step -j1`
- `./scripts/tests cbmc -kl 65 -p polyz_unpack_native_x86_64 --single-step -j1`
- `./scripts/tests cbmc -kl 87 -p polyz_unpack_native_x86_64 --single-step -j1`
- `./scripts/tests cbmc -kl ALL -p polyz_unpack_native --single-step -j1`

### DIFF
--- a/dev/x86_64/src/arith_native_x86_64.h
+++ b/dev/x86_64/src/arith_native_x86_64.h
@@ -137,6 +137,7 @@ void mld_poly_caddq_avx2_asm(int32_t *r)
  * in proofs/hol_light/x86_64/proofs/poly_caddq_avx2_asm.ml */
 __contract__(
   requires(memory_no_alias(r, sizeof(int32_t) * MLDSA_N))
+  requires(((uintptr_t)r % MLD_DEFAULT_ALIGN) == 0)
   requires(array_abs_bound(r, 0, MLDSA_N, MLDSA_Q))
   assigns(memory_slice(r, sizeof(int32_t) * MLDSA_N))
   ensures(array_bound(r, 0, MLDSA_N, 0, MLDSA_Q))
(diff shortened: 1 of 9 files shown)


dev/x86_64/src/arith_native_x86_64.h
```diff
@@ -109,17 +109,38 @@ void mld_poly_caddq_avx2_asm(int32_t *r)
  * in proofs/hol_light/x86_64/proofs/poly_caddq_avx2_asm.ml */
 __contract__(
   requires(memory_no_alias(r, sizeof(int32_t) * MLDSA_N))
+  requires(((uintptr_t)r % MLD_DEFAULT_ALIGN) == 0)
```
mkannwischer, line 140: Unfortunately, this doesn't work. There is no way to guarantee alignment with CBMC iirc.

mkannwischer:
As far as I know, there is no way to model an alignment precondition in CBMC.

Maybe we can consider not using `uint32_t *` for the native APIs, but instead something that guarantees alignment via `MLD_ALIGN`. It still won't be formally proven, but we can rely on the compiler to yell at us if we try to change the frontend in a way that breaks alignment for the backend.

mkannwischer:
Unfortunately, you can't model alignment with CBMC, so I don't see a way forward for this PR. Closing.

PR #1246, opened 2026-06-30
sign: Rewrite signing loop as `for` loop, rename `nonce` -> `attempt / kappa`.

### LINKED ISSUE #1245: MLD_CONFIG_MAX_SIGNING_ATTEMPTS incorrectly unset in multilevel monobuild
When using a multilevel monobuild with the MLD_CONFIG_MAX_SIGNING_ATTEMPTS option to limit the maximum number of signing attempts, the 2nd and 3rd security level (typically 65/87) do not see this option and instead use the default.

hanno-becker:
* Also resolves https://github.com/pq-code-package/mldsa-native/issues/1245

See individual commit messages.

### DIFF
--- a/mldsa/src/sign.c
+++ b/mldsa/src/sign.c
@@ -594,9 +594,7 @@ __contract__(
 /* User-facing bound on signing attempts. See MLD_CONFIG_MAX_SIGNING_ATTEMPTS
  * in mldsa_native_config.h. Default is chosen so that failure probability
  * is < 2^{-256}, that is, signatures will practically always succeed. */
-#ifndef MLD_CONFIG_MAX_SIGNING_ATTEMPTS
-#define MLD_CONFIG_MAX_SIGNING_ATTEMPTS MLD_MAX_SIGNING_ATTEMPTS
-#endif
+#if defined(MLD_CONFIG_MAX_SIGNING_ATTEMPTS)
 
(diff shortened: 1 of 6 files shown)


mldsa/src/sign.c
```diff
@@ -967,10 +967,13 @@ int mld_sign_signature_internal(uint8_t sig[MLDSA_CRYPTO_BYTES], size_t *siglen,
...
-  /* Reference: This code is re-structured using a while(1),  */
-  /* with explicit "continue" statements (rather than "goto") */
-  /* to implement rejection of invalid signatures.            */
-  while (1)
+  /* Reference: the reference implementation uses an unbounded loop. We instead
+   * iterate nonce over the statically bounded range [0, max_signing_attempts),
+   * which gives predictable termination and makes the increment of nonce
+   * provably type-safe. The loop is left early via goto cleanup once an attempt
+   * succeeds or fails fatally; if it instead runs to completion, every attempt
+   * was rejected and signing has exhausted its budget. */
```
mkannwischer: Here it sounds like it can happen that the loop runs to completion. Can you rephrase that please?
In the current shape this should really never happen.

mkannwischer (approved):
Thanks @hanno-becker. The code changes look good to me.
One comment is a little confusing.

mldsa/src/sign.c
hanno-becker: Done

PR #1247, opened 2026-06-30
Add Windows MSVC test support: rng_fail, alloc, wycheproof

willieyz:
- Ported from: https://github.com/pq-code-package/mlkem-native/pull/1738

mldsa-native already runs tests on Windows via Mingw-w64, but the MSVC/nmake build only covers the functional tests. This PR closes the remaining gap by porting the following tests to Makefile.Microsoft_nmake from mlkem-native PR [#1738](https://github.com/pq-code-package/mlkem-native/pull/1738), against the portable C backend:

- rng_fail: link rules added.
- alloc: added test/src/test_alloc.c and test/configs/test_alloc_config.h, referencing the Linux Makefile. Needs to rebuild the .obj files since it uses a different config from the default functional test. A new ALLOC_CFLAGS macro keeps the build rules short.
- wycheproof: test obj building rules and link rules added.
Unit tests are intentionally omitted from this PR and can be added later once native backend support becomes available on Windows.


mkannwischer (approved):
Thanks @willieyz. LGTM.

PR #1248, opened 2026-06-30
Remove `crypto_sign_*` API aliases and `MLD_CONFIG_NO_SUPERCOP`

### LINKED ISSUE #1249: Consider removing the remaining `crypto_sign_*` API aliases and `MLD_CONFIG_NO_SUPERCOP`
Now that the SUPERCOP signed-message API (`crypto_sign` / `crypto_sign_open`) is gone, we should consider dropping the rest of the `crypto_sign_*` naming scheme. Of these, only `crypto_sign_keypair`, `crypto_sign`, and `crypto_sign_open` were genuine SUPERCOP APIs; `crypto_sign_signature` / `crypto_sign_verify` (and the `_internal` / `_extmu` / `_pre_hash` variants) were mldsa-native additions stapled on top of the same prefix, so keeping them plus the `MLD_CONFIG_NO_SUPERCOP` / `MLD_CONFIG_API_NO_SUPERCOP` option that gates them adds little — consumers can use the native namespaced API (e.g. `mldsa_keypair`) directly. The `CRYPTO_{PUBLICKEY,SECRETKEY,}BYTES` size constants would be retained as plain size macros.

mkannwischer:
- Depends on #1206 
- Resolves https://github.com/pq-code-package/mldsa-native/issues/1249

mldsa_native.h exposed the public API under the crypto_sign_* naming
scheme, gated by MLD_CONFIG_NO_SUPERCOP.
Of these, only crypto_sign_keypair -- together with the already-removed
crypto_sign and crypto_sign_open -- is an actual SUPERCOP API.
crypto_sign_signature and crypto_sign_verify (and the _internal / _extmu /
_pre_hash variants used by the tests) were mldsa-native additions stapled
on top of the same crypto_sign_ prefix.

This commit removes those aliases and the MLD_CONFIG_NO_SUPERCOP configuration
entirely, along with the related documentation in mldsa_native_config.h,
the example configs/READMEs, and the #error messages in common.h.

The CRYPTO_{PUBLICKEY,SECRETKEY,}BYTES size constants are retained in
mldsa_native.h as plain size macros (derived from MLD_CONFIG_PARAMETER_SET
for single-parameter-set builds), decoupled from the removed aliases.

Consumers and tests now use the native namespaced API directly:
- Examples call the namespaced functions (e.g. mldsa_keypair).
- Tests reconstruct the namespaced names via test/src/test_namespace.h. The
  per-parameter namespace currently cannot be removed due to the
  check-namespace tests the building of libmldsa.a.


hanno-becker (commented):
If possible, I think this should be reordered with #1240: Remove any trace of SUPERCOP alignment first, and then modify the API.

Generally, I support the change, as well as the follow-up #1240.

hanno-becker:
@mkannwischer I took the liberty to rebase this atop #1206 instead of #1240.

README.md
```diff
@@ -191,7 +191,7 @@ To enable this mode, define `MLD_CONFIG_REDUCE_RAM` in [mldsa_native_config.h](m
 
 By default, mldsa-native uses the randomized "hedged" signing variant as specified in FIPS 204 Section 3.4. The hedged variant uses both fresh randomness at signing time and precomputed randomness from the private key. This helps mitigate fault injection attacks and side-channel attacks while protecting against potential flaws in the random number generator.
 
-If you need the deterministic variant of ML-DSA, you can call `crypto_sign_signature_internal`
+If you need the deterministic variant of ML-DSA, you can call `signature_internal`
```
hanno-becker, line 194: Without any prefix, the API references are a bit confusing to me. Any ideas?

hanno-becker: We can discuss this in a follow-up.

hanno-becker (approved):

PR #1264, opened 2026-07-02
CI: Add merge queue support

mkannwischer:
Run CI on merge_group events and add a "CI Summary" gate job intended as the single required status check. Add a workflow reporting the "DCO" check on merge groups since the DCO app does not support them (dcoapp/app#199); DCO is still enforced on the PR before it can enter the queue. Also run the HOL-Light proofs (self-gated by their changed-files check) and the legacy compiler tests on merge groups.


hanno-becker (approved):
Let's try it.

PR #1270, opened 2026-07-08
Fix double level-suffixing of default namespace in multi-level builds

### LINKED ISSUE #1266: Default namespace subject to inconsistent level-suffixing in multi-level build
The default namespace is `PQCP_MLDSA_NATIVE_MLDSA{44,65,87}_` but in a multi-level build, the prefix itself is suffixes by the level -- and what's more, the config is only imported once, for the 44-level. In effect, you'll get the namespace prefixes `PQCP_MLDSA_NATIVE_MLDSA44{44,65,87}_XXX`, which makes no sense of course.

There's no functional issue here, but the naming is undesired and confusing.

mkannwischer:
The default namespace prefix embeds the parameter set (PQCP_MLDSA_NATIVE_MLDSA44/65/87), which is correct for single-level builds but wrong in a multi-level build: the config is processed once (at the 44 level) and the namespacing machinery appends the parameter set again, yielding PQCP_MLDSA_NATIVE_MLDSA44{44,65,87}_. Use a level-independent default (PQCP_MLDSA_NATIVE_MLDSA) when MLD_CONFIG_MULTILEVEL_BUILD is set so the level is appended exactly once. Single-level builds are unchanged.

- Fixes #1266

### DIFF
--- a/examples/basic_deterministic/mldsa_native/mldsa_native_config.h
+++ b/examples/basic_deterministic/mldsa_native/mldsa_native_config.h
@@ -748,7 +748,11 @@
  * e.g., PQCP_MLDSA_NATIVE_MLDSA44_
  */
 
-#if MLD_CONFIG_PARAMETER_SET == 44
+#if defined(MLD_CONFIG_MULTILEVEL_BUILD)
+/* In a multi-level build the parameter set is appended by the namespacing
+ * machinery, so the default prefix must not embed it. */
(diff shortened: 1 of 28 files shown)


hanno-becker (approved):
That's ok as an immediate fix. I wonder though if we should just always append the parameter set and thereby get rid of the multilevel distinction.

mkannwischer:
> That's ok as an immediate fix. I wonder though if we should just always append the parameter set and thereby get rid of the multilevel distinction.

Yes, good idea to simplify the configuration.

PR #1271, opened 2026-07-08
Add Armv8.1-m MVE memcpy, memset, zeroize

### LINKED ISSUE #1334: Armv8.1-M: re-evaluate opt-in MVE memcpy, memset, and zeroize without C-only regression
Parent: https://github.com/pq-code-package/mldsa-native/issues/1319

Current draft PR: https://github.com/pq-code-package/mldsa-native/pull/1271

## Objective

Rebase, isolate, and re-evaluate the opt-in Armv8.1-M MVE implementations of
`memcpy`, `memset`, and zeroize after each preceding backend PR. Do not land
them while they cause an unexplained C-only regression.

This issue is deferred; it does not require SLOTHY generation.

## Scope

- Split the memory-helper change from already-merged platform history.
- Keep the helpers opt-in and gated on MVE availability.
- Preserve the existing C-only implementation and output path when the hooks
  are disabled.
- Document the tail-predicated loop, q/predicate/GPR clobbers, stack behavior,
  and inline-assembly memory contract.
- Preserve `memcpy` return and non-overlap semantics, `memset` byte semantics,
  and non-elidable zeroization.

## Acceptance criteria

- [ ] Focused tests cover lengths 0, 1-15, 16, 17, odd/tail cases, and
  representative large lengths and alignments.
- [ ] C-only and MVE-enabled QEMU functional tests pass for ML-DSA-44,
  ML-DSA-65, and ML-DSA-87.
- [ ] ABI, predicate initialization, vector clobber, and compiler-barrier
  behavior are checked.
- [ ] The final PR contains only the memory-helper/configuration change and its
  tests/docs.

## Required N657 comparison

- **Pre:** the standard memory hooks for each claimed configuration.
- **Post:** the MVE memory hooks with all other backends held constant.
- **Non-regression control:** C-only before and after the PR.

Record same-session keygen/sign/verify raw lines and min/median/p95 for each
claimed reduced-RAM or native configuration. Signing is rejection-variable, so
retain every raw sample and do not base the decision on a single observation.

Do not proceed while a repeatable C-only regression remains, unless maintainers
explicitly accept and document the trade-off.

bremoran:
Closes #1334

## Summary

Add opt-in Armv8.1-M MVE implementations of the existing custom `memcpy`,
`memset`, and zeroization hooks.

The helpers use tail-predicated 16-byte MVE loads and stores, preserving:

- `memcpy` return and non-overlap semantics
- `memset` byte semantics
- non-elidable zeroization
- the existing C implementation when the hooks are disabled

Documentation and an Armv8.1-M MVE test configuration are included.

## Status

This remains a draft. Initial STM32N657 measurements improved key generation
but regressed signing and verification. It should not merge until the change is
rebased and isolated, focused correctness and ABI tests pass, and same-session
N657 benchmarks demonstrate no unexplained C-only regression.

Existing CI and CBMC checks pass for ML-DSA-44, ML-DSA-65, and ML-DSA-87,
including reduced-RAM configurations.


bremoran:
I've marked this PR as a draft because while MVE memset/memcpy/zeroize did have a positive impact on reducing key pair generation, it had the opposite effect on signature generation and verification. This should not be the case and was not the case when I added it in local testing. However, I added it after other assembly optimisations, not before. Therefore, I'll defer this one until it makes an improvement in all benchmarks.

PR #1275, opened 2026-07-09
Fetch ACVP/Wycheproof vectors via authenticated GitHub API

mkannwischer:
Download test vectors from the api.github.com contents endpoint with a GH_TOKEN and retry on 429/403, replacing unauthenticated raw.githubusercontent requests that were hitting rate limits frequently.


hanno-becker (approved):
Worth trying, let's see if it helps.

PR #1277, opened 2026-07-10
Add Armv8.1-M Keccak x1 backend

### LINKED ISSUE #1322: Armv8.1-M: add regular and SLOTHY-optimized Armv7-M Keccak x1
Parent: https://github.com/pq-code-package/mldsa-native/issues/1319

Current PR: https://github.com/pq-code-package/mldsa-native/pull/1277

## Objective

Land the Adomnicai/XKCP-derived Armv7-M Keccak-f[1600] x1 backend for
Armv8.1-M and reduced-RAM mldsa-native configurations. Provide both the regular
assembly path and the SLOTHY-optimized assembly selected by the native backend.

## Scope

- Add the bit-interleaved x1 permutation and its 49-word round-constant table.
- Wire the backend through the native FIPS202 API and configuration guards.
- Add the x1 state-XOR and state-extract conversions required to keep the
  state bit-interleaved across permutations.
- Keep clean/regular generation inputs and the SLOTHY-optimized Armv7-M output
  reviewable.
- Document the SLOTHY revision, generation command, model/configuration, and
  timeout. Valid schedules may vary between solver runs.
- Document provenance and licensing for the XKCP/Adomnicai and SLOTHY-derived
  portions.
- Document the AAPCS32 interface, 8-byte state alignment, stack use, round
  constant buffer, and register clobbers.
- Keep assembly local labels safe when mldsa-native and mlkem-native objects
  are combined; see #1308.

## Acceptance criteria

- [x] Regular and SLOTHY-optimized assembly generation paths are documented
  and complete successfully.
- [x] Both generated implementations assemble and pass focused native-versus-C
  tests.
- [x] The SLOTHY correctness/self-check passes.
- [x] Autogeneration, formatting, lint, and generated-source checks pass.
- [x] Native-versus-C Keccak x1 differential/unit tests pass.
- [x] QEMU functional tests pass for ML-DSA-44, ML-DSA-65, and ML-DSA-87 with
  the native backend enabled.
- [x] The corresponding C-only configurations remain passing.
- [ ] The Armv8.1-M ABI check passes for the x1 assembly interface.
- [ ] The PR body explains any dev-tree moves or autogen changes that cannot be
  split from the backend.

## Required N657 comparison

- **Pre:** C-only reduced-RAM mldsa-native.
- **Post:** the SLOTHY-optimized Armv7-M Keccak x1 backend from this PR.

Use the same reduced-RAM configuration, toolchain, code/data placement, and
benchmark session. Record exact commits and commands, raw direct
`keccak_f1600_x1` component lines, and min/median/p95. Include full-operation
rows .

Related test infrastructure: #1318. Existing native Keccak unit-test groundwork
is recorded in #904.

bremoran:
Add the unmodified upstream Adomnicai Armv7-M Keccak source under armv81m_clean and the Slothy-generated M7 optimized x1 permutation under armv81m_opt, with a Makefile regeneration path.

Move the Armv8.1-M FIPS202 development sources to armv81m_opt and synchronize the production backend. Route x1 state XOR/extract through the clean-source assembly helpers so the optimized permutation can keep the state in the bit-interleaved native representation.

Keep the test changes focused on representation-aware Keccak x1/x4 unit coverage, including extracted-byte state dumps on x1 failures, and the static ML-DSA-87 unit-test workspace needed for Zephyr stack pressure.

Fixes #1322


bremoran:
Depends on either #1312 or #1318

bremoran:
Added metadata YAML to the input source file so that slothy will preserve it and autogen will pass. This is a comment-only change and benchmarks do not need to be re-run.

mldsa/src/fips202/keccakf1600.c
```diff
@@ -40,6 +40,13 @@ void mld_keccakf1600_extract_bytes(uint64_t *state, unsigned char *data,
...
   unsigned i;
+#if defined(MLD_USE_NATIVE_FIPS202_X1_EXTRACT_BYTES)
+  if (mld_keccakf1600_extract_bytes_x1_native(state, data, offset, length) ==
+      MLD_NATIVE_FUNC_SUCCESS)
+  {
+    return;
+  }
+#endif /* MLD_USE_NATIVE_FIPS202_X1_EXTRACT_BYTES */
 #if defined(MLD_SYS_LITTLE_ENDIAN)
   uint8_t *state_ptr = (uint8_t *)state + offset;
```
mkannwischer, line 51: This breaks C90 on platforms impkementing exract bytes due to the state_ptr declaration a little lower. that needs to be fixed. Also maybe worth switching to C90 to catch this in the future (unless this leads to more problems, then please open an issue).

dev/fips202/armv81m_opt/src/keccak_f1600_x1_armv7m.c
```diff
@@ -0,0 +1,101 @@
...
+    0x00000000, 0x80000081, /* RC16 */
+    0x00000000, 0x80000008, /* RC17 */
+    0x00000000, 0x00000083, /* RC18 */
+    0x00000000, 0x80008003, /* RC19 */
+    0x00000001, 0x80008088, /* RC20 */
+    0x00000000, 0x80000088, /* RC21 */
+    0x00000001, 0x00008000, /* RC22 */
+    0x00000000, 0x80008082, /* RC23 */
+    0x000000ff,             /* loop terminator */
+};
```
mkannwischer, line 54: Please auto generate in autogen this in a similar way it is autogenerated for other platforms.

dev/fips202/armv81m_opt/src/keccak_f1600_x1_armv7m.c
```diff
@@ -0,0 +1,101 @@
...
+void mld_keccak_f1600_x1_state_xor_bytes_asm(void *state, const uint8_t *data,
+                                             unsigned offset, unsigned length);
+
+#define mld_keccak_f1600_x1_state_extract_bytes_asm \
+  MLD_NAMESPACE(keccak_f1600_x1_state_extract_bytes_asm)
+void mld_keccak_f1600_x1_state_extract_bytes_asm(void *state, uint8_t *data,
+                                                 unsigned offset,
+                                                 unsigned length);
+
+/* The Adomnicai x1 core consumes 24 bit-interleaved round constants followed
```
mkannwischer, line 26: Let's acknowledge the source of this implementation once in the assembly and then keep all other comments general, i.e., do not talk about the `Adomnicai x1 core` - it only leads to confusion later on.

mldsa/src/fips202/native/armv81m/src/keccak_f1600_x1_state_extract_bytes_armv7m.S
```diff
@@ -0,0 +1,270 @@
+/*
+ * Copyright (c) The mlkem-native project authors
+ * Copyright (c) The mldsa-native project authors
+ * SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+ */
+
+/*
+ * This helper is derived from the public-domain XKCP implementation and the
```
mkannwischer, line 8: Be more specific here: CC0. public domain means different things for different people.

mldsa/src/fips202/native/armv81m/src/keccak_f1600_x1_state_extract_bytes_armv7m.S
```diff
@@ -0,0 +1,270 @@
...
+ * Armv7-M optimizations described in [ADOMNICAI23]. See the backend README,
```
mkannwischer, line 9: Please cite the implementation directly

mldsa/src/fips202/native/armv81m/src/keccak_f1600_x1_state_extract_bytes_armv7m.S
```diff
@@ -0,0 +1,270 @@
...
+ * BIBLIOGRAPHY, and LICENSE files for full provenance and license terms.
```
mkannwischer, line 10: You are referencing a README here that does not actually ship in mldsa/, but only in dev/. 
Consumers are not expected to import dev/, so either we need to sync over the README or we just make the .S file self-contained. I prefer the latter. 

Same applies to BIBLIOGRAPHY: it would be better to not assume that is present in the consumer tree.

dev/fips202/armv81m_opt/src/Makefile
```diff
@@ -0,0 +1,15 @@
...
+
+.PHONY: all clean
+
+PYTHON ?= python3
+SLOTHY_KECCAK_M4 ?= slothy_keccak_m4.py
+
+all: keccak_f1600_x1_armv7m_opt_m7.S
+
+keccak_f1600_x1_armv7m_opt_m7.S: ../../armv81m_clean/src/keccak_f1600_x1_armv7m.S $(SLOTHY_KECCAK_M4)
+	$(PYTHON) $(SLOTHY_KECCAK_M4) $< $@
```
mkannwischer, line 12: I don't think the keccak_f1600_x1_armv7m_opt_m7.S was produced like this. The input contains mld_keccak_f1600_x1_armv7m_asm_StatePermute_RoundConstantsWithTerminator, but the output doesn't.

Please either remove the SLOTHY optimization from this PR, or make sure that one can produce the committed output using this Makefile (modulo scheduling and register allocation, obviously).

test/zephyr/platform.mk
```diff
@@ -33,8 +33,12 @@ ZEPHYR_BOARD_mps3-an547 := mps3/corstone300/an547
 ZEPHYR_QEMU_mps3-an547  := mps3-an547                    # Cortex-M55
 ZEPHYR_BOARD_nucleo-n657x0-q := nucleo_n657x0_q          # Cortex-M55 (hardware)
 
-ZEPHYR_FIPS202_BACKEND_mps3-an547 := fips202/native/armv81m/mve.h
-ZEPHYR_FIPS202_BACKEND_nucleo-n657x0-q := fips202/native/armv81m/mve.h
+ZEPHYR_FIPS202_BACKEND_mps3-an547 := fips202/native/armv81m/mve_x1.h
+ZEPHYR_FIPS202_BACKEND_nucleo-n657x0-q := fips202/native/armv81m/mve_x1.h
```
mkannwischer, line 37: This swaps out the backend. Does this introduce a test gap as the x4 Keccak is no longer tested? We should not be regressing test coverage, so we need to at least add a x4 unit test to CI.

mldsa/src/fips202/native/armv81m/mve_x1.h
```diff
@@ -0,0 +1,65 @@
...
+#define MLD_USE_NATIVE_FIPS202_X1_EXTRACT_BYTES
+/* Guard for assembly files */
+#define MLD_FIPS202_ARMV81M_NEED_X1
+
+#if !defined(__ASSEMBLER__)
+#include "../api.h"
+
+#define mld_keccak_f1600_x1_native_impl \
+  MLD_NAMESPACE(keccak_f1600_x1_native_impl)
+int mld_keccak_f1600_x1_native_impl(uint64_t *state);
```
mkannwischer, line 24: These should be marked `MLD_INTERNAL_API`

test/zephyr/platform.mk
```diff
@@ -55,12 +59,27 @@ endif
...
+# key so command-line overrides invalidate binaries built with different counts.
+NTESTS ?= 3
+NUM_RANDOM_TESTS ?= 100
+
+# Forward the QEMU execution deadline to exec_wrapper.py. Direct wrapper
+# invocations use the same default when the variable is absent.
+QEMU_TIMEOUT ?= 300
+export QEMU_TIMEOUT
+
 # Native backends are an OPT=1 feature (an547 builds the Armv8.1-M MVE backend).
```
mkannwischer, line 73: stale

test/src/test_unit.c
```diff
@@ -65,48 +65,75 @@ unsigned int mld_rej_eta_c(int32_t *a, unsigned int target, unsigned int offset,
 void mld_keccakf1600_permute_c(uint64_t *state);
 
 #if defined(MLD_USE_NATIVE_FIPS202_X1)
```
mkannwischer, line 67: CI does not run unit tests for Armv8.1-M. Please enable it.

LICENSE
```diff
@@ -14,11 +14,17 @@ It is only used for testing purposes.
 The benchmarking code in test/hal/hal.c carries the
 MIT license. It is only used for testing purposes.
 
+The Armv8.1-M Keccak implementation contains portions adapted from SLOTHY
+examples distributed under the MIT license. The applicable SLOTHY copyright
+notices are included below.
+
```
mkannwischer, line 20: Which portions are MIT licensed? I did not see any file in this PR being MIT-only licensed and we wouldn't be able to take it for anything under mldsa/. Maybe this is an outdated notice that can be removed?

mkannwischer (changes requested):
Thanks @bremoran. 

I have a number of comments. Please look at the LICENSE first as that will likely block this PR.

mldsa/src/fips202/keccakf1600.c
bremoran: Good catch. I moved the state_ptr declarations in both the xor and extract paths ahead of native backend dispatch, so no declaration follows a statement and these paths remain C90-compatible. General -std=c90 coverage already exists; I’m keeping any additional backend-specific C90 CI work outside this PR’s reduced scope.

dev/fips202/armv81m_opt/src/keccak_f1600_x1_armv7m.c
bremoran: Agreed. scripts/autogen now emits the interleaved x1 constants into keccakf1600_round_constants_x1.c, and the wrapper uses the generated declaration. In the reduced version, the statically bounded 24-round loop will also eliminate the sentinel word and the associated test_bytes machinery.

dev/fips202/armv81m_opt/src/keccak_f1600_x1_armv7m.c
bremoran: Agreed. The wrapper wording is now implementation-neutral. As part of the reduced branch, I will consolidate the remaining source-specific attribution into a single assembly provenance block and regenerate the production copy. I’ll leave this thread open until that is visible.

mldsa/src/fips202/native/armv81m/src/keccak_f1600_x1_state_extract_bytes_armv7m.S
bremoran: Agreed. I replaced the ambiguous “public domain” wording with the explicit CC0-1.0 identifier.

mldsa/src/fips202/native/armv81m/src/keccak_f1600_x1_state_extract_bytes_armv7m.S
bremoran: Addressed. The provenance block now identifies the exact source file, repository URL, and pinned commit. The production assembly is therefore self-contained and does not depend on the development README or bibliography.

test/zephyr/platform.mk
bremoran: Good catch. The current workflow retains x1 as the default and adds a separate Armv8.1-M x4 job selecting mve.h and running the functional, KAT, ACVP, and unit suites. Since this is an independent CI repair, I will preserve it as a separate change if we agree on resetting the scope of #1277.

mldsa/src/fips202/native/armv81m/mve_x1.h
bremoran: Agreed. All three x1 implementation declarations and their definitions are now marked MLD_INTERNAL_API.

test/zephyr/platform.mk
bremoran: Agreed; that hunk no longer exists in the current version.

test/src/test_unit.c
bremoran: Agreed. The current workflow enables unit tests for both the default an547 x1 configuration and the explicit x4 job.

However, I also agree that the static-workspace workaround in test_unit.c is orthogonal and changes test allocation semantics for every platform. I will remove it from #1277 and leave this thread open until the reduced branch has Armv8.1-M unit coverage through a properly scoped solution.

LICENSE
bremoran: Thanks. I rechecked the provenance, and the added notice does not resolve this.

The notice was intended to cover the SLOTHY-derived scheduling input and optimized permutation, including the development source and its synchronized production output. I cannot currently establish a sufficiently clean file-level boundary to support the file-wide Apache-2.0 OR ISC OR MIT assertion, and the copyright block associated with the notice is not an adequate SLOTHY 0.2.2 notice.

For the reduced PR, I will remove all SLOTHY-derived input, scheduling scripts, generated output, and the corresponding LICENSE paragraph. I will retain only material traceable directly to the cited CC0-1.0 source and newly authored, compatibly licensed changes. A scheduled optimization can return separately only after an accepted independent re-derivation or explicit relicensing. I’ll leave this thread open until that removal is visible in the diff.

PR #1285, opened 2026-07-15
aarch64: Use array parameters consistently in asm backend

### LINKED ISSUE #1033: AArch64 backend: Use array parameters consistently
Currently we are using a mix of arrays and pointers:

```c
void mld_polyvecl_pointwise_acc_montgomery_l4_asm(int32_t *r,
                                                  const int32_t a[4][MLDSA_N],
                                                  const int32_t b[4][MLDSA_N])Collapse comment
```

This should be unified to use arrays consistently.

mkannwischer:
Convert all fixed-size pointer parameters in the AArch64 assembly interface to array notation.

 - Resolves #1033


hanno-becker (approved):
LGTM. Thanks @mkannwischer!

PR #1287, opened 2026-07-15
CBMC: Functional specs for `make_hint` and hint unpacking

### LINKED ISSUE #1099: CBMC: Consider providing functional specs for make_hint and unpack_hint
cf https://github.com/pq-code-package/mldsa-native/pull/1028#discussion_r3186069202

The current specs are enough for type-safety and memory-safety, but it seems we have an opportunity to capture (some aspect of) correctness of hint packing/unpacking as well. Investigate.

mkannwischer:
Strengthen the contracts beyond type/memory safety: pin mld_make_hint's return value to the exact hint predicate, and prove mld_sig_unpack_hints sets h->coeffs[p] exactly for the hint indices decoded for row i.

 - Resolves https://github.com/pq-code-package/mldsa-native/issues/1099

### DIFF
--- a/mldsa/src/packing.c
+++ b/mldsa/src/packing.c
@@ -166,6 +166,9 @@ int mld_sig_unpack_hints(mld_poly *h, const uint8_t sig[MLDSA_CRYPTO_BYTES],
     invariant(j >= old_hint_count && j <= new_hint_count &&
               new_hint_count <= MLDSA_OMEGA)
     invariant(array_bound(h->coeffs, 0, MLDSA_N, 0, 2))
+    invariant(forall(p, 0, MLDSA_N,
+      (h->coeffs[p] == 1) ==
+      exists(hj, old_hint_count, j, packed_hints[hj] == p)))
     decreases(new_hint_count - j)
   )
   {
@@ -192,6 +195,14 @@ int mld_sig_unpack_hints(mld_poly *h, const uint8_t sig[MLDSA_CRYPTO_BYTES],
       }
     }
   }
+
+  /* On success, h->coeffs[p] is 1 exactly for the hint indices decoded for this
+   * row, i.e. packed_hints[old_hint_count, new_hint_count). Asserted here
+   * rather than posted as a contract so callers need not assume it. */
+  cassert(forall(
+      p, 0, MLDSA_N,
+      (h->coeffs[p] == 1) ==
+          exists(hj, old_hint_count, new_hint_count, packed_hints[hj] == p)));
 
   return 0;
 }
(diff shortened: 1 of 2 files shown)


hanno-becker:
I support strengthening the functional specification as much as CBMC allows. However, we need to find a way around the unnecessary complexity it imposes on caller that don't care about the added functional properties. Perhaps there is a (clean) way one can mark `ensures` clauses as "check-only", so they are taken into account when the contract is validated, but ignored when the contract is used. That would be sound.

mkannwischer:
> I support strengthening the functional specification as much as CBMC allows. However, we need to find a way around the unnecessary complexity it imposes on caller that don't care about the added functional properties. Perhaps there is a (clean) way one can mark `ensures` clauses as "check-only", so they are taken into account when the contract is validated, but ignored when the contract is used. That would be sound.

Wouldn't that be the same as an assertion at the end of the function?
I'll first check if the `sign_verify_internal` proof can be rescued through the usual tricks - I had only tested the proofs for the changed contracts locally.

mkannwischer:
No luck with fixing `sign_verify_internal` using the usual tricks.
Adding a assertion in the end of the function does work, but I think we want the post-condition to also serve as documentation. 
I have added a new macro `functional` (naming TBD) that can be used to wrap the post-condition so that it is only active for the proof of `unpack_hints`, but not for the callers. That works for me locally.
We could also define a different variant of `ensures`. 
Any ideas @hanno-becker?

hanno-becker:
> Wouldn't that be the same as an assertion at the end of the function?

Yes, functionally it is, but as you say, conceptually it belongs in the contract.

hanno-becker:
> I have added a new macro functional (naming TBD) that can be used to wrap the post-condition so that it is only active for the proof of unpack_hints, but not for the callers. That works for me locally. We could also define a different variant of ensures. Any ideas @hanno-becker?

I can't think of anything I like. For the time being, I think we should start with an assertion.

mkannwischer:
> > I have added a new macro functional (naming TBD) that can be used to wrap the post-condition so that it is only active for the proof of unpack_hints, but not for the callers. That works for me locally. We could also define a different variant of ensures. Any ideas @hanno-becker?
> 
> I can't think of anything I like. For the time being, I think we should start with an assertion.

As in you do not like what i have right now?

hanno-becker:
Yes, I'm not convinced by the approach yet; you can shoot yourself in the foot by using a functional spec but not proving it. Of course we're doing it the other way around, but nothing enforces it?

mkannwischer:
> Yes, I'm not convinced by the approach yet; you can shoot yourself in the foot by using a functional spec but not proving it. Of course we're doing it the other way around, but nothing enforces it?

Indeed, that is a valid concern.
Let's change it to an assertion for now and open an issue to discuss how additional function specs can be added that don't affect callees (I went through the CBMC documentation, but didn't find anything that would help with that).

hanno-becker:
Yes, I think that's the right approach.

mkannwischer:
I have opened a follow-up issue: https://github.com/pq-code-package/mldsa-native/issues/1292

hanno-becker (approved):
Thank you @mkannwischer! I slightly tweaked the comment.

PR #1288, opened 2026-07-16
Add missing MLD_INTERNAL_API/MLD_EXTERNAL_API declarations

hanno-becker:
```
    Annotate mld_polyeta_unpack, mld_prepare_domain_separation_prefix,
    mld_debug_check_assert, mld_debug_check_bounds, and
    mld_keccak_f1600_x4_native_impl.
```

* Extracted from #1052.

### DIFF
--- a/dev/fips202/armv81m/mve.h
+++ b/dev/fips202/armv81m/mve.h
@@ -26,6 +26,7 @@
  */
 #define mld_keccak_f1600_x4_native_impl \
   MLD_NAMESPACE(keccak_f1600_x4_native_impl)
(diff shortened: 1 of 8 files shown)


mkannwischer (changes requested):
Should `mld_keccak_f1600_x4_native_impl` also be marked as internal?

(same applies to the avx2 intrinsics, but those are going to be removed shortly anyway).

hanno-becker:
> Should mld_keccak_f1600_x4_native_impl also be marked as internal?

Yes, updated.

mkannwischer (approved):

PR #1305, opened 2026-07-21
Add compile-time guard for use of Neon

### LINKED ISSUE #1304: Build fails for -march=armv8-a+nosimd
We don't check for __ARM_NEON support. This needs to be dealt with in the same way as AVX2.

- [ ] Add MLD_SYS_AARCH64_NEON
- [ ] Replace MLD_SYS_AARCH64 by MLD_SYS_AARCH64_NEON as appropriate (in most but not all backends; e.g. FIPS202 has a purely scalar backend as well)
- [ ] Test armv8-a+nosimd in CI

hanno-becker:
* Depends on: #1301 
* Resolves #1304


mldsa/src/sys.h
```diff
@@ -39,6 +39,21 @@
...
 #endif
 
+/* Check if the AArch64 compilation target supports NEON (Advanced SIMD).
+ *
+ * Some compilers also define __ARM_NEON__, but __ARM_NEON is the most reliable
+ * signal. Specifically, clang on Apple appears to keep __ARM_NEON__ set even if
+ * -march=armv8-a+nosimd is set.
+ *
+ * gcc 4.8 -- the first gcc version introducing Neon support -- sets neither
+ * __ARM_NEON nor __AMR_NEON__; in fact, there is no preprocessor signal that
```
mkannwischer: AMR -> ARM

test/abicheck/aarch64/abicheck_aarch64.mk
```diff
@@ -41,6 +41,7 @@ ABICHECK_REQ_NEON_FILES := \
   mldsa/src/native/aarch64/src/rej_uniform_eta2_aarch64_asm.S \
   mldsa/src/native/aarch64/src/rej_uniform_eta4_aarch64_asm.S
 ABICHECK_REQ_NEON_OBJS := $(call MAKE_OBJS,$(ABICHECK_DIR),$(ABICHECK_REQ_NEON_FILES))
+$(ABICHECK_REQ_NEON_OBJS): CFLAGS += -march=armv8-a+simd
```
mkannwischer, line 44: Does this intentionally overlap with the sha3 files?

mldsa/src/sys.h
```diff
@@ -39,6 +39,21 @@
...
+ * Neon is enabled. If you use gcc 4.8, you should set __ARM_NEON manually.
+ * gcc 4.9 onwards do set __ARM_NEON.
+ */
+#if defined(MLD_SYS_AARCH64) && defined(__ARM_NEON)
```
mkannwischer, line 53: This does not cover MSVC which only defines _M_ARM64.
Maybe this can be added later when we merge the armasm64 support

dev/fips202/aarch64/auto.h
mkannwischer, line 37: Does this also need the extra neon guard?
Claude claims one can enable sha3 instructions while disabling the rest of Neon in some toolchains. Sounds odd to me.

mldsa/src/sys.h
```diff
@@ -39,6 +39,21 @@
...
 
+/* Check if the AArch64 compilation target supports NEON (Advanced SIMD).
+ *
+ * Some compilers also define __ARM_NEON__, but __ARM_NEON is the most reliable
+ * signal. Specifically, clang on Apple appears to keep __ARM_NEON__ set even if
+ * -march=armv8-a+nosimd is set.
+ *
+ * gcc 4.8 -- the first gcc version introducing Neon support -- sets neither
+ * __ARM_NEON nor __AMR_NEON__; in fact, there is no preprocessor signal that
+ * Neon is enabled. If you use gcc 4.8, you should set __ARM_NEON manually.
```
mkannwischer: It feels cleaner to recommend defining the MLD_SYS_... macro directly rather than defining a macro that is usually set by the compiler. Can we make the macro user definable?

mkannwischer (approved):
Thanks @hanno-becker. I have a few comments/questions but none of them are blocking.

dev/fips202/aarch64/auto.h
hanno-becker: I deliberately left it out here to avoid clutter; I don't agree that having SHA3 instructions without Neon is a possibility.

test/abicheck/aarch64/abicheck_aarch64.mk
hanno-becker: Good catch, this is brittle, will fix.

mldsa/src/sys.h
hanno-becker: Yes

mldsa/src/sys.h
hanno-becker: Yes

test/abicheck/aarch64/abicheck_aarch64.mk
hanno-becker: Upon further consideration, I will leave this as-is for now. It works as-is, and improving the robustness of the ABI checker autogeneration in the face of multiple mutually non-inclusive architecture capabilities is a larger rework that is not needed and not really related to this PR.

PR #1307, opened 2026-07-22
ACVP: Support FIPS204-tr1 sigGen (v1.1.0.43)

### LINKED ISSUE #1303: ACVP: Support v1.1.0.43
From the release notes:

ML-KEM / encapDecap / FIPS203-tr1 and ML-DSA / sigGen / FIPS204-tr1 (Algo / Mode / Revision) - The new ML-KEM encapDecap and ML-DSA sigGen testing revisions add the possibility to use the seed, expanded private key format, or both in testing. This is accomplished by a new registration property: keyFormats.

Task:
- Extend the ACVP tests to additionally support the new tests types

mkannwischer:
The FIPS204-tr1 sigGen revision adds keyFormat 'seed'/'expanded' test groups: the private key may be supplied as a 32-byte seed to expand rather than as the expanded sk.

Add a decode_sk() entry point to the ACVP harness accepting either sk=HEX or seed=HEX (expanding the seed via keyGen), and route all sigGen sub-modes through it. Download and run the ML-DSA-sigGen-FIPS204-tr1 vectors when the version ships them (v1.1.0.43+).

- Resolves #1303


test/acvp/acvp_client.py
```diff
@@ -555,6 +597,13 @@ def test(
         info("No test data to run (all modes disabled in this build)")
         return
 
+    # Seed key-format cases require keyGen to expand the key; drop them when the
+    # build does not provide it (unconditionally, unlike --skip-unsupported).
+    if supported_modes is not None and "keyGen" not in supported_modes:
+        seed_reasons = filter_test_cases(data, seed_key_format)
+        if seed_reasons:
+            info(f"Skipping {len(seed_reasons)} test case(s): {seed_reasons[0]}")
```
hanno-becker: Nit: Could be cleaner -- if in the future we introduce multiple reasons of rejection in `seed_key_format` (or  change the filter to something with multiple reasons) then `seed_reasons[0]` would no longer be correct.

Or, one level up, one could implement a _single_ filter function taking care of hashes and the key filter.

test/acvp/acvp_mldsa.c
```diff
@@ -217,10 +218,49 @@ static void acvp_mldsa_keyGen_AFT(const unsigned char seed[MLDSA_RNDBYTES])
...
+#if !defined(MLD_CONFIG_NO_KEYPAIR_API)
+    unsigned char pk[MLDSA_PK_BYTES];
+    unsigned char *seed = decode_hex("seed", MLDSA_SEEDBYTES, arg);
+    if (seed == NULL)
+    {
+      return NULL;
+    }
+    if (mld_sign_keypair_internal(pk, acvp_expanded_sk, seed) != 0)
+    {
+      return NULL;
```
hanno-becker, line 249: This would deserve a separate error message. It currently falls through to the usage string being printed, which is misleading.

hanno-becker (approved):
Thanks @mkannwischer! Overall LGTM, two nits, none blocking.

test/acvp/acvp_client.py
mkannwischer: Fixed by joining the reasons.

test/acvp/acvp_mldsa.c
mkannwischer: Added.

PR #1309, opened 2026-07-23
Split ambiguous MLD_ERR_FAIL into specific error codes

### LINKED ISSUE #1306: MLD_ERR_FAIL is ambiguous
`MLD_ERR_FAIL` has a different meaning for different functions. For some functions, the meaning is unclear and possibly ambiguous (e.g. does `verify() == MLD_ERR_FAIL` mean an invalid signature, or can it be an invalid public key?).

At least “signature verification failure” and “invalid key” are externally observable distinct error conditions, and should have distinct error codes (like `MLD_ERR_SIGN_ATTEMPTS_EXHAUSTED` is now a distinct error code). “PCT failure” is not so distinguishable from a pure white-box perspective, but should probably have its own error code too.

From the perspective of an integrator of mldsa-native into a multi-mechanism cryptographic library, I'd like to have a single error code translation function into my library's error codes, but the ambiguity of `MLD_ERR_FAIL` requires extra work.

(I can submit a patch for that quickly, but let me know what's acceptable at this stage of the beta, and if there are any design considerations such as alignment with other pqcp projects.)

hanno-becker:
Previously, MLD_ERR_FAIL was returned for several externally distinct failure conditions: an invalid signature, an invalid secret key, a failed pairwise consistency test (PCT), and invalid arguments such as an unsupported pre-hash algorithm or an over-long context string. An integrator translating mldsa-native return codes into their own error space could not tell these apart without extra work.

This commit adds four dedicated error codes:

  MLD_ERR_INVALID_SIGNATURE (-6)  verification rejected the signature
  MLD_ERR_INVALID_KEY       (-7)  pk_from_sk found the sk inconsistent
  MLD_ERR_PCT_FAIL          (-8)  keygen PCT self-test failed
  MLD_ERR_INVALID_ARG       (-9)  bad pre-hash algorithm or context > 255

The existing codes (-1..-5) are unchanged. In particular, MLD_ERR_FAIL remains, but is currently only used as an internal signal and not externally observable.

The commit also makes the CBMC ensures contracts of the keygen and signing entry points specific instead of using the unspecific MLD_ANY_ERROR.

Resolves: https://github.com/pq-code-package/mldsa-native/issues/1306


hanno-becker:
@gilles-peskine-arm @mkannwischer Let's continue the discussion here.

mldsa/src/common.h
```diff
@@ -276,16 +276,19 @@
...
- * Intended for use in top-level `ensures` clauses that admit every
- * possible error. Narrower contracts should enumerate only the
- * specific errors they can actually return. */
-#define MLD_ANY_ERROR(err)                                                  \
-  ((err) == MLD_ERR_FAIL || (err) == MLD_ERR_OUT_OF_MEMORY ||               \
-   (err) == MLD_ERR_RNG_FAIL || (err) == MLD_ERR_SIGN_ATTEMPTS_EXHAUSTED || \
-   (err) == MLD_ERR_SIGNING_PAUSED)
+/* Signature verification failed: the signature is not valid for the given
+ * message and public key. Returned by the verification API. */
+#define MLD_ERR_INVALID_SIGNATURE (-6)
```
mkannwischer, line 282: I wonder if we should keep `-1` as the return value for invalid signatures as `-1` should not be used for anything else anyway anymore, right?
That way in case some did write `verify() == -1` (which of course you shouldn't do), we don't risk wrongly accepting an invalid signature.

mldsa/mldsa_native.h
```diff
@@ -191,9 +205,16 @@ extern "C"
  *                                         MLD_CONFIG_MAX_SIGNING_ATTEMPTS
  *                                         iterations. Only possible when
  *                                         MLD_CONFIG_KEYGEN_PCT is enabled.
- * @retval MLD_ERR_FAIL                    Other kinds of failure, including
- *                                         PCT failure if
- *                                         MLD_CONFIG_KEYGEN_PCT is enabled.
+ * @retval MLD_ERR_SIGNING_PAUSED          The PCT's signing step was paused by
```
gilles-peskine-arm, line 204: Minor, preexisting: Doxygen doesn't automatically detect macro names. To link a macro to its definition, you need to write `#MLD_ERR_SIGNING_PAUSED`.

mldsa/mldsa_native.h
```diff
@@ -231,7 +252,15 @@ int MLD_API_NAMESPACE(keypair_internal)(
  *                                         MLD_CONFIG_MAX_SIGNING_ATTEMPTS
  *                                         iterations. Only possible when
  *                                         MLD_CONFIG_KEYGEN_PCT is enabled.
```
gilles-peskine-arm: Preexisting: I find it a little weird that a PCT error doesn't return `MLD_ERR_PCT_FAIL`. Is “max signing attempts reached” an actionable condition when deriving a key? I guess when randomly generating a key, a fresh seed is likely to pass the PCT in that case, but not when the PCT uncovers an implementation bug?

mldsa/src/common.h
gilles-peskine-arm: That would make sense, but it's only a very narrow advantage.

An argument against changing the value is that it would break the alignment of error values with mlkem-native, unless users upgrade both in synch. Mind you, what can trigger `MLK_ERR_FAIL` in practice?

gilles-peskine-arm (approved):
Thank you very much for implementing this so quickly! The granularity makes sense to me. I've reviewed the code and documentation updates, but I can't vouch for anything missing due to my lack of familiarity with the project, and I'll take your word (and CMBC's word, AFAICT) that the low-level uses of `MLD_ERR_FAIL` don't filter to public functions,

mldsa/mldsa_native.h
hanno-becker: Thank you! I opened #1313

mldsa/mldsa_native.h
hanno-becker: This is a good point. With PCT, one has two or perhaps three options: A) Mask the internal error and return `PCT_FAIL`. B) Return the internal error. C) Return a combination of both.

@mkannwischer wdyt

mkannwischer: It feels weird to me too to return `MLD_ERR_SIGN_ATTEMPTS_EXHAUSTED`.
I'm fine with returning `PCT_FAIL`, or if there is value in more fine grained error codes one could define sub-types of `PCT_FAIL`.

hanno-becker: It's unclear what sub-types/errors of `PCT_FAIL` would look like. One either needs to chain errors explicitly (like `PCT_FAIL_RNG_FAIL`) or arithmetically -- either way I don't want to rush into it. I think Mbed TLS does this in a principled way with error codes at different levels that can be combined; we could do something like this, too. One could even consider generalizing things and returning a low-level error code -- the original source of failure, e.g. RNG fail -- with higher bits indicating the caller-chain. The library is small enough to allow that granularity. Happy about thoughts.

For this PR, though, I've changed the code to map verification failure and signing attempt exhaustion to PCT fail (suggesting a buggy/corrupted implementation) and to keep other failures hinting at platform issues (e.g. RNG fail or OOM). Please take a look.

gilles-peskine-arm: >  think Mbed TLS does this in a principled way with error codes at different levels that can be combined

It used to, but we stopped. Combining error codes was risk-prone (occasionally we'd end up adding error codes from the same level), made static analysis harder (because it's sometimes hard to figure out whether `err1 + err2` might wrap around and might be 0), and I've yet to see any code relying on it.

> One could even consider generalizing things and returning a low-level error code -- the original source of failure, e.g. RNG fail -- with higher bits indicating the caller-chain.

Tracking the origin of an error can occasionally be useful, but at a higher level (e.g. PKCS#11 distinguishes “out of memory in the secure element” from “out of memory on the host”). A theoretically-not-fatal error during PCT isn't worth complicating the architecture (and I kind of regret raising this now).

hanno-becker: @gilles-peskine-arm Are you happy with the status quo of the PR?

gilles-peskine-arm: Yes!

mldsa/src/sign.c
```diff
@@ -1197,10 +1210,11 @@ int mld_sign_verify_internal(const uint8_t sig[MLDSA_CRYPTO_BYTES],
   mld_polyvecl_unpack_z(z, sig + MLDSA_SIG_Z_OFFSET);
 
   /* mld_polyvecl_chknorm signals failure through a single non-zero error code
-   * that's not yet aligned with MLD_ERR_XXX. Map it to MLD_ERR_FAIL. */
+   * that's not yet aligned with MLD_ERR_XXX. A norm-check failure here means
+   * the signature is invalid, so map it to MLD_ERR_INVALID_SIGNATURE. */
   if (mld_polyvecl_chknorm(z, MLDSA_GAMMA1 - MLDSA_BETA))
```
mkannwischer, line 1219: We may want to consolidate this. I have opened https://github.com/pq-code-package/mldsa-native/issues/1316 so we don't forget.

mkannwischer:
I clarified the documentation regarding PCT failures and tightened the tests expecting non-zero return values. @hanno-becker, could you double check if I got that right, please?

mkannwischer (approved):
Thanks @hanno-becker for adjusting this, and @gilles-peskine-arm for reviewing!
This looks good to me. I'll add this to the v1.0.0 release notes shortly.

hanno-becker:
@mkannwischer Yes this looks good! Did you check if there are more cases where we should tighten tests?

PR #1310, opened 2026-07-23
Backends: Namespace assembly local labels with mld_ prefix

hanno-becker:
Previously, local labels in the assembly backends were prefixed only by the name of the enclosing function, e.g. Lntt_layer123_start in ntt_aarch64_asm.S. Global entry symbols are namespaced per project via MLD_ASM_NAMESPACE(), but local labels are emitted as literal text and survive verbatim into the autogenerated mldsa/*.S files.

mldsa-native derives its assembly from mlkem-native and kept the same function and label names, so its local labels are byte-for-byte identical (Lntt_layer123_start, Lrej_uniform_loop48, ...). This is harmless as long as each backend is assembled into its own object file, but breaks when both backends are merged into a single link unit: AWS-LC's FIPS build concatenates all sources and runs them through the delocate tool, which flattens them into one symbol namespace and panics on the duplicate labels.

This commit adds a fixed mld_ project prefix to every local label across all backends (AArch64 arithmetic and FIPS202, x86_64 AVX2, Armv8.1-M MVE), not just the ones that clash today, so mldsa-native emits Lmld_ntt_layer123_start and can no longer collide with the mlk_-prefixed labels emitted by mlkem-native, independent of the integrator or platform.

The change is made in the dev/ assembly sources and propagated to mldsa/*.S and the proofs/hol_light/* drops via scripts/simpasm. Local labels emit no object-code bytes, and branch targets encode PC-relative offsets rather than names, so the assembled .text is byte-identical; the byte-locked HOL Light and CBMC proofs are unaffected.

check_asm_loop_labels in scripts/autogen is updated to require the mld_-prefixed function-derived prefix, so a newly added un-namespaced label fails CI.


scripts/autogen
```diff
@@ -2484,7 +2484,15 @@ def check_asm_loop_labels_for_file(filename):
...
+    # Local labels are additionally namespaced with a fixed `mld_` project
+    # prefix. Function/entry symbols are namespaced at preprocessing time via
+    # MLD_ASM_NAMESPACE(), but local labels are emitted as literal text and
+    # survive verbatim into the autogenerated .S files. Without a per-project
+    # prefix they clash with the identically-named labels in mlkem-native's
+    # assembly (from which mldsa-native's is derived) when both backends are
+    # merged into a single link unit (e.g. AWS-LC's FIPS `delocate` step). See
+    # the `mld_` prefix below.
+    lbl_prefix = "mld_" + re.sub(r"(_(aarch64|avx2|mve))?_asm$", "", funcname) + "_"
     lbl_pattern = r"^(\w+):"
```
mkannwischer: I believe this misses indented labels. May be worth fixing as those are used, e.g., int he MVE backend.

mkannwischer (changes requested):
Thanks @hanno-becker - this looks great! I have one commentregarding the linting.

scripts/autogen
hanno-becker: Done

hanno-becker:
Marking v1 because it enables a cleaner import into AWS-LC.

mkannwischer (approved):
Thanks! LGTM

PR #1311, opened 2026-07-23
liboqs: Declare derandomized keypair in integration META

mkannwischer:
Expose ML-DSA's seed-based keygen (@[FIPS204, Algorithm 6, ML-DSA.KeyGen_internal]) to the liboqs SIG derand API by declaring it in the per-level META files.

See open-quantum-safe/liboqs#2476.


mkannwischer:
Holding back until consensus is reached in https://github.com/open-quantum-safe/liboqs/pull/2476

mkannwischer:
There is currently no consensus downstream if liboqs should support a derandomized keypair. I'm closing this for now and will reopen it in case it is needed.

PR #1312, opened 2026-07-24
CI: Disable ABI checks on Zephyr targets

bremoran:
## Summary

- add a default-enabled `abicheck` input to the `functest` composite action;
- set `abicheck: false` for Zephyr functional-test jobs.

This follows the same action-input pattern used for the other optional test
suites. Existing `functest` callers retain ABI checking by default.

## Why

The current ABI checker is not built as a Zephyr application. Normal Zephyr
tests go through the Zephyr CMake build, but `abicheck` uses the generic Make
rules and derives its architecture from the CI host. On `ubuntu-latest`, that
produces an x86_64 executable which the Zephyr wrapper then passes to
`qemu-system-arm`.

On #1277, all preceding Cortex-M55 functional, KAT, ACVP, and sign-hook tests
passed before that host ABI executable reached QEMU and hard-locked at `PC=0`:

- historical failure: https://github.com/pq-code-package/mldsa-native/actions/runs/30073952185/job/89422620672
- fresh reproduction on unchanged head `655be044b8450b2713a5cf24ab8dd0bf53a8b847`: https://github.com/pq-code-package/mldsa-native/actions/runs/30107040830

Only M55 reached this path because it is the only Zephyr matrix target using
`opt=all`; ABI checking is enabled only for optimized builds.

This change makes the current Zephyr exclusion explicit without affecting the
standalone ABI-check command or other `functest` users. Supporting ABI checking
as a Zephyr application can be handled separately from the Armv8.1-M Keccak
backend work in #1277.

## Validation

- repository-pinned `actionlint`: passed
- `./scripts/tests all --help`: confirms `--abicheck` / `--no-abicheck`
- `git diff --check`: passed
- exact head: `059167a7dbf4d7bca518732a8070c7361177e7d5`
- exact-head Zephyr run: https://github.com/pq-code-package/mldsa-native/actions/runs/30282130963 (all five targets passed)
- Cortex-M55 job: https://github.com/pq-code-package/mldsa-native/actions/runs/30282130963/job/90030891843 (`abicheck: false` expanded to `--no-abicheck`; `extra_args` remained `--no-auto`; functional, KAT, ACVP, sign-hook, and smoke tests passed)
- exact-head PR CI: https://github.com/pq-code-package/mldsa-native/actions/runs/30282114266 (in progress with no failures at the time of this update)

### DIFF
--- a/.github/workflows/zephyr.yml
+++ b/.github/workflows/zephyr.yml
@@ -41,6 +41,7 @@ jobs:
           stack: false
           alloc: false
           rng_fail: false
+          abicheck: false
           check_namespace: false
           # Zephyr's CMake selects the target arch; disable the host-arch
           # auto-detection that would otherwise leak into the forwarded CFLAGS.
--- a/.github/actions/functest/action.yml
+++ b/.github/actions/functest/action.yml
@@ -69,6 +69,9 @@ inputs:
   rng_fail:
     description: Determine whether to run rng fail tests or not
     default: "true"
+  abicheck:
+    description: Determine whether to run ABI compliance tests or not
+    default: "true"
   extra_args:
     description: Additional arguments to pass to the tests script
     default: ""
@@ -93,6 +96,7 @@ runs:
           echo ALLOC="${{ inputs.alloc == 'true' && 'alloc' || 'no-alloc' }}" >> $GITHUB_ENV
           echo SIGN_HOOK="${{ inputs.sign_hook == 'true' && 'sign-hook' || 'no-sign-hook' }}" >> $GITHUB_ENV
           echo RNGFAIL="${{ inputs.rng_fail == 'true' && 'rng-fail' || 'no-rng-fail' }}" >> $GITHUB_ENV
+          echo ABICHECK="${{ inputs.abicheck == 'true' && 'abicheck' || 'no-abicheck' }}" >> $GITHUB_ENV
       - name: Setup nix
         uses: ./.github/actions/setup-shell
         with:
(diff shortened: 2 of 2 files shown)


hanno-becker (commented):
_Why_ does the ABI checker fail? There's no reason it couldn't run on Zephyr targets, AFAIU.

.github/workflows/zephyr.yml
```diff
@@ -44,7 +44,7 @@ jobs:
           check_namespace: false
           # Zephyr's CMake selects the target arch; disable the host-arch
           # auto-detection that would otherwise leak into the forwarded CFLAGS.
-          extra_args: --no-auto
+          extra_args: --no-auto --no-abicheck
```
mkannwischer: Other tests are disabled via an argument to `functest`, so this is somewhat inconsistent.
Could you please use an Github action argument instead?

mkannwischer (changes requested):

bremoran:
> _Why_ does the ABI checker fail? There's no reason it couldn't run on Zephyr targets, AFAIU.

You’re right: there is no fundamental reason the ABI checker could not run on Zephyr. The issue is that it is not currently integrated as a Zephyr application.

Normal Zephyr tests are built as Arm firmware through Zephyr CMake. abicheck, however, follows the generic Make path and derives its architecture from the CI host. On ubuntu-latest, it therefore builds an x86_64 executable. The Zephyr wrapper then passes that executable to qemu-system-arm, which results in a HardFault at PC=0.

Only M55 encounters this because it is the sole Zephyr matrix target using opt=all; ABI checking runs only for optimized builds. The other targets use no_opt and never enter this path.

Because the current setup does not provide valid ABI-check coverage on Zephyr, I disabled it there as an issue orthogonal to the x1 optimization work in #1277. I also agree with @mkannwischer's  consistency point: the exclusion should be expressed as a dedicated functest action argument rather than through extra_args.

A working Zephyr ABI checker would require us to:

1. Select armv81m from the Zephyr target rather than the CI host.
2. Build the ABI driver, call stub, self-tests, generated checks, and implementation assembly through Zephyr CMake.
3. Preserve the per-source Arm/MVE feature flags.
4. Integrate the checker’s entry point, output, and exit status with the Zephyr shim.
5. Confirm that the checker’s deliberate register-corruption self-tests fail as expected before relying on kernel results.

I agree this would be useful. Could you clarify whether Zephyr-native ABI checking is a prerequisite for merging #1277? If so, we will prioritize it as such; otherwise, I suggest tracking it as a separate test-infrastructure follow-up.

.github/workflows/zephyr.yml
bremoran: Thanks — addressed in `059167a7d`. The `functest` action now exposes a default-enabled `abicheck` input and translates it to `--abicheck` / `--no-abicheck`, consistently with the other test selectors. The Zephyr workflow sets `abicheck: false`; `extra_args` is back to only `--no-auto`.

The [exact-head Zephyr run](https://github.com/pq-code-package/mldsa-native/actions/runs/30282130963) passed all five targets. In the [M55 job](https://github.com/pq-code-package/mldsa-native/actions/runs/30282130963/job/90030891843), the log shows `abicheck: false`, `extra_args: --no-auto`, and the resulting command contains `--no-abicheck`. Full PR CI is [still running](https://github.com/pq-code-package/mldsa-native/actions/runs/30282114266) without failures.

bremoran:
Superseded by #1318, which runs ABI checks as Zephyr applications instead of disabling them. Closing in favor of #1318.

PR #1345, opened 2026-08-03
lowram: Keep w1 in packed form during signature generation

mkannwischer:
The signing attempt held w1 as a full mld_polyveck, but w1 is only needed as the w1Encode input to H and, coefficient-wise, as the a1 argument of MakeHint. Both work from the packed encoding, so Decompose and w1Encode now run one polynomial at a time into a MLDSA_K * MLDSA_POLYW1_PACKEDBYTES buffer, and mld_pack_sig_h recovers each row with the new mld_polyw1_unpack.

With w1 gone, the matrix-vector scratch is the sole remaining user of the buffer the two shared. In REDUCE_RAM mode that scratch is a single polynomial rather than a polyvecl, and it now shares storage with z, so the buffer disappears.

Signing allocation in bytes:

   | |      default |           REDUCE_RAM |
| --- | ---- | ----|
|  ML-DSA-44 |   44704 -> 44448  |   13120 ->  9792 |
| ML-DSA-65   | 69312 -> 68032    | 17248 -> 11872 | 
|  ML-DSA-87   | 108224 -> 107200  |  21344 -> 14176 |

mld_polyveck_decompose and mld_polyveck_pack_w1 have no other callers and are replaced by the fused mld_polyveck_decompose_pack_w1.


jakemas:
Interesting! What motivated this idea? Was it just from poking around in the area?

mkannwischer:
> Interesting! What motivated this idea? Was it just from poking around in the area?

The motivation is to reduce memory consumption during signing. For many consumers that's the biggest issue for ML-DSA so every KiB we can save is going to make someone more happy.

The idea isn't new. We've done that on embedded platforms for a while, see https://eprint.iacr.org/2022/323. 
I was curious what the performance impact would be if we implemented it unconditionally and I'm happy to see that it seems to be negligible.

I still have to go through the implementation to see if something can be simplified. I'd be very happy to see this merged for v2.1 though. If this makes it into the release, we should also revisit keygen as that then needs more memory than signing which is only because I stopped optimizing after it was below signing memory.

mldsa/src/packing.h
```diff
@@ -107,9 +107,12 @@ __contract__(
...
  * comment in mld_attempt_signature_generation (sign.c).}
  *
- * @param[in,out] sig Byte array containing signature.
- * @param[in]     w0  Pointer to low part of input vector.
- * @param[in]     w1  Pointer to high part of input vector.
+ * @param[in,out] sig       Byte array containing signature.
+ * @param[in]     w0        Pointer to low part of input vector.
+ * @param[in]     w1_packed Bit-packed high part of input vector, as produced
+ *                          by mld_polyw1_pack.
+ * @param[out]    w1        Scratch polynomial holding the unpacked row of the
```
hanno-becker: Calling this `w1` is a bit confusing -- this is an implementation detail. My first reaction was confusion "What is the index of this specific entry? Why do I need just one, but all of the [packed] rest?"

Perhaps just `tmp` or `scratch`?

mldsa/src/poly_kl.h
```diff
@@ -364,4 +364,34 @@ __contract__(
 }
 #endif /* !MLD_CONFIG_NO_SIGN_API || !MLD_CONFIG_NO_VERIFY_API */
 
+#if !defined(MLD_CONFIG_NO_SIGN_API)
+#define mld_polyw1_unpack MLD_NAMESPACE_KL(polyw1_unpack)
+/**
+ * Unpack polynomial w1. Inverse of mld_polyw1_pack; dispatches to the
+ * value-specialized variant for the selected parameter set.
+ *
+ * @spec{Implements @[FIPS204, Algorithm 18, SimpleBitUnpack].}
```
hanno-becker: Might be worth specifying which values of `b` are covered.

mldsa/src/poly.h
```diff
@@ -461,4 +461,51 @@ __contract__(
...
 #endif /* !MLD_CONFIG_NO_SIGN_API || !MLD_CONFIG_NO_VERIFY_API */
 
+#if !defined(MLD_CONFIG_NO_SIGN_API)
+#if defined(MLD_CONFIG_MULTILEVEL_WITH_SHARED) || MLD_CONFIG_PARAMETER_SET == 44
+#define mld_polyw1_unpack_88 MLD_NAMESPACE(polyw1_unpack_88)
+/**
+ * Unpack polynomial w1 packed with 6 bits per coefficient.
+ * This is the variant for parameter sets with MLDSA_GAMMA2 = (MLDSA_Q-1)/88
+ * (ML-DSA-44). Inverse of mld_polyw1_pack_88.
+ *
```
hanno-becker, line 471: Add FIPS-204 reference?

mldsa/src/poly.h
```diff
@@ -461,4 +461,51 @@ __contract__(
...
+);
+#endif /* MLD_CONFIG_MULTILEVEL_WITH_SHARED || MLD_CONFIG_PARAMETER_SET == 44 \
+        */
+
+#if defined(MLD_CONFIG_MULTILEVEL_WITH_SHARED) || \
+    (MLD_CONFIG_PARAMETER_SET == 65 || MLD_CONFIG_PARAMETER_SET == 87)
+#define mld_polyw1_unpack_32 MLD_NAMESPACE(polyw1_unpack_32)
+/**
+ * Unpack polynomial w1 packed with 4 bits per coefficient.
+ * This is the variant for parameter sets with MLDSA_GAMMA2 = (MLDSA_Q-1)/32
```
hanno-becker, line 494: Add FIPS 204 reference?

mldsa/src/poly_kl.h
```diff
@@ -364,4 +364,34 @@ __contract__(
...
+ */
+static MLD_INLINE void mld_polyw1_unpack(
+    mld_poly *r, const uint8_t a[MLDSA_POLYW1_PACKEDBYTES])
+__contract__(
+  requires(memory_no_alias(r, sizeof(mld_poly)))
+  requires(memory_no_alias(a, MLDSA_POLYW1_PACKEDBYTES))
+  assigns(memory_slice(r, sizeof(mld_poly)))
+  /* One coefficient per 8 * MLDSA_POLYW1_PACKEDBYTES / MLDSA_N bits. */
+  ensures(array_bound(r->coeffs, 0, MLDSA_N, 0,
+                      1 << (8 * MLDSA_POLYW1_PACKEDBYTES / MLDSA_N)))
```
hanno-becker: Is it worth introducing a `#define` for the packed bitwidth of `w1`? That would also slightly clarify the origin of the currently somewhat ad-hoc
```c
#define MLDSA_POLYW1_PACKEDBYTES_88 192
#define MLDSA_POLYW1_PACKEDBYTES_32 128
```

mldsa/src/polyvec.h
```diff
@@ -266,60 +266,50 @@ __contract__(
...
  * -ALPHA/2 < a0 <= ALPHA/2 except a1 = (MLDSA_Q-1)/ALPHA where we set
  * a1 = 0 and -ALPHA/2 <= a0 = a mod MLDSA_Q - MLDSA_Q < 0. Assumes
- * coefficients to be standard representatives.
+ * coefficients to be standard representatives. The low bits replace the input
+ * vector, the high bits are bit-packed into r.
  *
- * @reference{The reference implementation has the input polynomial as a
- * separate argument that may be aliased with either of the outputs. Removing
- * the aliasing eases CBMC proofs.}
+ * Decomposing and encoding are fused so that the high bits are never held as a
```
hanno-becker: Nit: Unify language so we consistently use encode or pack? I think mostly it's "pack"?

mldsa/src/polyvec.h
```diff
@@ -266,60 +266,50 @@ __contract__(
...
+ * vector of polynomials: each one is packed through scratch as it is produced.
```
hanno-becker: The implementation-specific remark "each one is packed through scratch as it is produced." is slightly confusing here. I'd drop it. You say below what the `scratch` parameter is for, and at that point, it's clear you're referring to a parameter. At the present point, it may be unclear.

mldsa/src/polyvec.h
```diff
@@ -266,60 +266,50 @@ __contract__(
...
+ * This saves MLDSA_K * sizeof(mld_poly) bytes, and costs nothing, as the packed
+ * form is all that a caller needs for both hashing and MakeHint.
```
hanno-becker: This is not entirely true? You do need to unpack w1 one polynomial at a time during MakeHint. 

Suggestion:
```
 * Decomposing and encoding are fused so that the high bits are never held as a
 * vector of polynomials, saving MLDSA_K * sizeof(mld_poly) bytes. During MakeHint,
 * a single entry of w1 is instead unpacked on demand.
```

mldsa/src/polyvec_lazy.h
```diff
@@ -569,33 +570,30 @@ __contract__(
...
+ * sampling of A[k][l].
  *
- * @param[out]    w       Pointer to output vector.
- * @param[in,out] mat     Pointer to input matrix.
- * @param[in]     y       Pointer to y seed/kappa.
- * @param[out]    scratch Scratch (only &scratch->vec[0] used).
+ * @param[out]    w     Pointer to output vector.
+ * @param[in,out] mat   Pointer to input matrix.
+ * @param[in]     y     Pointer to y seed/kappa.
+ * @param[out]    y_ntt Scratch for the NTT of the current column of y.
```
hanno-becker: Nit: If this is merely a scratch parameter, I'd keep calling it `scratch` (the "for the NTT of the current column of y" addition is fine though).

mldsa/src/polyvec_lazy.h
```diff
@@ -569,33 +570,30 @@ __contract__(
...
  */
 MLD_INTERNAL_API
 void mld_polyvec_matrix_pointwise_montgomery_yvec_lazy(mld_polyveck *w,
                                                        mld_polymat_lazy *mat,
                                                        const mld_yvec_lazy *y,
-                                                       mld_polyvecl *scratch)
+                                                       mld_poly *y_ntt)
```
hanno-becker: In line with `mld_polymat_lazy` and `mld_yvec_lazy`, I think this should be called `mld_yvec_scratch_lazy`? One could otherwise reasonably assume that arguments without `_lazy` qualification are stable across lazy/eager.

hanno-becker: Similarly, one would adjust the eager signature.

mldsa/src/sign.c
```diff
@@ -761,20 +761,19 @@ __contract__(
...
   mld_polyveck_caddq(w0);
-  mld_polyveck_decompose(w1, w0);
-
-  /* @[FIPS204, Algorithm 7, line 15] ctilde <- H(mu || w1Encode(w1), lambda/4).
-   * w1Encode(w1) is packed into the w1 region of sig (mld_polyveck_pack_w1),
-   * then absorbed by H together with mu. */
-  mld_polyveck_pack_w1(sig, w1);
+  mld_polyveck_decompose_pack_w1(w1_packed, w0, z);
 
-  mld_H(challenge_bytes, MLDSA_CTILDEBYTES, mu, MLDSA_CRHBYTES, sig,
```
hanno-becker, line 777: Incidentally, this reduces writes to the external `sig` buffer, which is a good thing.

mldsa/src/sign.c
```diff
@@ -731,28 +731,28 @@ __contract__(
...
   typedef union
   {
-    mld_polyveck w1;
-    mld_polyvecl tmp;
-  } w1tmp_u;
-  mld_polyveck *w1;
-  mld_polyvecl *tmp;
+    mld_yvec_scratch matvec;
+    mld_poly z;
+  } zmatvec_u;
```
hanno-becker: Nit: Both fields in this union are scratch buffers, and for reasoning about the high-level signing logic one does not need to know anything about how they are used. I'd suggest just calling this `scratch_u` with fields `y` and `w`, and the pointers `matvec` -> `scratch_yvec` and `z` -> `scratch_w1` or similar.

mldsa/src/sign.c
```diff
@@ -761,20 +761,19 @@ __contract__(
...
 
   /* @[FIPS204, Algorithm 7, line 13] w1 <- HighBits(w), here together with the
-   * low part: Decompose yields w = 2*GAMMA2*w1 + w0, keeping both w1 and w0
-   * (w0 is reused below in the line-21/26 alternative, see further down). */
+   * low part: Decompose yields w = 2*GAMMA2*w1 + w0. w0 is kept in full for
+   * the line-21/26 alternative below; w1 only in its w1Encode form, which is
+   * all that line 15 and the hint computation need.
+   *
+   * @[FIPS204, Algorithm 7, line 15] ctilde <- H(mu || w1Encode(w1), lambda/4),
+   * computed one polynomial at a time with z holding w1[k]. */
```
hanno-becker: The "one polynomial at a time" is a bit confusing here. The packed form of `w1` is used as an opaque blob in `mld_H`, isn't it?

Also, should this comment be moved to just before `mld_H(...)`?

hanno-becker (changes requested):
This is an interesting and impactful optimization, thank you for proposing it @mkannwischer. 

Overall, I think this is worth implementing.

However, we are diverging more and more from the spec and reference implementation here. Especially with the lazy/eager split for yvec, we have to take utmost care to keep the documentation of both source and its relation to FIPS-204 accurate and crisp in terms of providing exactly the information that is needed to reason about the code at any point -- and nothing more to, to avoid unnecessary cognitive/reasoning load. I left a few comments in this direction.

hanno-becker:
> We've done that on embedded platforms for a while, see https://eprint.iacr.org/2022/323.

We should cite the relevant papers.

mldsa/src/packing.h
mkannwischer: Changed to scratch.

mldsa/src/poly_kl.h
mkannwischer: Added, but we don't specify it for the packing either.

mldsa/src/poly.h
mkannwischer: added

mldsa/src/poly.h
mkannwischer: added

mldsa/src/poly_kl.h
mkannwischer: added

mldsa/src/polyvec.h
mkannwischer: yes, lets use packing. changed

mldsa/src/polyvec.h
mkannwischer: dropped.

mldsa/src/polyvec.h
mkannwischer: thx. changed.

mldsa/src/polyvec_lazy.h
mkannwischer: yes, lets use the convention everywhere. changed.

mldsa/src/polyvec_lazy.h
mkannwischer: aligned.

mldsa/src/sign.c
mkannwischer: trimmed and moved.

mldsa/src/polyvec_lazy.h
```diff
@@ -502,21 +511,20 @@ __contract__(
  * @param[out] scratch Scratch polyvecl for NTT'd copy of y.
  */
 MLD_INTERNAL_API
-void mld_polyvec_matrix_pointwise_montgomery_yvec_eager(mld_polyveck *w,
-                                                        mld_polymat_eager *mat,
-                                                        const mld_yvec_eager *y,
-                                                        mld_polyvecl *scratch)
+void mld_polyvec_matrix_pointwise_montgomery_yvec_eager(
+    mld_polyveck *w, mld_polymat_eager *mat, const mld_yvec_eager *y,
+    mld_yvec_scratch_eager *scratch)
```
hanno-becker, line 516: Have you considered whether the scratch can be embedded into `mld_yvec`, which anyway is an abstract type? Needs some checking of lifetimes in `sign.c`.

This would simplify the interface.

mkannwischer: I don't see a clean way to do that. It would require a union inside of `mld_yvec` (or pay extra 1 KiB), and you would have to change `mld_compute_pack_z`. This may lead to issues with the CBMC proofs.
I'd prefer to keep it as is.

mldsa/src/poly.h
```diff
@@ -461,4 +461,55 @@ __contract__(
...
+ * @param[in]  a Byte array with bit-packed polynomial.
+ */
+MLD_INTERNAL_API
+void mld_polyw1_unpack_88(mld_poly *r,
+                          const uint8_t a[MLDSA_POLYW1_PACKEDBYTES_88])
+__contract__(
+  requires(memory_no_alias(r, sizeof(mld_poly)))
+  requires(memory_no_alias(a, MLDSA_POLYW1_PACKEDBYTES_88))
+  assigns(memory_slice(r, sizeof(mld_poly)))
+  ensures(array_bound(r->coeffs, 0, MLDSA_N, 0, 1 << 6))
```
hanno-becker: MLDSA_POLYW1_PACKED_BITS_88 ?

mldsa/src/poly.h
```diff
@@ -461,4 +461,55 @@ __contract__(
...
+ * @param[in]  a Byte array with bit-packed polynomial.
+ */
+MLD_INTERNAL_API
+void mld_polyw1_unpack_32(mld_poly *r,
+                          const uint8_t a[MLDSA_POLYW1_PACKEDBYTES_32])
+__contract__(
+  requires(memory_no_alias(r, sizeof(mld_poly)))
+  requires(memory_no_alias(a, MLDSA_POLYW1_PACKEDBYTES_32))
+  assigns(memory_slice(r, sizeof(mld_poly)))
+  ensures(array_bound(r->coeffs, 0, MLDSA_N, 0, 1 << 4))
```
hanno-becker: MLDSA_POLYW1_PACKED_BITS_32 ?

mldsa/src/polyvec_lazy.c
```diff
@@ -231,18 +230,12 @@ void mld_polyvec_matrix_pointwise_montgomery_row_lazy(mld_poly *t_row,
 
 #if !defined(MLD_CONFIG_NO_SIGN_API)
 MLD_INTERNAL_API
-void mld_polyvec_matrix_pointwise_montgomery_yvec_lazy(mld_polyveck *w,
-                                                       mld_polymat_lazy *mat,
```
hanno-becker, line 235: ISTM that the introduction of `mld_yvec_scratch_{lazy/eager}` is a preparatory refactor that can stand alone. Seeing the large complexity of the commit already, can you please hoist this into a separate commit?

mldsa/src/poly.c
```diff
@@ -1057,6 +1057,57 @@ void mld_polyw1_pack_32(uint8_t r[MLDSA_POLYW1_PACKEDBYTES_32],
...
+MLD_INTERNAL_API
+void mld_polyw1_unpack_88(mld_poly *r,
+                          const uint8_t a[MLDSA_POLYW1_PACKEDBYTES_88])
+{
+  unsigned int i;
+
+  for (i = 0; i < MLDSA_N / 4; ++i)
+  __loop__(
+    invariant(i <= MLDSA_N/4)
+    invariant(array_bound(r->coeffs, 0, i*4, 0, 1 << 6))
```
hanno-becker: MLDSA_POLYW1_PACKED_BITS_88

mldsa/src/poly.c
```diff
@@ -1057,6 +1057,57 @@ void mld_polyw1_pack_32(uint8_t r[MLDSA_POLYW1_PACKEDBYTES_32],
...
+MLD_INTERNAL_API
+void mld_polyw1_unpack_32(mld_poly *r,
+                          const uint8_t a[MLDSA_POLYW1_PACKEDBYTES_32])
+{
+  unsigned int i;
+
+  for (i = 0; i < MLDSA_N / 2; ++i)
+  __loop__(
+    invariant(i <= MLDSA_N/2)
+    invariant(array_bound(r->coeffs, 0, i*2, 0, 1 << 4))
```
hanno-becker: MLDSA_POLYW1_PACKED_BITS_32

mldsa/src/poly.c
```diff
@@ -1057,6 +1057,57 @@ void mld_polyw1_pack_32(uint8_t r[MLDSA_POLYW1_PACKEDBYTES_32],
...
+
+#if defined(MLD_CONFIG_MULTILEVEL_WITH_SHARED) || \
+    (MLD_CONFIG_PARAMETER_SET == 65 || MLD_CONFIG_PARAMETER_SET == 87)
+MLD_INTERNAL_API
+void mld_polyw1_unpack_32(mld_poly *r,
+                          const uint8_t a[MLDSA_POLYW1_PACKEDBYTES_32])
+{
+  unsigned int i;
+
+  for (i = 0; i < MLDSA_N / 2; ++i)
```
hanno-becker: Inconsistent use of `++i` vs prevailing `i++` in the rest of the code base

mldsa/src/poly.c
```diff
@@ -1057,6 +1057,57 @@ void mld_polyw1_pack_32(uint8_t r[MLDSA_POLYW1_PACKEDBYTES_32],
...
 
+#if !defined(MLD_CONFIG_NO_SIGN_API)
+#if defined(MLD_CONFIG_MULTILEVEL_WITH_SHARED) || MLD_CONFIG_PARAMETER_SET == 44
+MLD_INTERNAL_API
+void mld_polyw1_unpack_88(mld_poly *r,
+                          const uint8_t a[MLDSA_POLYW1_PACKEDBYTES_88])
+{
+  unsigned int i;
+
+  for (i = 0; i < MLDSA_N / 4; ++i)
```
hanno-becker: Inconsistent use of `++i` vs prevailing `i++` in the rest of the code base

mldsa/src/params.h
```diff
@@ -16,8 +16,12 @@
 
 #define MLDSA_GAMMA2_88 ((MLDSA_Q - 1) / 88)
 #define MLDSA_GAMMA2_32 ((MLDSA_Q - 1) / 32)
-#define MLDSA_POLYW1_PACKEDBYTES_88 192
-#define MLDSA_POLYW1_PACKEDBYTES_32 128
+/* Bits per w1 coefficient: bitlen((MLDSA_Q-1)/(2*MLDSA_GAMMA2) - 1), i.e.
+ * bitlen(43) for MLDSA_GAMMA2_88 and bitlen(15) for MLDSA_GAMMA2_32. */
+#define MLDSA_POLYW1_PACKED_BITS_88 6
+#define MLDSA_POLYW1_PACKED_BITS_32 4
```
hanno-becker: Introducing those is slightly inconsistent with existing packing routines, which hardcode their bit widths. E.g. `mld_polyt1_[un]pack`.

mldsa/src/poly.c
```diff
@@ -1057,6 +1057,57 @@ void mld_polyw1_pack_32(uint8_t r[MLDSA_POLYW1_PACKEDBYTES_32],
           || MLD_CONFIG_PARAMETER_SET == 87 */
 #endif /* !MLD_CONFIG_NO_SIGN_API || !MLD_CONFIG_NO_VERIFY_API */
 
+#if !defined(MLD_CONFIG_NO_SIGN_API)
+#if defined(MLD_CONFIG_MULTILEVEL_WITH_SHARED) || MLD_CONFIG_PARAMETER_SET == 44
+MLD_INTERNAL_API
+void mld_polyw1_unpack_88(mld_poly *r,
```
hanno-becker, line 1063: I realize that those have no use prior to the final optimization, but can you still hoist the introduction of the `w1` unpacking routines and their CBMC proofs into a separate commit?

hanno-becker (changes requested):
I am unhappy with the complexity creep, but have so far not found any approach to avoiding it.

In the meantime, some notes:
* Can you split the work into multiple commits? As indicated in the comments, you can at least hoist the yvec scratch refactor and the w1_unpack changes into standalone commits.
* The use of `MLDSA_POLYW1_PACKED_BITS_XX` is inconsistent, and its presence inconsistent with other pack bit width, e.g. for `t1`.

mldsa/src/poly.c
mkannwischer: ++i is prevailing actually, but let me standardize on i++ everywhere.

mldsa/src/poly.c
mkannwischer: Done.

mldsa/src/params.h
mkannwischer: I've also introduced the bit widths for the other routines in the final commit.
Do you think that's better?

mldsa/src/poly.c
mkannwischer: ++i is prevailing actually, but let me standardize on i++ everywhere.

mldsa/src/poly.c
mkannwischer: Fixed.

mldsa/src/poly.c
mkannwischer: Fixed.

mldsa/src/polyvec_lazy.c
mkannwischer: Moved to its own commit.

mldsa/src/poly.h
mkannwischer: Fixed.

mldsa/src/poly.h
mkannwischer: Fixed.

mldsa/src/params.h
```diff
@@ -32,6 +36,7 @@
 #define MLDSA_CTILDEBYTES 32
 #define MLDSA_POLYZ_PACKEDBYTES 576
 #define MLDSA_POLYW1_PACKEDBYTES MLDSA_POLYW1_PACKEDBYTES_88
+#define MLDSA_POLYW1_PACKED_BITS MLDSA_POLYW1_PACKED_BITS_88
```
hanno-becker: Nit, not a blocker: `PACKEDBYTES` vs `PACKED_BITS`

mldsa/src/polyvec.c
```diff
@@ -365,54 +365,34 @@ uint32_t mld_polyveck_chknorm(const mld_polyveck *v, int32_t bound)
...
-                     array_bound(v0->vec[k3].coeffs, 0, MLDSA_N, 0, MLDSA_Q)))
+                     array_abs_bound(v0->vec[k1].coeffs, 0, MLDSA_N, MLDSA_GAMMA2+1)))
+    invariant(forall(k2, i, MLDSA_K,
+                     array_bound(v0->vec[k2].coeffs, 0, MLDSA_N, 0, MLDSA_Q)))
     decreases(MLDSA_K - i)
   )
   {
-    mld_poly_decompose(&v1->vec[i], &v0->vec[i]);
+    mld_poly_decompose(scratch, &v0->vec[i]);
+    mld_polyw1_pack(&r[i * MLDSA_POLYW1_PACKEDBYTES], scratch);
```
hanno-becker, line 389: Do we need to retain `mld_poly_decompose` as a standalone function or can we similarly introduce `mld_poly_decompose_pack_w1` in its place?

mldsa/src/params.h
```diff
@@ -66,15 +66,19 @@
...
 #endif /* MLD_CONFIG_PARAMETER_SET == 87 */
 
-#define MLDSA_POLYT1_PACKEDBYTES 320
-#define MLDSA_POLYT0_PACKEDBYTES 416
+#define MLDSA_POLYT1_PACKED_BITS 10
+
+#define MLDSA_POLYT1_PACKEDBYTES (MLDSA_N * MLDSA_POLYT1_PACKED_BITS / 8)
+#define MLDSA_POLYT0_PACKEDBYTES (MLDSA_N * MLDSA_D / 8)
+#define MLDSA_POLYZ_PACKEDBYTES (MLDSA_N * MLDSA_POLYZ_PACKED_BITS / 8)
+#define MLDSA_POLYETA_PACKEDBYTES (MLDSA_N * MLDSA_POLYETA_PACKED_BITS / 8)
```
hanno-becker: Optional nit: There is some use in seeing concrete values rather than symbolic expressions. You could consider keeping the concrete values but using a `/* check-magic: ... */` line to connect them to the symbolic expressions.

mldsa/src/poly.c
hanno-becker: Thank you!

mldsa/src/polyvec.c
mkannwischer: poly_decompose() is part of the native API. We can push it further down the call stack for uniformity, but that won't allow us to eliminate the scratch poly.

hanno-becker (approved):
Thank you @mkannwischer. This is a significant memory reduction. The added complexity is tolerable; as mentioned, I would prefer to hide the scratch object better, but I cannot see a way to do it at the moment.

mldsa/src/polyvec.c
hanno-becker: This can be considered as a follow-up.

mldsa/src/params.h
mkannwischer: our current check-magic logic does support constants with less than 4 digits.
Let's do this as a follow-up.

mldsa/src/params.h
mkannwischer: changing to PACKEDBITS

mldsa/src/params.h
hanno-becker: No, it's that magic constants of 4 or more digits _must_ have an explanation. But you can explain any constant.

mkannwischer: you can explain any constant, but it's not checked. 
```
/* check-magic: 320 == 999 */
```
works just fine. 
It's a bug in the script.
Going to fix it in a follow-up PR.

PR #1354, opened 2026-08-05
mldsa_native.h: Hoist domain separation constants above API guard

### LINKED ISSUE #1353: MLD_CONFIG_CONSTANTS_ONLY hides the constants needed to call prepare_domain_separation_prefix
`MLD_PREHASH_*` and `MLD_DOMAIN_SEPARATION_MAX_BYTES` sit inside the `#if !defined(MLD_CONFIG_CONSTANTS_ONLY)` block in `mldsa_native.h`, so a constants-only caller cannot call `prepare_domain_separation_prefix()`: it can neither size the `prefix` buffer nor name a `hashalg`. Both are constants, which `mldsa_native_config.h` documents the option as still exposing. Moving them above the guard would fix it.

mkannwischer:
MLD_PREHASH_* and MLD_DOMAIN_SEPARATION_MAX_BYTES sat inside the !MLD_CONFIG_CONSTANTS_ONLY block, so a constants-only caller could neither size the prefix buffer for prepare_domain_separation_prefix() nor name a hash algorithm. Move both above the guard.

- Resolves https://github.com/pq-code-package/mldsa-native/issues/1353

### DIFF
--- a/mldsa/mldsa_native.h
+++ b/mldsa/mldsa_native.h
@@ -165,6 +165,24 @@
 #define MLD_API_QUALIFIER
 #endif
 
+/* Hash algorithm constants for domain separation */
+#define MLD_PREHASH_NONE 0
+#define MLD_PREHASH_SHA2_224 1
+#define MLD_PREHASH_SHA2_256 2
+#define MLD_PREHASH_SHA2_384 3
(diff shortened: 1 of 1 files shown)


hanno-becker (approved):
LGTM

PR #1361, opened 2026-08-06
CBMC: Check that outputs are zeroized or unmodified on error

hanno-becker:
* Port of https://github.com/pq-code-package/mlkem-native/pull/1850


rod-chapman (commented):
Should we zeroize prefix[] when mld_prepare_domain_separation_prefix() returns an error and specify that in its ensures clause? The code leaves it unchanged in the case of error. We should at least add a comment to say it's unchanged on error. Nit: if "prefix is unchanged" is the desired behaviour, then its mode should be "in out" since it is possible for its final value to depend on its initial value.

mkannwischer:
> Should we zeroize prefix[] when mld_prepare_domain_separation_prefix() returns an error and specify that in its ensures clause? The code leaves it unchanged in the case of error. We should at least add a comment to say it's unchanged on error. Nit: if "prefix is unchanged" is the desired behaviour, then its mode should be "in out" since it is possible for its final value to depend on its initial value.

The desired behavior is "unchanged or zero" - see https://github.com/pq-code-package/mldsa-native/blob/main/API-CONVENTIONS.md#output-buffers-on-error.

rod-chapman:
You also seem to be missing
```
USE_FUNCTION_CONTRACTS+=mld_zeroize
```
in a few proof Makefiles...

mkannwischer (approved):
Thanks @hanno-becker! LGTM.

PR #1376, opened 2026-08-08
CI: Mirror sibling issues flagged for porting

### LINKED ISSUE #1359: Consider automatically creating mirror issues between mldsa-native and mlkem-native
We often have issues affecting both mlkem-native and mldsa-native, e.g., [this one](https://github.com/pq-code-package/mldsa-native/issues/1358). What usually happens is that we create it in one repo and then maybe create a mirror issue in the other repo, but it is not done consistently. A more convenient way would be to use a Github label, e.g., the pre-existing `needs-mlkem-native-port` and have a bot (e.g., a Github action) create the corresponding mirror issue in the other repo.
There are a lot of prexisting issues that only exist in one of the two repos.

mkannwischer:
Add a scheduled workflow that opens local tracking issues for pq-code-package/mlkem-native items labelled needs-mldsa-native-port: open issues become mirror issues, merged pull requests become port issues.

The job runs in this repository and only writes here, so it needs nothing beyond the default GITHUB_TOKEN. The source link in each generated body both backlinks the sibling item and serves as the deduplication key.

 - Resolves https://github.com/pq-code-package/mldsa-native/issues/1359

### DIFF
--- a/.github/scripts/mirror-issues.py
+++ b/.github/scripts/mirror-issues.py
@@ -1,0 +1,117 @@
+#!/usr/bin/env python3
+# Copyright (c) The mldsa-native project authors
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+
+#
+# CI helper: mirror the sibling repository's issues and merged pull requests
+# flagged as needing a port into $GITHUB_REPOSITORY.
+#
+# Open issues become mirror issues; merged pull requests become port issues.
+# Anything else carrying the label is skipped, so a labelled pull request is
+# picked up by a later run once it merges.
+#
+# This runs in the target repository and only ever writes there, so the default
+# GITHUB_TOKEN suffices; the sibling repository is only read.
+#
+
+import json
+import os
+import re
+import urllib.parse
+import urllib.request
+
+# Sibling repository, and the label marking its issues as needing a port here.
+SOURCE_REPO = "pq-code-package/mlkem-native"
+SOURCE_LABEL = "needs-mldsa-native-port"
+
+# Labels applied to the issues created here, also used to find them again.
+MIRROR_LABEL = "mlkem-native-mirror"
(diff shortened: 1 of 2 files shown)


.github/scripts/mirror-issues.py
```diff
@@ -0,0 +1,129 @@
...
+def still_labelled(entry):
+    """Confirm the label is really still set on the source entry.
+
+    The label-filtered listing is eventually consistent: it keeps returning
+    entries for minutes after the label was removed, while a direct read of the
+    entry is already correct. Without this check, a port that was just marked as
+    no longer needed would still be mirrored.
+    """
+    fresh = request("GET", f"{API}/repos/{SOURCE_REPO}/issues/{entry['number']}")
+    return any(label["name"] == SOURCE_LABEL for label in fresh["labels"])
```
hanno-becker: This seems over-engineered. One has to work off a snapshot of the repository state. If one is unlucky and removes a needs-port label just as the workflow creates the mirror, one has to correct it manually. Re-checking within the script doesn't make sense to me.

hanno-becker (commented):
I like the idea, thank you @mkannwischer. One aspect of the script seems over-engineered, but functionally it's OK either way. Did you have a chance to test it?

.github/scripts/mirror-issues.py
mkannwischer: I hit that case when testing it. Github's index seems sometimes to take many minutes to update after a change. But we can remove it since it's unlikely to matter outside of testing.

mkannwischer: I'd say Github is under-engineered.

mkannwischer: Removed.

hanno-becker (approved):

jakemas (approved):

Issue #1346, opened 2026-08-03
Exposing MLD_ALIGN for allocators and customization


gilles-peskine-arm:
When integrating [`MLD_CONFIG_CUSTOM_ALLOC_FREE` into TF-PSA-Crypto](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/786), we ran into [two interface gaps with `MLD_ALIGN`](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/786/changes#r3505942030). I suspect resolving it properly requires a minor API extension in mldsa-native. I'm not sure what exactly that should be.

Our custom allocator grabs memory from a global workspace, which can be used by a single ML-DSA operation at a time.

1. We want our workspace to be suitably aligned. A global `unsigned char` array might not be suitably aligned. In C11 (which is good enough for us), `_Alignas(max_align_t)` should be fine for any type, but that's not necessarily good enough: if mldsa-native wants a stricter alignment, it might skip some memory at the beginning of the buffer, which means the buffer might not be large enough. E.g. if `_Alignas(max_align_t)` is 16 but `MLD_ALIGN` is 32 and `MLD_TOTAL_ALLOC` is based on 1000 blocks then we might end up with a workspace spanning addresses 16..32016, but mldsa-native would start at address 32 and therefore might overflow to address 32031.

2. The definition of `MLD_ALIGN` is not customizable, and so some variables might end up misaligned on compilers other than GCC-like and MSVC. (Although I _think_ that would only impact performance, since all variables should be suitably aligned for their C type anyway, but I haven't fully reviewed how `MLD_ALIGN_UP` is used in the code.)

The obvious way to resolve (2) would be to allow `MLD_ALIGN` to be configured.

An obvious way to resolve (1) would be to expose `MLD_ALIGN` or `MLD_DEFAULT_ALIGN` in the API, but there might be other sensible choices.

[Currently, in TF-PSA-Crypto](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/786/changes#r3505942030), we're including an internal header for (1) and not yet officially supporting IAR for (2).

hanno-becker:
Thank you @gilles-peskine-arm for raising this. This mirrors https://github.com/pq-code-package/mlkem-native/issues/1386 from mlkem-native.

[...]

PR #1379, opened 2026-08-11
Make the alignment configurable

mkannwischer:
Add two experimental config options: MLD_CONFIG_ALIGN sets the alignment of large buffers and structures in bytes, MLD_CONFIG_ALIGN_ATTRIBUTE replaces the declarator prefix for compilers whose spelling mldsa-native does not know. MLD_DEFAULT_ALIGN, MLD_ALIGN and MLD_ALIGN_UP derive from them.

The x86_64 and AArch64 backends reject an alignment below what their load and store instructions require.

MLD_TOTAL_ALLOC_* is measured for the default alignment, so it is no longer defined once the alignment is customized, and the allocation test skips itself in that case.

- Partially addresses #1346


mldsa/mldsa_native_config.h
```diff
@@ -365,6 +365,47 @@
...
 
+/**
+ * MLD_CONFIG_ALIGN [EXPERIMENTAL]
+ *
+ * Alignment, in bytes, of large buffers and structures.
+ * Must be a power of two. If unset, 32 bytes are used.
+ *
+ * Set this only to change the alignment. Wherever mldsa-native knows a
+ * declarator prefix for the compiler, that prefix follows the new value
+ * automatically; otherwise set MLD_CONFIG_ALIGN_ATTRIBUTE as well.
```
hanno-becker, line 376: It's not clear to me what "herever mldsa-native knows a declarator prefix for the compiler, that prefix follows the new value automatically" means. Can you rephrase?

mldsa/mldsa_native_config.h
```diff
@@ -365,6 +365,47 @@
...
+ *
+ * Native backends require a minimum alignment to avoid faulting on
+ * alignment-sensitive load and store instructions; this is enforced at
+ * compile time.
```
hanno-becker, line 380: It should be clarified if this means a silent alignment bump or a rejection.

mldsa/mldsa_native_config.h
```diff
@@ -365,6 +365,47 @@
...
+ *          stable API in a future version.
+ */
+/* #define MLD_CONFIG_ALIGN 32 */
+
+/**
+ * MLD_CONFIG_ALIGN_ATTRIBUTE [EXPERIMENTAL]
+ *
+ * Declarator prefix aligning the declared object.
+ *
+ * Set this only if mldsa-native knows no spelling for your compiler. Without
```
hanno-becker, line 396: It's unclear what "Set this only if mldsa-native knows no spelling for your compiler."  means.

mldsa/src/sys.h
```diff
@@ -205,10 +205,29 @@
...
 #endif /* restrict */
 
+/*
+ * MLD_DEFAULT_ALIGN: Alignment, in bytes, of large buffers and structures.
+ * Set through MLD_CONFIG_ALIGN.
+ */
+#if defined(MLD_CONFIG_ALIGN)
+#define MLD_DEFAULT_ALIGN MLD_CONFIG_ALIGN
+#else
 #define MLD_DEFAULT_ALIGN 32
```
hanno-becker, line 215: `MLD_DEFAULT_ALIGN` is perhaps no longer the right name?

mldsa/src/native/aarch64/meta.h
```diff
@@ -7,6 +7,12 @@
 #ifndef MLD_NATIVE_AARCH64_META_H
 #define MLD_NATIVE_AARCH64_META_H
 
+/* This backend loads and stores polynomials through Q registers, which fault
+ * on operands that are not 16-byte aligned if alignment checking is on. */
+#if MLD_DEFAULT_ALIGN < 16
+#error Bad configuration: The AArch64 backend requires MLD_CONFIG_ALIGN to be at least 16
```
hanno-becker, line 13: Perhaps say why

mldsa/src/native/x86_64/meta.h
```diff
@@ -7,6 +7,12 @@
 #ifndef MLD_NATIVE_X86_64_META_H
 #define MLD_NATIVE_X86_64_META_H
 
+/* This backend loads and stores polynomials with vmovdqa, which faults on
+ * operands that are not 32-byte aligned. */
+#if MLD_DEFAULT_ALIGN < 32
+#error Bad configuration: The x86_64 backend requires MLD_CONFIG_ALIGN to be at least 32
```
hanno-becker, line 13: Perhaps say why here, too

mldsa/src/native/aarch64/meta.h
```diff
@@ -7,6 +7,12 @@
 #ifndef MLD_NATIVE_AARCH64_META_H
 #define MLD_NATIVE_AARCH64_META_H
 
+/* This backend loads and stores polynomials through Q registers, which fault
+ * on operands that are not 16-byte aligned if alignment checking is on. */
+#if MLD_DEFAULT_ALIGN < 16
```
hanno-becker, line 12: We should also catch the case where we don't know an alignment declarator. In that case, one should arguably refuse to build the backends. One way to do this would be to check first if an alignment declarator is known, and if not, set the alignment to `1`.

hanno-becker (changes requested):
Overall this looks good, thanks @mkannwischer! Some documentation nits in comments. We should also use this opportunity to fix the pre-existing defect that if no alignment declarator is known, we silently build the backends without alignment constraints.

PR #1380, opened 2026-08-11
Consolidate error codes for lower level functions

### LINKED ISSUE #1316: Consolidate error codes for lower level functions
See, e.g, 
https://github.com/pq-code-package/mldsa-native/blob/9a180a96b8574ee54981b955b1003a75dc2c52a3/mldsa/src/sign.c#L1199-L1205

Task
 - Move normalization of error codes into lower level functions where possible

willieyz:
- Resolves: #1316

This PR moves error-code normalization from `mld_sign_verify_internal` into the function that detects the failure, so the caller no longer needs to re-label a generic return value.

`mld_sig_unpack_hints` now returns `MLD_ERR_INVALID_SIGNATURE` instead of `MLD_ERR_FAIL`. Since this function is verify-only, all three of its failure paths indicate malformed hint encoding, which means the signature must be rejected.

`mld_polyvecl_chknorm` keeps its existing `0 / 0xFFFFFFFF` contract. It is also used to validate the secret key in `mld_sign_pk_from_sk`, where `MLD_ERR_INVALID_SIGNATURE` would not be meaningful, and a generic code would still have to be mapped by each caller. Only the corresponding comment is refined.

### DIFF
--- a/mldsa/src/packing.c
+++ b/mldsa/src/packing.c
@@ -156,7 +156,7 @@ int mld_sig_unpack_hints(mld_poly *h, const uint8_t sig[MLDSA_CRYPTO_BYTES],
 
   if (new_hint_count < old_hint_count || new_hint_count > MLDSA_OMEGA)
   {
-    return MLD_ERR_FAIL;
+    return MLD_ERR_INVALID_SIGNATURE;
   }
 
   mld_memset(h, 0, sizeof(mld_poly));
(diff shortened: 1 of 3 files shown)


mkannwischer:
Thanks @willieyz for looking into this!

Thinking about this more, I don't think we want this change.
We cannot set proper return codes in lower level functions in all places as it can depend on context (e.g., chknorm is used in verification and also in signing.) So the only reasonable convention for low level functions is to return 0 or `MLD_ERR_FAIL`. We follow that convention in everything except chknorm (and constant time functions of course).

We cannot change chknorm, as `pk_from_sk` needs the comparison result as a mask.

Overall I don't see enough reason to change anything right now.

Issue #1160, opened 2026-06-09
HOL-Light: prove secret-independence for `rej_uniform_eta{2,4}`


mkannwischer:
The AArch64 proofs for `rej_uniform_eta2_aarch64_asm` and `rej_uniform_eta4_aarch64_asm` currently establish functional correctness and memory safety only (`*_MEMSAFE`). They lack the secret-independence clause (`f_events`) that the other AArch64 routines' `*_SUBROUTINE_SAFE` theorems carry.

This was carried over from `rej_uniform`, but `rej_uniform` expands the public matrix A. It does not operate on secrets, so its timing is secret-independent and proving memory safety alone is sufficient.

However, `rej_uniform_eta{2,4}` produce the secret vectors s1/s2, so secret-independence must be proven. The rejection decision itself is not secret: it is acceptable for control flow and timing to depend on which candidate values are rejected.

hanno-becker:
This should be solved by allowing the bitmap of in-bound indices to be an argument to `f_events` -- but _not_ the entire contents of the array.

hanno-becker:
@dkostic Can you do this?

dkostic:
> [@dkostic](https://github.com/dkostic) Can you do this?

I'll try

mkannwischer:
@dkostic , are you still looking into this?

mkannwischer:
@dkostic, I'm assuming you don't have time to finish this?

@jakemas, @rod-chapman, do you have time to look into this? I think that's now our weakest point in the assurance story.
Tentatively adding this to v2.1.

jakemas:
Yes I’ll take a look

PR #1382, opened 2026-08-13
Add ML-DSA rejection acceptance foundation

BenKnill:
## Summary

- add a reusable HOL Light model of rejection-sampling acceptance bitmaps, prefix counts, stopping indices, table indices, and loop-event coordinates;
- connect the eta2 and eta4 nibble filters to that shared acceptance model;
- make same-directory proof dependencies explicit with `./...` paths.

This materially advances issue #1160, but it does **not** complete either `*_SUBROUTINE_SAFE` result. The remaining work is to connect the proved public acceptance/event skeleton to the concrete machine-event trace and then discharge the subroutine wrappers.

## Proof evidence and boundary

- The development loop and durable replay milestones succeeded for the staged foundation, culminating in the final 21-binding helper skeleton with zero added axioms.
- The helper committed here is the exact proved source (SHA-256 `1b0bcc7154c98740d8e09a048d1539f14a942205a43c4d88e153e36701a79384`).
- A later, separate full-source eta4 attempt was killed by `SIGKILL` (signal 9 / exit 137, approximately 5.45 GB resident memory). It produced no proof claim.
- The experimental, unproved `mldsa_rej_uniform_eta4_body_events.ml` theorem is deliberately excluded from this PR.

Issue #1160 therefore remains open.

## Packaging checks

- isolated worktree based on exact upstream commit `41c00dac3f3ce4f578f16aee9f9cd25e6bc05add`;
- three intended paths compared byte-for-byte with the proved development checkout;
- `git diff --check` passed;
- no Python bytecode or unrelated `.grok*` artifacts included;
- HOL/profile execution was intentionally not repeated during packaging.

### DIFF
--- a/proofs/hol_light/aarch64/proofs/mldsa_rej_uniform_eta_acceptance.ml
+++ b/proofs/hol_light/aarch64/proofs/mldsa_rej_uniform_eta_acceptance.ml
@@ -1,0 +1,525 @@
+(*
+ * Copyright (c) The mldsa-native project authors
+ * SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT-0
+ *)
+
+(* ========================================================================= *)
+(* Public acceptance pattern for eta rejection sampling.                    *)
+(*                                                                           *)
+(* The variable-time eta2/eta4 kernels may reveal which candidate nibbles   *)
+(* are accepted. They must not reveal the accepted nibble values themselves. *)
+(* The boolean list below records exactly that declassified pattern, in the  *)
+(* same low-nibble/high-nibble order as NIBBLES_OF_BYTES.                    *)
+(* ========================================================================= *)
+
+needs "s2n_bignum/arm/proofs/base.ml";;
+
+let MLDSA_REJ_ACCEPTANCE_BITMAP = define
+ `MLDSA_REJ_ACCEPTANCE_BITMAP bound ([]:byte list) = [] /\
+  MLDSA_REJ_ACCEPTANCE_BITMAP bound (CONS (b:byte) t) =
+    CONS (val b MOD 16 < bound)
+      (CONS (val b DIV 16 < bound)
+        (MLDSA_REJ_ACCEPTANCE_BITMAP bound t))`;;
+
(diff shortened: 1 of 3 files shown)


proofs/hol_light/aarch64/proofs/mldsa_rej_uniform_eta_acceptance.ml
```diff
@@ -0,0 +1,525 @@
...
+(* ========================================================================= *)
+
+needs "s2n_bignum/arm/proofs/base.ml";;
+
+let MLDSA_REJ_ACCEPTANCE_BITMAP = define
+ `MLDSA_REJ_ACCEPTANCE_BITMAP bound ([]:byte list) = [] /\
+  MLDSA_REJ_ACCEPTANCE_BITMAP bound (CONS (b:byte) t) =
+    CONS (val b MOD 16 < bound)
+      (CONS (val b DIV 16 < bound)
+        (MLDSA_REJ_ACCEPTANCE_BITMAP bound t))`;;
```
hanno-becker, line 22: The above bitmap definition seems to replicate logic that's already there: In the `MLDSA_REJ_UNIFORM_ETA2_SUBROUTINE_CORRECT` specification we already unpack the byte input into a list of nibbles of the right length (in that case, for example, it's `inlist : 4 word list`), so the bitmap in question is just a `map` applied on that `inlist`.

My expectation here is that we can largely follow the existing constant-time specification, but that instead of passing `inlist` to the event-generating function, we just pass the bitmap. Please note also that I believe @jakemas was looking into this issue in particular. You should either consult with him on the respective issue or pick a different one for a first contribution.

hanno-becker (changes requested):
Thank you @BenKnill for your interest and contribution! We don't currently accept contributions from people we don't know. If you are interested in contributing, would you kindly reach out on Discord or via mail to introduce yourself? Thank you!

hanno-becker:
As mentioned, @jakemas had been assigned to this and has since developed a proof. Closing this.

PR #1384, opened 2026-08-17
CI: gate SLOTHY and add label to trigger it

### LINKED ISSUE #1175: CI: Only run SLOTHY in main and when relevant files change
Around 1/3 of our EC2 CI cost is just SLOTHY. While #1073 will decrease that by maybe 1/3, I still consider this too high.
We should change the workflow to only run on PRs if either the asm sources, the SLOTHY Makefile, or any of the relevant nix files change. There should be a way to trigger it manually by attaching a label. It ~~shouldn't~~ should continue to run on every merge in main.

L-series:
- **CI: Only run SLOTHY when its inputs change**
- **CI: Allow requesting SLOTHY with the slothy-tests label**

- Resolves https://github.com/pq-code-package/mldsa-native/issues/1175

### DIFF
--- a/.github/workflows/slothy.yml
+++ b/.github/workflows/slothy.yml
@@ -10,6 +10,12 @@ on:
   workflow_dispatch:
   push:
     branches: ["main"]
+  # Standalone trigger so SLOTHY can be requested on an existing PR by adding
+  # the `slothy-tests` label. Adding `labeled` to all.yml instead would re-run
+  # all of CI on every label change.
+  pull_request:
+    branches: ["main"]
+    types: [ "labeled" ]
 
 jobs:
   # Coarse gate so the EC2 instance only starts when the SLOTHY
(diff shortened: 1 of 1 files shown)


L-series:
Resolves: https://github.com/pq-code-package/mldsa-native/issues/1175

mkannwischer:
@L-series, please next time check if someone is already assigned to an issue. I have this half finished locally, but you can take over.

L-series:
@mkannwischer apologies, its late and i completely glossed over the assignee tab.. Won't do that again. 

On the topic of SLOTHY however, what do you think of the addition of a per-file gate? I see that most of the time is spent optimizing NTT and iNTT, and so it might make sense to just have a finer grain gate for these two routines (i.e if any other routine is modified, only re-optimize what has been affected)

mkannwischer:
> @mkannwischer apologies, its late and i completely glossed over the assignee tab.. Won't do that again.

No worries! Thank you for your contributions.

> 
> On the topic of SLOTHY however, what do you think of the addition of a per-file gate? I see that most of the time is spent optimizing NTT and iNTT, and so it might make sense to just have a finer grain gate for these two routines (i.e if any other routine is modified, only re-optimize what has been affected)

We are not planning many more asm changes at this point, so the vast majority of re-runs will be triggered by nix flake changes, I believe. For those we have to re-run for all files anyway. I prefer to keep it simple for now.

mkannwischer (changes requested):
Thanks @L-series. This looks great. 
I think we can afford to keep running on main. Could you add that, please? Sorry I realized I had a typo in my original issue that suggested otherwise.

L-series:
> Thanks @L-series. This looks great. I think we can afford to keep running on main. Could you add that, please? Sorry I realized I had a typo in my original issue that suggested otherwise.

No worries! I added a:  
```
    push:
    branches: ["main"]
````
and 
` ${{ github.event_name == 'push' || ` which will run slothy on push to main unconditionally.

mkannwischer (approved):
Thanks @L-series for addressing my feedback. LGTM now.

PR #1392, opened 2026-08-21
Config: Move API-affecting options out of build-only section

### LINKED ISSUE #1391: Config: Some config options are wrongly listed as build internal
I am trying to enable the `MLD_CONFIG_CONTEXT_PARAMETER` option in my project but it is not working

The problem seem to be missing
**#define MLD_BUILD_INTERNAL**
before including CONFIG file
**#include MLD_CONFIG_FILE**

at below line

https://github.com/pq-code-package/mldsa-native/blob/main/mldsa/mldsa_native.h#L137

mkannwischer:
mldsa_native.h evaluates MLD_CONFIG_KEYGEN_PCT,
MLD_CONFIG_CONTEXT_PARAMETER, MLD_CONFIG_CONTEXT_PARAMETER_TYPE and MLD_CONFIG_REDUCE_RAM, but they were documented behind the MLD_BUILD_INTERNAL guard. Consumers include mldsa_native.h without MLD_BUILD_INTERNAL, so setting any of them left the consumer with prototypes lacking the context parameter and with MLD_TOTAL_ALLOC_* values for a configuration other than the one the library was built with.

 - Fixes #1391

### DIFF
--- a/examples/basic_deterministic/mldsa_native/mldsa_native_config.h
+++ b/examples/basic_deterministic/mldsa_native/mldsa_native_config.h
@@ -200,6 +200,83 @@
  * the implementation.
  */
 /* #define MLD_CONFIG_CONSTANTS_ONLY */
+
+/**
+ * MLD_CONFIG_KEYGEN_PCT
+ *
+ * Compliance with @[FIPS140_3_IG, p.87] requires a
+ * Pairwise Consistency Test (PCT) to be carried out on a freshly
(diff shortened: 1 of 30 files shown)


hanno-becker (approved):
Good catch. I double-checked that the public header indeed depends on all of those.

hanno-becker:
Why have our tests not caught that, though?

PR #1394, opened 2026-08-23
Update to CBMC 6.11

rod-chapman:
1. Update NIX config for CBMC 6.11 release/tag
2. Remove dummy calls to free() in CBMC harnesses, since these are no longer required.


mkannwischer (approved):
Thanks @rod-chapman. LGTM.

PR #1396, opened 2026-08-24
Pavona: Test Pavona integration against latest main

mkannwischer:
v2.0.0 has been merged with the required changes, so we no longer need the local patches.


hanno-becker (approved):
LGTM

PR #1399, opened 2026-08-28
Add example for the use of context parameters

hanno-becker:
* Port of https://github.com/pq-code-package/mlkem-native/pull/1877
* Based on #1400


mkannwischer (approved):
Thanks @hanno-becker. I checked that this is a faithful port of https://github.com/pq-code-package/mlkem-native/pull/1877

PR #1406, opened 2026-08-30
ACVP: Run test cases concurrently

### LINKED ISSUE #1350: CI: Decrease runtime of AVR tests
The AVR CI tests now take more than 100 minutes in CI due to the expanded ACVP coverage from #1307.
Such long CI time isn't sustainable.

Task:
 - Speed-up AVR CI (e.g., by skipping some ACVP tests).

mkannwischer:
The ACVP client spawns one subprocess per test case and runs them strictly serially. On emulated targets this dominates CI runtime.

Add --jobs to the ACVP client to run that many cases concurrently. scripts/tests hands the ACVP client the same -j it already uses for make, which defaults to os.cpu_count().

 - Resolve #1350

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ run_unit_87: unit_87
 run_unit: run_unit_44 run_unit_65 run_unit_87
 
 run_acvp: acvp
-	EXEC_WRAPPER="$(EXEC_WRAPPER)" python3 ./test/acvp/acvp_client.py $(if $(ACVP_VERSION),--version $(ACVP_VERSION)) $(ACVP_OPTIONS)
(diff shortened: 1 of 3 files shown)


mkannwischer:
The AVR tests are almost 2x faster, but 55 minutes is still excessive. Any other ideas @hanno-becker?

hanno-becker:
> The AVR tests are almost 2x faster, but 55 minutes is still excessive.

Why only 2x? Is that because it happens to run on a small instance?

mkannwischer:
> > The AVR tests are almost 2x faster, but 55 minutes is still excessive.
> 
> Why only 2x? Is that because it happens to run on a small instance?

Yes, the Github runners appear to have 2 vCPUs (4 threads).
One could move the AVR tests to an EC2 instance, but that requires more rework of the workflow.

hanno-becker:
> Yes, the Github runners appear to have 2 vCPUs (4 threads).
> One could move the AVR tests to an EC2 instance, but that requires more rework of the workflow.

Then I think the improvement in this PR is the right groundwork, and we can further improve by moving to a more powerful runner.

hanno-becker (approved):

PR #1407, opened 2026-08-30
Express alignment portably with alignas (C23) / _Alignas (C11/C17)

### LINKED ISSUE #1378: Consider adding support for `alignas` (C23) and `_Alignas` (C11/C17)
- Mirror issue of https://github.com/pq-code-package/mlkem-native/issues/1868

mkannwischer:
MLD_ALIGN had no portable definition: `__attribute__((aligned))` for GCC/Clang, `__declspec(align)` for MSVC, and on anything else it expanded to nothing at all. A toolchain outside those two families therefore silently dropped the 32-byte alignment.

C11 added _Alignas and C23 renamed it to alignas, so any compiler claiming either standard can express the constraint without an extension. Prefer those when `__STDC_VERSION__` advertises them and keep the extensions as the fallback for C90 and C99.

An alignment specifier belongs to the declaration specifiers, so unlike an attribute it cannot be attached to a typedef or a struct tag.

 - Resolve #1378

### DIFF
--- a/examples/context_parameter/example_context.h
+++ b/examples/context_parameter/example_context.h
@@ -18,7 +18,11 @@
 /* This mirrors MLD_ALIGN from mldsa/src/sys.h, including the fallback to no
  * alignment at all: on a toolchain that cannot express the constraint, the
  * library's own buffers are equally unaligned, so we are no worse off. */
-#if defined(__GNUC__)
+#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L
+#define EXAMPLE_ALIGN alignas(EXAMPLE_ALLOC_ALIGN)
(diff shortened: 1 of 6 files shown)


examples/context_parameter/example_context.h
```diff
@@ -18,7 +18,11 @@
 /* This mirrors MLD_ALIGN from mldsa/src/sys.h, including the fallback to no
  * alignment at all: on a toolchain that cannot express the constraint, the
  * library's own buffers are equally unaligned, so we are no worse off. */
-#if defined(__GNUC__)
+#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L
+#define EXAMPLE_ALIGN alignas(EXAMPLE_ALLOC_ALIGN)
```
hanno-becker, line 22: Have you confirmed that this works as expected?

PR #1411, opened 2026-09-01
mldsa: Explain packed byte counts via check-magic

mkannwischer:


hanno-becker (approved):
Thank you @mkannwischer for getting to the bottom of this. I agree with the fixes and also tested the script locally, confirming that it does break when proposing invalid explanations for small-digit constants.

PR #1415, opened 2026-09-05
Port: hal: Fix macOS cycle counting on Apple M4 and macOS 26.6

### LINKED ISSUE #1413: Benchmarks fail on AArch64/macOS 26.6.2
- Mirror issue of https://github.com/pq-code-package/mlkem-native/issues/1888

mkannwischer:
The kpc-based cycle counter hardcodes 0x02 as selector of the CORE_ACTIVE_CYCLE event. Since M4, Apple Silicon uses 0x11 instead, and the macOS 26.6 kernel rejects configurations with selectors that are unknown for the running CPU, so kpc_set_config fails. Look up the configuration word in the kperfdata database of the running CPU instead of hardcoding it.

The macOS 26.6 kernel further only applies the configuration when counting is started, so set it before enabling counting.

Tested on Apple M1 (macOS 26.4), where cycle counts are unchanged. Also tested on Apple M4 (macOS 26.6).

Ports pq-code-package/mlkem-native@240d2e96133408c14ce2a739cff4f91dc0d92a1c

 - Resolves #1413

### DIFF
--- a/test/hal/hal.c
+++ b/test/hal/hal.c
@@ -251,10 +251,23 @@ uint64_t get_cyclecounter(void)
   F(int, kperf_sample_get, int *)              \
   F(int, kpc_get_thread_counters, int, unsigned, void *)
 
+typedef struct kpep_db kpep_db;
+typedef struct kpep_config kpep_config;
+typedef struct kpep_event kpep_event;
+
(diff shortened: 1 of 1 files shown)


hanno-becker (approved):
I confirm this is a faithful port, and that it works on M5.

PR #1422, opened 2026-09-09
Unit tests: Expose native primitives in reduced-API builds

hanno-becker:
```
Several backend unit tests were additionally guarded by the public APIs
which use the tested primitives. Reduced-API builds therefore skipped
available native implementations, while declarations and generated
backend files used different guards.

Make `MLD_UNIT_TEST` expose decomposition, hint use, `polyz`
unpacking, and eta rejection sampling throughout the core and native
backends. Simplify these tests, and the pointwise multiplication test,
to depend only on their `MLD_USE_NATIVE_*` macros. Keep keypair-only
eta helpers and `polyw1` packing under their original public API
guards.

The vector pointwise accumulator has the opposite issue: reduced-RAM
builds use the per-polynomial path, but still include its L4, L5, and
L7 native implementations. Align the AArch64 and x86_64 declarations,
wrappers, and assembly with the core `MLD_CONFIG_REDUCE_RAM` guard,
while retaining those implementations for unit tests.

Update autogen and the generated AArch64 and x86_64 files
accordingly.
```


hanno-becker:
@mkannwischer This works, but the hierarchy of `#ifdef`s is just a huge mess. I wonder if rather than piling more on -- even if for the sake of consistency -- we should rethink our approach and simplify.

